PR #2, opened 2022-12-30
[Orca-114] tower metrics dag

BWMac:
This PR adds the tower metrics report airflow dag for `workflows-nextflow-dev`. It has been tested successfully on the development tower database. In the future, it will run on the production database once we have a permanent solution for storing production security credentials.

With the DAG being ~500 lines long, I moved the bulk of the logic, tasks, and variables needed for it to run to a new script in `dags/dag_content/tower_metrics_content.py`. This leaves only the DAG configuration, task flow, and dependency setting in `tower_metrics_dag.py`.


dags/tower_metrics_dag.py
```diff
@@ -0,0 +1,63 @@
...
+        aws_creds=AWS_CREDS,
+        db_name=DATABASE_NAME,
+        clone_name=CLONE_DATABASE_NAME,
+        subnet_group=prod_db_info["subnet_group"],
+        security_group=prod_db_info["security_group"],
+    )
+    # generate new password while clone is spinning up
+    password = generate_random_password(aws_creds=AWS_CREDS)
+    # update cloned database to have new password
+    modify = modify_database_clone(
```
BWMac, line 31: It may not be ideal, but it seems like in order to explicitly set dependencies, a task must be instantiated, even if there are no XCOMs to pass along. In this example, `modify_database_clone` does not `return` anything, but creating the object `modify` is necessary to then ensure that modifying the database comes after password generation and before the database is queried.

dags/dag_content/tower_metrics_content.py
```diff
@@ -0,0 +1,411 @@
...
+}
+#AWS region
+AWS_REGION = "us-east-1"
+
+# cluster names
+DATABASE_NAME = "tower"
+CLONE_DATABASE_NAME = "tower-clone"
+
+# query dictionary - maps query names to queries
+QUERY_DICT = {
```
BWMac, line 27: I think this is fine for now, but we may want this to live somewhere else in the future

dags/dag_content/tower_metrics_content.py
```diff
@@ -0,0 +1,411 @@
...
+import boto3
+import synapseclient
+
+#VARIABLES
+
+# AWS creds
+AWS_CREDS = {
+    "AWS_ACCESS_KEY_ID": Variable.get(
+        "TOWER_DB_ACCESS_KEY"
+    ),
```
BrunoGrandePhD: Just to confirm, we still use `Variable.get()` when values are stored with the secrets backend?

dags/dag_content/tower_metrics_content.py
BrunoGrandePhD: I agree that this is fine for the time being

dags/dag_content/tower_metrics_content.py
```diff
@@ -0,0 +1,411 @@
...
+            ) as set_up
+            group by project_name
+            order by count(*)/total_runs*100 desc;      
+            """,
+    "total_tower_users": """
+            select count(*) as total_users from tw_user;
+            """,
+    "active_users_past_month": """
+            select count(distinct user_name) as users_past_month from tw_workflow where date_created between date_sub(now(),INTERVAL 1 MONTH) and now();
+            """,
```
BrunoGrandePhD, line 82: I'm leaving a thought here for a future iteration: 

When we move these queries, we should think of storing them in a way that we can easily test them against a test database (_e.g._ [an in-memory RDBMS](https://stackoverflow.com/a/3047435)) so we can make sure that they return the values we expect.

dags/dag_content/tower_metrics_content.py
```diff
@@ -0,0 +1,411 @@
...
+}
+
+#FUNCTIONS
+
+# takes responses from query requests and packages them into dictionaries to be combined into JSON later
+def package_query_data(query_name: str, response: dict) -> dict:
```
BrunoGrandePhD: I found this function hard to follow. Instead of leaving several comments, I opted to present an alternative implementation that it easier to read. Note that I decided to drop the `query_name` from the function arguments because I think that logic belongs to the calling context. So, if you do replace this function with my version, you'll need to use `query_name` where it's called. 

```python
from itertools import chain
from typing import Any

def package_query_data(response: dict) -> list[dict[str, Any]]:
    result = list()
    column_names = [col["name"] for col in response["columnMetadata"]]
    for row in response["records"]:
        row_values = chain.from_iterable(field.values() for field in row)
        row_dict = dict(zip(column_names, row_values))
        result.append(row_dict)
    return result
```

Another big difference is that I'm returning a list of rows instead of a dictionary of rows organized by the first column/key. The main reason is that it results in much simpler and robust code. For example, if you return a dictionary, you would need to include code to check if there are any duplicate keys (see below). Also, it means that the SQL author needs to remember that the first `SELECT` column needs to contain unique values. 

```python
if row_key in data_dict:
    raise ValueError(f"Duplicate row key ({row_key}).")
```

Finally, by no longer needing the `row_key`, you no longer have to force the iteration of iterators using `list()`, resulting in cleaner code thanks to `chain()`. 

As far as I can tell, no other changes are needed to handle the list of rows instead of a dictionary of rows since the return value is only the serialized to JSON and uploaded to Synapse.

dags/dag_content/tower_metrics_content.py
```diff
@@ -0,0 +1,411 @@
...
+    """
+    rds = create_rds_client(aws_creds)
+
+    response = rds.describe_db_clusters(
+        DBClusterIdentifier=db_name,
+    )
+    subnet_group = response["DBClusters"][0].get("DBSubnetGroup")
+    security_group = response["DBClusters"][0]["VpcSecurityGroups"][0].get(
+        "VpcSecurityGroupId"
+    )
```
BrunoGrandePhD: You're using `dict.get()`, but you're not handling the case where the keys are absent and `None` is returned. Is this intentional? I've noticed this pattern elsewhere too.

dags/dag_content/tower_metrics_content.py
```diff
@@ -0,0 +1,411 @@
...
+    Returns:
+        str: secret arn string needed to access clone for queries
+    """
+    secrets = create_secret_client(aws_creds)
+
+    secret_string = (
+        "{"
+        + f'"dbInstanceIdentifier":"{clone_name}","engine":"aurora-mysql","host":"{db_info["host"]}","port":3306,"resourceId":"{db_info["resource_id"]}","username":"{db_info["user"]}","password":"{password}"'
+        + "}"
+    )
```
BrunoGrandePhD: You can use `json.dumps()` to serialize a Python dictionary to a string. 

```suggestion
    secret_dict = {
        "dbInstanceIdentifier": clone_name,
        "resourceId": db_info["resource_id"],
        "engine": "aurora-mysql",
        "host": db_info["host"],
        "port": 3306,
        "username": db_info["user"],
        "password": password,
    }
    secret_string = json.dumps(secret_dict)
```

dags/dag_content/tower_metrics_content.py
```diff
@@ -0,0 +1,411 @@
...
+    for query_name, query in QUERY_DICT.items():
+        response = rdsData.execute_statement(
+            resourceArn=resource_arn,
+            secretArn=secret_arn,
+            database="tower",
+            includeResultMetadata=True,
+            sql=query,
+        )
+
+        json_list.append(package_query_data(query_name, response))
```
BrunoGrandePhD: If you use my version of `package_query_data()` (see other comment), you'll need to update this part as follows:

```suggestion
        query_data = package_query_data(response)
        query_dict = {query_name: query_data}
        json_list.append(query_dict)
```

BrunoGrandePhD (changes requested):

dags/dag_content/tower_metrics_content.py
BWMac: Yes, that is correct.

dags/dag_content/tower_metrics_content.py
BWMac: I suppose I am banking on those parts of the response dictionary always being present as described in the boto3 docs. I can certainly add handling for `None` being returned if that is your recommendation.

BrunoGrandePhD: In that case, is there a reason why you aren't using regular dictionary indexing using `[]`? I'm just wondering why you switch back and forth between `[]` and `get()`.

BrunoGrandePhD (approved):
Looks good! 

Something that I came across that could be useful in the future is the `formatRecordsAs` argument for the boto3 `execute_statement()` method, which you can set to `JSON` ([docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds-data.html#RDSDataService.Client.execute_statement)). This might yield a more readily useful result compared to the more annoying records response that you're currently parsing.

dags/dag_content/tower_metrics_content.py
BWMac: I don't have a particular reason for using `.get()` in this DAG. After reading a bit about this method and how it is best used, I changed to using `[]` because we have no need to handle `None` or set default values in the absence of data.

PR #4, opened 2023-01-27
refactor reusable code

BWMac:
creates NextflowTowerService and AWSService to hold utility functions for interactions with external services

### DIFF
--- a/dags/services/AWSService.py
+++ b/dags/services/AWSService.py
@@ -1,0 +1,40 @@
+import boto3
+from airflow.models import Variable
+
+# AWS creds
+AWS_CREDS = {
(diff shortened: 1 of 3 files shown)


dags/services/AWSService.py
```diff
@@ -0,0 +1,40 @@
+import boto3
+from airflow.models import Variable
+
+# AWS creds
+AWS_CREDS = {
+    "AWS_ACCESS_KEY_ID": Variable.get("TOWER_DB_ACCESS_KEY"),
+    "AWS_SECRET_ACCESS_KEY": Variable.get("TOWER_DB_SECRET_ACCESS_KEY"),
+}
+# AWS region
```
thomasyu888, line 9: @BWMac Just a comment here, should we be using https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/operators/s3.html?

PR #5, opened 2023-01-30
[ORCA-91] - htan dag

BWMac:
This PR creates a DAG to launch the `nf-dcqc` workflow that Bruno has been working on. It also begins to lay the groundwork for future refactoring work where I will pull common code out of DAGs and into "services" which will eventually become classes with methods that can be pulled into DAG tasks, making the process of creating new DAGs less repetitive and more iterative. 

The DAG in its current version follows the following steps:

1. Get the input file for the workflow from Synapse
2. stage the input file in an S3 bucket within the Tower S3 structure
3. Launch the workflow, providing the S3 URI for the input file to the workflow.

I am initially opening this PR as a draft while I finish testing the DAG using Tower.


dags/services/AWSService.py
```diff
@@ -0,0 +1,40 @@
+import boto3
```
BWMac: This and `NextflowTowerService.py` will soon hold classes with all of the attributes and methods one would need to create a new DAG connecting to AWS resources or either of our Nextflow Tower instances. I created a ticket for this work in the "Migration of Airflow" Epic. The overall goal is to make DAGs more modular and make DAGs simple and readable by removing repetitive code from them.

BWMac (commented):
Leaving this as a draft while I finish testing launching the `nf-dcqc` workflow on tower. For now this will only launch workflows in the tower-dev space until we get airflow migrated over to the new AWS account

dags/htan-validate-dag.py
```diff
@@ -0,0 +1,83 @@
...
+    def get_synapse_input_file() -> str:
+        """
+        Gets synapse file (input csv) from synapse and saves it in buffer.
+        Passes along dict object with the path to the downloaded file and the name of the file.
+
+        Returns:
+            str: String containing the path to the downloaded file and the name of the file.
+        """
+        # simple param passing from run with config - sets my test file as default for now
+        syn_id = get_current_context()["params"].get("syn_id") or "syn50919899"
```
BWMac: the default synapse file currently is the `input_full.csv` file from the `nf-dcqc` repo, but [located](https://www.synapse.org/#!Synapse:syn50919899) within a test project that I have on Synpase

BWMac:
I was able to successfully run the workflow on tower using the DAG, so I am marking this PR as ready for review.

dags/htan-validate-dag.py
BrunoGrandePhD: FYI: The second argument to `dict.get()` (assuming that it's indeed a `dict`) is the default value is the key doesn't exist. 

```suggestion
        syn_id = get_current_context()["params"].get("syn_id", "syn50919899")
```

dags/htan-validate-dag.py
```diff
@@ -0,0 +1,83 @@
...
+        syn_token = Variable.get("SYNAPSE_AUTH_TOKEN")
+        syn = synapseclient.login(authToken=syn_token)
+        file_path = syn.get(syn_id).path
```
BrunoGrandePhD: Not yet, but we should soon consider using `fs-synapse` for these types of operations to remain Synapse-agnostic. Don't make this change for this PR since there is a discussion that needs to happen about fsspec (see [thread](https://github.com/Sage-Bionetworks/fs-synapse/pull/1#issuecomment-1407757995)).

dags/htan-validate-dag.py
```diff
@@ -0,0 +1,83 @@
...
+    def stage_input_in_s3(file_path: str) -> str:
+        """
+        Uploads file to Nextflow Tower S3 bucket, returns string path to file
+        Args:
+            file_path (str): String containing the path to the downloaded file to be uploaded to S3
+
+        Returns:
+            s3_uri (str): Path to S3 bucket location of file
+        """
+        s3_uri = upload_file_s3(file_path=file_path, bucket_name="orca-dev-project-tower-bucket")
```
BrunoGrandePhD, line 37: An assumption that might be worth documenting is that the file is small enough to fit on the Airflow instance. Do these files get deleted automatically? If not, it might be worth downloading into a temporary directory that gets automatically deleted to prevent bloat. You can look into using `tempfile.TemporaryDirectory()` with a context manager.

BrunoGrandePhD: Just noting that once we move to `fs-synapse`, we could leverage some [inter-FS copying functions](https://docs.pyfilesystem.org/en/latest/guide.html#moving-and-copying) available in PyFileSystem. I don't know if this avoids storing a local copy though. Probably not.

dags/htan-validate-dag.py
```diff
@@ -0,0 +1,83 @@
...
+        Args:
+            tower_utils (sagetasks.nextflowtower.utils.TowerUtils): TowerUtils class instance
+            workspace_id (str): Workspace ID for tower run
+            s3_path (str): Path to S3 location of input file
+        """
+        tower_utils = create_and_open_tower_workspace(platform="sage-dev", workspace_id=workspace_id)
+        tower_utils.launch_workflow(
+            compute_env_id="635ROvIWp5w17QVdRy0jkk",
+            pipeline="Sage-Bionetworks-Workflows/nf-dcqc",
+            run_name="nf-dcqc-airflow-test",
```
BrunoGrandePhD: What happens if you launch multiple workflows with the same run name? I know that Tower sometimes complains about this, but I'm unsure about the consequences. Do you run into any errors if you relaunch the same workflow via Airflow?

dags/htan-validate-dag.py
```diff
@@ -0,0 +1,83 @@
...
+            profiles=["docker"],
+            workspace_secrets=["SYNAPSE_AUTH_TOKEN"],
+            revision="main",
+            params_yaml=f'''
+                input: {s3_uri}
+                outdir: results/
```
BrunoGrandePhD: This should be set to an S3 prefix in the target workspace. Something like:

```suggestion
                outdir: s3://orca-dev-project-tower-scratch/07_days/nf-dcqc-airflow-test/
```

dags/htan-validate-dag.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        tower_utils = create_and_open_tower_workspace(platform="sage-dev", workspace_id=workspace_id)
+        tower_utils.launch_workflow(
+            compute_env_id="635ROvIWp5w17QVdRy0jkk",
+            pipeline="Sage-Bionetworks-Workflows/nf-dcqc",
+            run_name="nf-dcqc-airflow-test",
+            profiles=["docker"],
+            workspace_secrets=["SYNAPSE_AUTH_TOKEN"],
+            revision="main",
+            params_yaml=f'''
```
BrunoGrandePhD, line 59: Just in case you weren't aware of it, there's a handy `textwrap.dedent()` function to remove the leading indentation with multi-line strings.

dags/htan-validate-dag.py
BrunoGrandePhD: As a heads-up, I've run into issues when using dashes in run names. I think it's safer to use underscores.

dags/services/AWSService.py
BrunoGrandePhD: Some of this code can (should?) eventually live in sagetasks so we can test it more easily. We should meet at some point to align on our strategy for reusable Airflow code.

dags/services/AWSService.py
```diff
@@ -0,0 +1,40 @@
...
+
+    Args:
+        file_path (str): Path to file to be uploaded
+        bucket_name (str): Location in S3 for file to be uploaded
+    
+    Returns:
+        str: uri pointing to new uploaded file location in s3
+    """
+    s3_client = initialize_aws_client(resource="s3", aws_creds=AWS_CREDS, aws_region=AWS_REGION)
+    file_name = file_path.split("/")[-1]
```
BrunoGrandePhD: Functionally the same, but I think it's more readable: 

```suggestion
    file_name = os.path.basename(file_path)
```

You can also use `pathlib.Path()` to achieve the same, which would be even simpler if the function expected a `Path` object to begin with in the type hints: 

```suggestion
    file_name = Path(file_path).name
    
    # Or if it's already a Path object
    file_name = file_path.name
```

BrunoGrandePhD (commented):
Here's an initial review!

dags/services/AWSService.py
BWMac: I think I agree with the "should" point. For now, I will continue with my plan to create these separate services this sprint and then they could be moved to sagetasks from there.

dags/htan-validate-dag.py
BWMac: there is no problem launching multiple workflows with the same name, they both seem to launch just fine. I am going to remove the `run_name` for now though because it is not really necessary.

BWMac: And I will use underscores in the future

dags/htan-validate-dag.py
BWMac: I thought that the files would be deleted automatically, but they are not. I will look at `tempfile`

dags/htan-validate-dag.py
BWMac: I played around with it, but I don't think that `dedent` changes anything in this case

dags/htan-validate-dag.py
BWMac: added in latest commit

dags/htan-validate-dag.py
BWMac: added in latest commit

dags/services/AWSService.py
BWMac: added in latest commit

dags/htan-validate-dag.py
BWMac: I just added a commit that creates a `tempfile.TemporaryDirectory()`, saves the synapse file there, and uploads it to s3 before cleaning up the `temp_dir` all in the first step. I thought it made more sense than passing a `temp_dir` object from step one to step 2. Now there are only 2 total tasks which I think is fine as they are both still very simple.

dags/htan-validate-dag.py
BrunoGrandePhD: If you can use the same run name for different "launches", you can set it to `airflow-{{ run_id }}` so we can more easily cross-reference Airflow and Tower runs and query Tower runs.

dags/htan-validate-dag.py
```diff
@@ -19,32 +20,21 @@
...
-        """
-        Uploads file to Nextflow Tower S3 bucket, returns string path to file
-        Args:
-            file_path (str): String containing the path to the downloaded file to be uploaded to S3
-
-        Returns:
-            s3_uri (str): Path to S3 bucket location of file
-        """
+        temp_dir = TemporaryDirectory()
+        file_path = Path(syn.get(syn_id, downloadLocation=temp_dir.name).path)
```
BrunoGrandePhD: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** I would avoid calling more than one function in a line for readability. Even if you only use a variable once, it helps understand the right-hand side of the assignment without needing a comment. 

```suggestion
        syn_file = syn.get(syn_id, downloadLocation=temp_dir.name)
        file_path = Path(syn_file.path)
```

BrunoGrandePhD (approved):

dags/services/AWSService.py
```diff
@@ -0,0 +1,40 @@
...
+
+# AWS creds
+AWS_CREDS = {
+    "AWS_ACCESS_KEY_ID": Variable.get("TOWER_DB_ACCESS_KEY"),
+    "AWS_SECRET_ACCESS_KEY": Variable.get("TOWER_DB_SECRET_ACCESS_KEY"),
+}
+# AWS region
+AWS_REGION = "us-east-1"
+
+def initialize_aws_client(resource: str, aws_creds: dict = AWS_CREDS, aws_region: str = AWS_REGION):
```
thomasyu888: I think as an initial iteration, these are good, but what are your thoughts of using existing operators so we don't need to maintain our own: https://airflow.apache.org/docs/apache-airflow-providers-amazon/stable/operators/s3.html?

dags/services/AWSService.py
thomasyu888: Nit: I'm generally in favor of module names being snake_case. -> `aws_service.py` or `nextflow_tower_service.py`.  

Since we have plans on moving this to sagetasks and using airflow provided operators, this isn't a big deal.

thomasyu888 (approved):
🚀 Awesome work here! Just some comments, but going to approve.

dags/services/AWSService.py
BWMac: I will look more into this as part of the code refactoring ticket. I believe that the reason I did not use existing operators was that there were some things we needed that could not be done with existing operators alone, but it might make more sense to use them where possible.

dags/services/AWSService.py
BWMac: changed in latest commit

PR #7, opened 2023-02-08
[IBCDPE-469] - refactor dags

BWMac:
Not as much has changed as I originally intended when I created this ticket, but there are some important changes included in this PR:

1. Beginning the shift from secrets to connections in Airflow starting with connections to AWS services by leveraging the AWS provider `RdsHook`, `SecretsManagerHook`, and `AwsBaseHook`. This also makes replacing some custom logic with `aws.provider` functions simple and easy. Any remaining `boto3` functions are now accessed by getting a client connection from a hook.
2. I reverted `aws_service` and `nextflow_tower_service` back to scripts with reuseable functions/tasks while we work on the new `py-orca` package which will provide all of the advantages that a class structure would have provided here.
3. Along with the introduction of connections, we will also follow the rule that from now on, sets of secrets and connections stored in AWS SecretManager will be unique to the `dev` or `prod` accounts and instances, so that we do not need to change anything in the airflow repository to deploy to either instance.
4. the `create_and_open_tower_workspace` function is applied to simplify all nextflow tower launching dags.


dags/dag_content/tower_metrics_functions.py
```diff
@@ -311,16 +277,16 @@ def update_secret(
         "password": password,
     }
     secret_string = json.dumps(secret_dict)
-    response = secrets.update_secret(
+    response = secrets_client.update_secret(
         SecretId="Programmatic-DB-Clone-Access", SecretString=secret_string
     )
     secret_arn = response["ARN"]
-
+    time.sleep(30) #query_database always fails the first time because secret modification isnt done yet
```
BWMac: This is here because after refactoring, I noticed that the `query_database` task always fails on the first try (I think this may have been happening before as well, but not consistently). Since I know that the db_cluster status is "available" thanks to the waiter at the end of `modify_database`, I know that it has to have something to do with the modification of the secret. There is no waiter for secret modification to finish, and my attempts to create a waiter myself failed given that the secret does seem to be updated (new password applied) without any delay. It works consistently with this 30-second sleep at the moment, but I know that this is not ideal. I will continue to look into this while I await review. Any suggestions?

Pipfile
```diff
@@ -7,6 +7,7 @@ name = "pypi"
 apache-airflow = "*"
 sagetasks = "*"
 boto3 = "*"
+apache-airflow-providers-amazon = "*"
```
BrunoGrandePhD, line 10: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** I've started using `~=` for specifying package versions to avoid breaking changes. You don't need to update this PR, but I suggest you adopt this approach the next time you change a dependency. 

What I do is look up the latest version on PyPI. For example, the latest release of the `sevenbridges-python` package is 2.9.1. When I pin the package version, I'll use `sevenbridges-python~=2.9`, which allows for minor and patch updates, but not major updates with potentially backwards-incompatible changes.

BrunoGrandePhD, line 10: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** I've started using `~=` for specifying package versions to avoid breaking changes. You don't need to update this PR, but I suggest you adopt this approach the next time you change a dependency. 

What I do is look up the latest version on PyPI. For example, the latest release of the `sevenbridges-python` package is 2.9.1. When I pin the package version, I'll use `sevenbridges-python~=2.9`, which allows for minor and patch updates, but not major updates with potentially backwards-incompatible changes.

dags/dag_content/tower_metrics_content.py
```diff
@@ -1,21 +1,20 @@
...
 
 # VARIABLES
 
-# AWS creds
-AWS_CREDS = {
-    "AWS_ACCESS_KEY_ID": Variable.get("TOWER_DB_ACCESS_KEY"),
-    "AWS_SECRET_ACCESS_KEY": Variable.get("TOWER_DB_SECRET_ACCESS_KEY"),
-}
+# AWS conn
+AWS_CONN_ID = "TOWER_DB_CONNECTION"
```
BrunoGrandePhD, line 17: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** If this works, then we don't need to change anything, but I think the connection ID is a good candidate to be a DAG parameter ([docs](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/params.html)). This way, it can be overridden on launch.

dags/nf-validate-test.py
```diff
@@ -49,8 +33,7 @@ def launch_tower_workflow(tower_utils: TowerUtils, workspace_id: str):
             workspace_secrets=["SYNAPSE_AUTH_TOKEN"],
         )
 
-    tower_utils = open_tower_workspace()
-    launch_tower_workflow(tower_utils["tower_utils"], "4034472240746")
+    launch_tower_workflow("4034472240746")
```
BrunoGrandePhD, line 36: My instinct was to ask whether we can make this a DAG parameter, but we can wait until we have a Nextflow Tower hook that supports workspaces being encoded in the connection URI.

dags/htan-nf_dcqc_dag.py
```diff
@@ -48,18 +60,20 @@ def launch_tower_workflow(workspace_id: str, s3_uri: str):
             workspace_id (str): Workspace ID for tower run
             s3_uri (str): Path to S3 location of input file
         """
-        tower_utils = create_and_open_tower_workspace(platform="sage-dev", workspace_id=workspace_id)
+        tower_utils = create_and_open_tower_workspace(
+            platform="sage-dev", workspace_id=workspace_id
```
BrunoGrandePhD, line 64: We should tackle the Nextflow Tower hook sooner rather than later so you can get rid of hard-coded values like `sage-dev`, which hurt the portability of the DAG.

BrunoGrandePhD (approved):
I love seeing all of that red! 🟥 

My comments are mostly nits and future directions. Great work!

dags/nf-validate-test.py
BWMac: I agree on waiting

dags/htan-nf_dcqc_dag.py
BWMac: I agree, especially considering most of our DAGs launch tower workflows

dags/dag_content/tower_metrics_content.py
BWMac: The way this is currently structured, with the DAG tasks in a separate script, the `AWS_CONN_ID` and the hooks never actually touch the DAG script itself so parameterizing isn't really possible I don't think. We would either need to import the hooks to the DAG or provide the `aws_conn_id` as a parameter to each task

dags/dag_content/tower_metrics_content.py
```diff
@@ -246,9 +214,8 @@ def generate_random_password(aws_creds: dict) -> str:
     Returns:
         str: generated password
     """
-    secrets = create_secret_client(aws_creds)
-
-    response = secrets.get_random_password(
+    secrets_client = create_secrets_manager_connection()[1]
```
thomasyu888, line 217: Nit: one thing i see you doing in other places is something like:

```
hook, secrets_client = create_secrets_manager_connection()
```

But since you don't need the hook, the client is just fine

thomasyu888 (commented):
🚀 I forgot to submit my review, but looks great to me!

PR #9, opened 2023-02-10
Draft PR for ORCA-5, trigger daily genie validation

rxu17:
For sprint review show and tell

**Purpose:** Per the ticket here: [ORCA-5] (https://sagebionetworks.jira.com/browse/ORCA-5), the goal of the finalized PR would allow us to have an airflow DAG that triggers genie validation on production and development using Nextflow Workflow daily.

As part of this change, [nextflow_tower_functions.py](https://github.com/Sage-Bionetworks-Workflows/orca-recipes-airflow/pull/9/files#diff-87c3bb7a8106acdf2ecbecdb9046eb586f05eecfdaf4818f5349ac50679f349e) needed to have its `create_and_open_tower_workspace` function modified to take a `tower_access_token` parameter because different DAGs will use different Nextflow Tower Access Tokens

[ORCA-5]: https://sagebionetworks.jira.com/browse/ORCA-5?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thomasyu888 (approved):
🚀 @rxu17 can you add a better description here?  This LGTM.

dags/genie-nf-validate.py
```diff
@@ -0,0 +1,46 @@
...
+        Args:
+            workspace_id (str): Workspace ID for tower run
+        """
+        tower_utils = create_and_open_tower_workspace(
+            tower_access_token="TOWER_ACCESS_TOKEN_GENIE", platform="sage-dev", workspace_id=workspace_id
+        )
+        tower_utils.launch_workflow(
+            compute_env_id="1IEIFEJSVdDzQhwQlZRrHW",
+            pipeline="Sage-Bionetworks-Workflows/nf-genie",
+            run_name="airflow-genie-validate-test",
```
thomasyu888, line 32: One thing I will say is that are we going to have "prod" and "test" DAGs?  Or should the DAG just be configurable to launch the production validation job?

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -2,19 +2,20 @@
...
-    
+
     Returns:
         tower_utils (TowerUtils): Authenticated Tower Utils class
-    
+
     """
     client_args = TowerUtils.bundle_client_args(
-        auth_token=Variable.get("TOWER_ACCESS_TOKEN"), platform=platform
+        tower_access_token="TOWER_ACCESS_TOKEN", auth_token=Variable.get(tower_access_token), platform=platform
```
BrunoGrandePhD, line 18: Why are you providing this value to `bundle_client_args()`? While that function allows additional values to be provided thanks to `**kwargs`, it doesn't use this value. It seems like you only need this value for `Variable.get()`. 

Also, you can set the default value to `"TOWER_ACCESS_TOKEN"` so you don't have to change all of the DAGs. Since it would have a default value, this argument should be moved to the end of `create_and_open_tower_workspace()`.

dags/genie-nf-validate.py
BrunoGrandePhD: Ideally, there is no notion of test/dev vs prod in the DAGs. The variable bits between test/dev and prod should be stored in parameters or connections.

BrunoGrandePhD (changes requested):
As per Tom's comment, could you provide more context on what you're trying to achieve with this PR in the description? It really helps with code review since I'm not familiar with your assigned Jira tickets. Thanks!

rxu17:
Sorry both, I didn't mean to have you guys review in-progress untested code. Thank you for the comments. I'm just going to close this draft PR because the code is not actually ready for review per the initial comment on the PR about it being just for sprint review show and tell. I haven't had a chance to actually run this test code yet (hoping to sort out what is needed for that during the Airflow development meeting today), and subsequently make edits to it as a result.

PR #10, opened 2023-02-10
Fix inconsistent package name as Airflow provider

BrunoGrandePhD:


BrunoGrandePhD:
Closing this because I actually don't need to update the pinned version.

PR #12, opened 2023-02-21
[ORCA-5] Trigger daily main genie validation

rxu17:
**Purpose:** Per the ticket here: [ORCA-5](https://sagebionetworks.jira.com/browse/ORCA-5?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ) (https://sagebionetworks.jira.com/browse/ORCA-5), the goal of the finalized PR would allow us to have an airflow DAG that triggers genie validation on the main genie production pipeline using Nextflow Workflow daily.

As part of this change, [nextflow_tower_functions.py](https://github.com/Sage-Bionetworks-Workflows/orca-recipes-airflow/pull/9/files#diff-87c3bb7a8106acdf2ecbecdb9046eb586f05eecfdaf4818f5349ac50679f349e) needed to have its `create_and_open_tower_workspace` function modified to take a `tower_access_token` parameter because different DAGs will use different Nextflow Tower Access Tokens

[ORCA-5]: https://sagebionetworks.jira.com/browse/ORCA-5?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dags/dag_content/nextflow_tower_functions.py
```diff
@@ -2,19 +2,20 @@
 from sagetasks.nextflowtower.utils import TowerUtils
 
 
-def create_and_open_tower_workspace(platform: str, workspace_id: str) -> TowerUtils:
+def create_and_open_tower_workspace(tower_access_token : str, platform: str, workspace_id: str) -> TowerUtils:
     """Creates and opens Nextflow Tower Workspace using sagetasks TowerUtils class
 
     Args:
-        platform (str): Which Nextflow Tower instance to run the workflow on. 
+        tower_access_token (str): Which Nextflow Tower Access Token to use for workflow run
```
thomasyu888: Nit: Since this is the key to the secret that holds the access token that we want, this can be misconstrued as the token itself.

dags/genie-nf-validate.py
```diff
@@ -0,0 +1,65 @@
...
+    "revision": Param("main", type="string"),
+    "profile": Param("aws_prod", type="string"),
+    "only_validate": Param("true", type="string"),
+    "production": Param("true", type="string"),
+    "release": Param("13.3-consortium", type="string"),
+}
+
+
+dag_config = {
+    "schedule_interval": "@daily",
```
thomasyu888, line 22: Nit: I like the `@daily` for now as it runs everyday at midnight.  Eventually, we may want to execute this pipeline more often.  Something like this:

```
0 9,17 * * *
```

Run twice a day at 9am and 5pm - although since I'm sure this will be UTC time, we would have to convert it to match pacific time.

thomasyu888 (approved):
🚀 LGTM!  This is exciting!  We will be able to deprecate the old infrastructure really soon!

dags/genie-nf-validate.py
```diff
@@ -0,0 +1,65 @@
+from datetime import datetime
+
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+
+from dag_content.nextflow_tower_functions import create_and_open_tower_workspace
+
+dag_params = {
```
BWMac, line 8: This is awesome. A great example of a configurable dag for launching Nextflow Tower runs

requirements.txt
```diff
@@ -1,154 +1,153 @@
...
-markupsafe==2.1.1 ; python_version >= '3.7'
-marshmallow==3.19.0 ; python_version >= '3.7'
-marshmallow-enum==1.5.1
-marshmallow-oneofschema==3.0.1 ; python_version >= '3.6'
-marshmallow-sqlalchemy==0.26.1 ; python_version >= '3.6'
-mdit-py-plugins==0.3.3 ; python_version >= '3.7'
-mdurl==0.1.2 ; python_version >= '3.7'
-numpy==1.23.5 ; python_version >= '3.10'
-oauthlib==3.2.2 ; python_version >= '3.6'
-py-orca~=1.0
```
BWMac, line 94: I think we probably ought to keep this in. This diff is probably just a result of a difference between Bruno's last commit to `main` and your feature branch

dags/dag_content/nextflow_tower_functions.py
BWMac: NIT: it might be better for tower_access_token to be tower_secret_id or somehting like that. So that it can't be confused with the actual token string itself

BWMac (approved):

dags/dag_content/nextflow_tower_functions.py
BWMac: Just noticed that Tom already pointed this out. I am resolving this thread.

dags/genie-nf-validate.py
```diff
@@ -0,0 +1,65 @@
...
+        tower_utils.launch_workflow(
+            compute_env_id=context["params"]["compute_env_id"],
+            pipeline=context["params"]["pipeline"],
+            run_name=context["params"]["run_name"],
+            revision=context["params"]["revision"],
+            profiles=[context["params"]["profile"]],
+            workspace_secrets=["SYNAPSE_AUTH_TOKEN"],
+            params_yaml=f"""
+                only_validate: {context["params"]["only_validate"]},
+                production: {context["params"]["production"]},
```
thomasyu888: Remove the commas, because of yaml

dags/genie-nf-validate.py
```diff
@@ -0,0 +1,65 @@
...
+            platform="sage",
+            workspace_id=workspace_id,
+        )
+
+        tower_utils.launch_workflow(
+            compute_env_id=context["params"]["compute_env_id"],
+            pipeline=context["params"]["pipeline"],
+            run_name=context["params"]["run_name"],
+            revision=context["params"]["revision"],
+            profiles=[context["params"]["profile"]],
```
thomasyu888, line 55: Can you add the `work_dir` parameter and specify the scratch directory that removed files after 10 days? .  Specifically it will be this: `s3://genie-bpc-project-tower-scratch/10days`

thomasyu888 (changes requested):
One final small change, then feel free to merge

thomasyu888 (approved):

PR #13, opened 2023-02-24
[ORCA-145] Add new nf tower function to request tower api for latest compute env

rxu17:
**Purpose:** Pulls the latest compute environments from Nextflow Tower API via TowerClient and requests to the API programmatically to avoid genie validation airflow DAG from failing when the compute environments get updated in the tower workspace.

This adds two new functions: `get_latest_compute_environment` and `get_available_compute_environments` to `nextflow_tower_functions.py` to accomplish the above as well as additional Params to the `genie-nf-validate` DAG to pull the specific compute environment type, in this instance the ondemand environment.

This code heavily drew upon a similarly purposed function here that actually creates the compute environments: [create_compute_environment function in configure-tower-projects.py](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/26a961819100d68a752183fe7844ec73a15ce57e/bin/configure-tower-projects.py#L691)


dags/dag_content/nextflow_tower_functions.py
```diff
@@ -1,8 +1,12 @@
+from typing import Dict, Iterator, List, Optional, Sequence, Set, Tuple
```
BrunoGrandePhD: Some of these imports aren't used. You can use [autoflake](https://pypi.org/project/autoflake/) to do this automatically. We should add pre-commit hooks to this repository as well.

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -20,3 +24,55 @@ def create_and_open_tower_workspace(tower_secret_key : str, platform: str, works
...
+                int(comp_env["name"].split("-")[-1].replace("v", ""))
+                for comp_env in response["computeEnvs"]
+                if comp_env["platform"] == "aws-batch"
+                and comp_env["status"] == "AVAILABLE"
+            ]
+        )
+    except:
+        raise ValueError(
+            f"Cannot get compute environment version from Tower API request"
+        )
```
BrunoGrandePhD: It's a best practice to at least specify `except Exception:`. In this case, you might want to avoid losing the full stack trace by "extending" the exception using `from`. I tend to write the error message on a separate line to avoid additional indentation, but that's a [nit](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.). 

```suggestion
    except Exception as error:
        message = "Cannot get compute environment version from Tower API request"
        raise ValueError(message) from error
```

dags/genie-nf-validate.py
```diff
@@ -3,10 +3,14 @@
...
-from dag_content.nextflow_tower_functions import create_and_open_tower_workspace
+from dag_content.nextflow_tower_functions import (
+    create_and_open_tower_workspace,
+    get_latest_compute_environment
+)
 
 dag_params = {
-    "compute_env_id": Param("tJLngX5lvcm4EKhCVtAT", type="string"),
+    "compute_env_type": Param("EC2", type="string"),
+    "stack_name": Param("genie-bpc-project", type="string"),
```
BrunoGrandePhD, line 13: You can technically obtain this from the Tower workspace name. That said, we can worry about refactoring this bit when we move this logic into the orca package.

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -20,3 +24,55 @@ def create_and_open_tower_workspace(tower_secret_key : str, platform: str, works
...
+
+    # check for compute env id
+    for comp_env in response["computeEnvs"]:
+        if comp_env["platform"] == "aws-batch" and comp_env["status"] == "AVAILABLE":
+            if comp_env["name"] == comp_env_spot:
+                compute_env_ids["SPOT"] = comp_env["id"]
+            elif comp_env["name"] == comp_env_ec2:
+                compute_env_ids["EC2"] = comp_env["id"]
+    assert (
+        compute_env_ids["SPOT"] is not None and compute_env_ids["EC2"] is not None
```
thomasyu888: Nit: does it make sense to say "ON DEMAND" or EC2?

dags/genie-nf-validate.py
```diff
@@ -45,9 +48,14 @@ def launch_tower_workflow(workspace_id: str, **context):
...
         )
+        compute_envs = get_latest_compute_environment(
+            tower_utils = tower_utils,
+            stack_name = context["params"]["stack_name"],
+            workspace_id = workspace_id
+        )
 
         tower_utils.launch_workflow(
-            compute_env_id=context["params"]["compute_env_id"],
+            compute_env_id=compute_envs[context["params"]["compute_env_type"]],
```
BrunoGrandePhD: In your `get_latest_compute_environment()` function, you check if both `EC2` and `SPOT` are None. However, you don't handle the case if `compute_env_type` doesn't exist in the returned dictionary. Now, because the two compute environments are created automatically, the risk of an error here is low, so we can worry about better error handling when we port this over to `orca`.

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -20,3 +24,55 @@ def create_and_open_tower_workspace(tower_secret_key : str, platform: str, works
...
+    comp_env_ec2 = f"{stack_name}-ondemand-v{ce_version}"
+
+    # check for compute env id
+    for comp_env in response["computeEnvs"]:
+        if comp_env["platform"] == "aws-batch" and comp_env["status"] == "AVAILABLE":
+            if comp_env["name"] == comp_env_spot:
+                compute_env_ids["SPOT"] = comp_env["id"]
+            elif comp_env["name"] == comp_env_ec2:
+                compute_env_ids["EC2"] = comp_env["id"]
+    assert (
```
thomasyu888: Nit: try not to use assert for none testing functions.

thomasyu888 (approved):
🥏  LGTM! just some small nits but I'm approving.

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -20,3 +24,55 @@ def create_and_open_tower_workspace(tower_secret_key : str, platform: str, works
...
+    response = tower_utils.client.request("GET", endpoint, params=params)
+
+    # figure out highest compute env version by name
+    try:
+        ce_version = max(
+            [
+                int(comp_env["name"].split("-")[-1].replace("v", ""))
+                for comp_env in response["computeEnvs"]
+                if comp_env["platform"] == "aws-batch"
+                and comp_env["status"] == "AVAILABLE"
```
BrunoGrandePhD: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** I noticed that the `/compute-envs` endpoint provides a `status` query parameter that will presumably filter for the given status ([docs](https://tower.nf/openapi/index.html#get-/compute-envs)). So, you can move this filtering logic to the request.

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -20,3 +24,55 @@ def create_and_open_tower_workspace(tower_secret_key : str, platform: str, works
...
+    endpoint = "/compute-envs"
+    params = {"workspaceId": workspace_id}
+    response = tower_utils.client.request("GET", endpoint, params=params)
+
+    # figure out highest compute env version by name
+    try:
+        ce_version = max(
+            [
+                int(comp_env["name"].split("-")[-1].replace("v", ""))
+                for comp_env in response["computeEnvs"]
```
BrunoGrandePhD: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** This list comprehension is doing two things: (1) filtering for relevant compute environments and (2) parsing the version number. For readability, this should be split into separate operations.

dags/dag_content/nextflow_tower_functions.py
rxu17: Adding this to airflow discussion next week for how testing functions should be contributed to this repo because currently unsure the workflow for adding them for this repo

dags/dag_content/nextflow_tower_functions.py
rxu17: Hmm... I would have picked ON DEMAND, but in the nextflow-infra code here [create_compute_environment function in configure-tower-projects.py](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/26a961819100d68a752183fe7844ec73a15ce57e/bin/configure-tower-projects.py#L691), they use EC2 instead of ON DEMAND, so for consistency I chose EC2. ON DEMAND is more intuitive... 🤔

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -20,3 +24,55 @@ def create_and_open_tower_workspace(tower_secret_key : str, platform: str, works
...
+    compute_env_ids: dict[str, Optional[str]] = {"SPOT": None, "EC2": None}
+    endpoint = "/compute-envs"
+    params = {"workspaceId": workspace_id}
+    response = tower_utils.client.request("GET", endpoint, params=params)
+
+    # figure out highest compute env version by name
+    try:
+        ce_version = max(
+            [
+                int(comp_env["name"].split("-")[-1].replace("v", ""))
```
BrunoGrandePhD: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** It might be worth using regular expressions here for added robustness. I'm giving an example regex below, but in practice, you should handle the case if there are no matches. 

```python
>>> re.search(r"-v(\d+)", "ntap-add5-project-spot-v9_copy").group(1)
'9'
```

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -20,3 +24,55 @@ def create_and_open_tower_workspace(tower_secret_key : str, platform: str, works
...
+        raise ValueError(
+            f"Cannot get compute environment version from Tower API request"
+        )
+    # set env names
+    comp_env_spot = f"{stack_name}-spot-v{ce_version}"
+    comp_env_ec2 = f"{stack_name}-ondemand-v{ce_version}"
+
+    # check for compute env id
+    for comp_env in response["computeEnvs"]:
+        if comp_env["platform"] == "aws-batch" and comp_env["status"] == "AVAILABLE":
```
BrunoGrandePhD: Didn't you already filter for this above?

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -20,3 +24,55 @@ def create_and_open_tower_workspace(tower_secret_key : str, platform: str, works
...
+                and comp_env["status"] == "AVAILABLE"
+            ]
+        )
+    except:
+        raise ValueError(
+            f"Cannot get compute environment version from Tower API request"
+        )
+    # set env names
+    comp_env_spot = f"{stack_name}-spot-v{ce_version}"
+    comp_env_ec2 = f"{stack_name}-ondemand-v{ce_version}"
```
BrunoGrandePhD: I just want to point out that there are Tower workspaces with other compute environment names. For example, one has GPU-enabled compute environments like this one: `mc2-mcmicro-project-ondemand-gpus-v9`. GENIE doesn't need this, so addressing this limitation can happen when we refactor the code.

dags/dag_content/nextflow_tower_functions.py
rxu17: To do for a future refactor

dags/dag_content/nextflow_tower_functions.py
rxu17: To do for a future refactor

dags/genie-nf-validate.py
rxu17: Part of this is resolved by changing the function to be returning one compute env and adding an exception. Error handling refactor to do for a future refactor

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -1,8 +1,14 @@
+from typing import Dict, Optional
```
BrunoGrandePhD: Neither of these imported values are being used here. They should be removed.

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -28,51 +30,77 @@ def create_and_open_tower_workspace(
...
+            list(
+                set(
+                    [
+                        int(comp_env.split("-")[-1].replace("v", ""))
+                        for comp_env in available_envs.keys()
+                        if comp_env.startswith(compute_env_prefix)
+                    ]
+                )
+            )
         )
```
BrunoGrandePhD: This amount of nesting should be avoided. It makes the code harder to read. 

In this case, you are producing a list, converting it to a set, then converting it back to a list, and finally taking the max of that. As far as I can tell, you can pass the generator (list-comprehension without the square brackets) directly to `max()`. 

```python
>>> mylist = ["2", "5", "8", "2", "3", "8"]
>>> max(int(x) for x in mylist if not x.startswith("9"))
8
```

In your case, it would look as follows (not tested). Note that you don't need `.keys()` since keys are automatically  used by default when you iterate over a dictionary. 

```suggestion
        max_ce_version = max(
            int(comp_env.split("-")[-1].replace("v", ""))
            for comp_env in available_envs
            if comp_env.startswith(compute_env_prefix)
        )
```

dags/dag_content/nextflow_tower_functions.py
```diff
@@ -28,51 +30,77 @@ def create_and_open_tower_workspace(
...
+        message = "Cannot get compute environment version from Tower API request"
+        raise ValueError(message) from error
+
+    compute_env_name = (
+        f"{stack_name}-{COMPUTE_ENV_MODELS[compute_env_model]}-v{max_ce_version}"
+    )
+
+    try:
+        compute_env_id = available_envs[compute_env_name]
+    except Exception as error:
```
BrunoGrandePhD: In general, you should be specific about which errors you want to handle. In this case, I'm guessing you were thinking of `KeyError`. My earlier [comment](https://github.com/Sage-Bionetworks-Workflows/orca-recipes-airflow/pull/13#discussion_r1117809287) about this was to specify **at least** `Exception`. The more specific, the better.

BrunoGrandePhD (approved):
I'm not going to require another round of review, but I made a few comments to improve the code and its readability.

dags/dag_content/nextflow_tower_functions.py
rxu17: Wow. I have no idea why I did list -> set -> list here ...

PR #15, opened 2023-03-13
adds  airflow_password retrieval script and updates README

BWMac:
This PR adds a bash script used to retrieve the production airflow password from AWS secret manager. This way, when airflow is deployed, this script can be used in the CloudFormation template to retrieve the password and configure the admin account along with the added bonus of allowing users and developers to retrieve the password themselves and use it to log in. CC @BrunoGrandePhD

### DIFF
--- a/airflow_password.sh
+++ b/airflow_password.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+airflow_password=$(aws --output text secretsmanager get-secret-value --secret-id airflow_password --query SecretString --region us-east-1)
+echo $airflow_password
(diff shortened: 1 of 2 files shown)


airflow_password.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+airflow_password=$(aws --output text secretsmanager get-secret-value --secret-id airflow_password --query SecretString --region us-east-1)
+echo $airflow_password
```
BrunoGrandePhD: Is there a reason to store the output in a variable that's echoed? The following version should achieve the same. 

```suggestion
aws --output text secretsmanager get-secret-value --secret-id airflow_password --query SecretString --region us-east-1
```

BrunoGrandePhD (approved):

PR #16, opened 2023-03-13
[ORCA-143] Add Tom to Tower Metrics Report

BWMac:
This PR adds functionality for the list of synapse users that are sent the tower metrics dag report to be parameterized and adds Tom to the list of users that receive the report. Because there is apparently no support for python lists in airflow `Param`, the list is provided as a comma-separated string which is converted to a list when the usernames are needed in the `send_synapse_notification` task. I have also changed the schedule on this DAG to run every Friday at 3PM PST, let me know if there is a time or cadence that you would prefer instead @thomasyu888.

This change has been successfully tested on the `dnt-dev` airflow instance.

### DIFF
--- a/dags/tower_metrics_dag.py
+++ b/dags/tower_metrics_dag.py
@@ -46,13 +46,13 @@ def tower_metrics_dag():
     # delete cloned database - wait for it to be gone before completing process
     delete = delete_clone_database(clone_name=CLONE_DATABASE_NAME)
 
-
-    #add missing dependencies
+    # add missing dependencies
     prod_db_info >> clone_db_info
     [clone_db_info, password] >> modify
     [clone_db_info, password] >> secret_arn
     [secret_arn, modify] >> json_list
     json_list >> [export, delete]
     export >> send
+
 
 tower_metrics_dag = tower_metrics_dag()
--- a/dags/tower_metrics_dag.py
+++ b/dags/tower_metrics_dag.py
@@ -14,7 +14,7 @@ dag_params = {
 
 
 @dag(
-    schedule_interval="0 15 * * 5",  # Fridays at 3PM PST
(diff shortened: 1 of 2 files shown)


BrunoGrandePhD:
I marked this PR as a draft since it's a WIP. If you still want a review, let us know!

BWMac:
> I marked this PR as a draft since it's a WIP. If you still want a review, let us know!

Thanks for doing that. I created it in VSCode and selected the "create as draft" option, so not sure why it was not created that way!

dags/tower_metrics_dag.py
```diff
@@ -1,18 +1,27 @@
 from datetime import datetime
 
 from airflow.decorators import dag
+from airflow.models.param import Param
 
 from dag_content.tower_metrics_content import *
 
+dag_params = {
+    "user_list": Param(
+        "bwmac,thomas.yu",  # no support for python lists
```
thomasyu888, line 10: Thats very unfortunate, but totally fine.

dags/tower_metrics_dag.py
```diff
@@ -1,18 +1,27 @@
...
+        type="string",
+    )
+}
+
 
 @dag(
-    schedule_interval="@weekly",
+    schedule_interval="0 15 * * 5",  # Fridays at 3PM PST
```
thomasyu888: Will this be UTC time?

thomasyu888 (changes requested):
LG, but just one question.

dags/tower_metrics_dag.py
BWMac: I did forget to adjust for UTC, now I think it is correct

thomasyu888 (approved):
💯 LGTM! Thanks!

PR #17, opened 2023-04-05
[ORCA-146] Configure docker compose stack to pass through AWS credentials

BrunoGrandePhD:
In order to leverage GitHub Codespaces, we need to provide Airflow with access to the connection credentials in AWS Secrets Manager. I wasn't able to get this to work when I first tested Codespaces. I realized that the AWS credentials being set in environment variables weren't being picked up by the Secrets Manager backend in Airflow unless they were provided in the `.env` file. That's when I realized that the docker compose stack needs to explicitly pass through environment variables. Since then, I'm able to access secrets from AWS in Codespaces! 

image

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -40,5 +40,10 @@
     // "customizations": {},
 
     // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
-    "remoteUser": "vscode"
+    "remoteUser": "vscode",
+
+    // Ensure enough memory for multi-container Airflow stack
+    "hostRequirements": {
+        "memory": "8gb"
+    }
 }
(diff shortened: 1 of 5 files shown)


thomasyu888 (commented):
🐉 Superb! I was testing everything, and I think there are a couple things that need to be updated.

1. Update instructions to make sure to include the DPE-prod credentials in the `.env` file
2. ~~I keep getting a nginx bad gateway~~ ignore, it spins up after awhile if I set the right credentials (DPE-prod?)

@BrunoGrandePhD I am unfortunately getting a workflow failure: 

> [2023-04-05, 03:34:02 UTC] {connection.py:429} ERROR - Unable to retrieve connection from secrets backend (SecretsManagerBackend). Checking subsequent secrets backend.

image

BrunoGrandePhD:
@thomasyu888: You don't need to include credentials in the `.env` file. They're already set as Codespaces secrets on the repo. Right now, this is set as an IAM user on `dnt-dev`, which might explain why your test with `dpe-prod` doesn't work. I don't know if the `cavatica_test` connection exists yet in `dpe-prod`. 

image

thomasyu888 (approved):
Thanks @BrunoGrandePhD !   That makes sense - I'm going to pre-approve, but still think we should document some minor things. (Nits)

image

1. If you encounter the `nginx bad gateway` when navigating to the port provided by codespaces, just wait and refresh a couple of times - Airflow takes awhile to setup load.
2. There is no need to set the `.env` with AWS credentials because they are set as github secrets.

BrunoGrandePhD:
I've updated the docs

PR #18, opened 2023-04-14
[ORCA-158] Adds Workflow Count and Workflow Distribution to Tower Metrics

BWMac:
This PR adds two queries to `tower_metrics_content.py` to be run with the weekly Tower Metrics report. They are the following:

For the past 7 days...
1. Display the distribution of workflows executed per workspace
2. Display the number of workflows executed per workspace

An example of the new report can be seen [here](https://www.synapse.org/#!Synapse:syn48188330). So far this has only been run on the `tower-dev` database and this example report reflects that. Once the updates to this DAG have been added to our `prod` instance we can run the report with the real metrics.

### DIFF
--- a/dags/dag_content/tower_metrics_content.py
+++ b/dags/dag_content/tower_metrics_content.py
@@ -38,12 +38,12 @@ QUERY_DICT = {
             w.full_name) as total_workflow_runs;
     """,
     "workflow_distribution_past_week": """
-            select workspace, workflow, runs, run_distribution from (
+            select workspace, workflow, runs, percentage from (
             select
             w.full_name as workspace,
             wf.project_name as workflow,
             count(*) AS runs,
-            count(*) / t.total_runs * 100 as run_distribution
+            count(*) / t.total_runs * 100 as percentage
             from 
             tw_workflow wf 
             join tw_workspace w on wf.workspace_id = w.id 
--- a/dags/dag_content/tower_metrics_content.py
+++ b/dags/dag_content/tower_metrics_content.py
@@ -24,6 +24,40 @@ CLONE_DATABASE_NAME = "tower-clone"
 
 # query dictionary - maps query names to queries
 QUERY_DICT = {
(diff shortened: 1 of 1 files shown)


dags/dag_content/tower_metrics_content.py
```diff
@@ -24,6 +24,40 @@
...
+            group by 
+            w.full_name) as total_workflow_runs;
+    """,
+    "workflow_distribution_past_week": """
+            select workspace, workflow, runs, run_distribution from (
+            select
+            w.full_name as workspace,
+            wf.project_name as workflow,
+            count(*) AS runs,
+            count(*) / t.total_runs * 100 as run_distribution
```
thomasyu888: Nit: Since everything else in the output is "percentage", I vote for consistency.
```suggestion
            count(*) / t.total_runs * 100 as percentage
```

dags/dag_content/tower_metrics_content.py
```diff
@@ -24,6 +24,40 @@
...
+            from 
+            tw_workflow wf 
+            join tw_workspace w on wf.workspace_id = w.id
+            where 
+            wf.date_created >= date_sub(now(), interval 1 week) 
+            group by 
+            w.full_name) as total_workflow_runs;
+    """,
+    "workflow_distribution_past_week": """
+            select workspace, workflow, runs, run_distribution from (
```
thomasyu888: Nit: Same comment here about "percentage"

thomasyu888 (approved):
🔥 Awesome!  Just a minor comment, but going to pre-approve!

dags/dag_content/tower_metrics_content.py
BWMac: addressed in latest commit

dags/dag_content/tower_metrics_content.py
BWMac: addressed in latest commit

PR #19, opened 2023-06-01
[ORCA-88] Challenge Submission Evaluation DAG

BWMac:
This PR adds the new `challenge-submission-dag.py` DAG. It also updates the version of py-orca, and adds its dependent package `fs-synapse`.

The new DAG works in three basic steps. First by using `@task.branch()`, it determines if there are new submissions in the evaluation queue for the challenge by using the new `monitor_evaluation_queue` method in `py-orca`. If there are no new submissions, the DAG is terminated by the `stop_dag` task. If there are, the DAG continues on to the `launch_model2data_workflow` task, and `nf-model2data` is triggered in Tower. Finally, `monitor_model2data_workflow` monitors the Tower run and terminates the DAG once the run has reached a terminal state.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -7,7 +7,8 @@ name = "pypi"
 apache-airflow = "~=2.5"
 apache-airflow-providers-amazon = "~=7.2"
 sagetasks = "~=0.4"
-py-orca = "~=1.0"
+py-orca = ">1.3.0"
+fs-synapse = "~=1.0"
 
 [dev-packages]
 
--- a/Pipfile
+++ b/Pipfile
@@ -7,8 +7,7 @@ name = "pypi"
 apache-airflow = "~=2.5"
 apache-airflow-providers-amazon = "~=7.2"
 sagetasks = "~=0.4"
-py-orca = ">1.3.0"
-fs-synapse = "~=1.0"
+py-orca = {version = "~=1.3", extras = ["all"]}
 
 [dev-packages]
 
--- a/dags/challenge-submission-dag.py
+++ b/dags/challenge-submission-dag.py
@@ -12,6 +12,7 @@ from orca.services.synapse import SynapseHook
 
 dag_params = {
     "synapse_conn_id": Param("SYNAPSE_CHALLENGE_CONN", type="string"),
+    "synapse_evaluation_id": Param("9615332", type="string"),
     "tower_conn_id": Param("EXAMPLE_PROJECT_TOWER_CONN", type="string"),
     "tower_run_name": Param("model_submission_evaluation", type="string"),
     "tower_view_id": Param("syn51356905", type="string"),
@@ -42,7 +43,9 @@ def challenge_submission_dag():
             evaluation_id (str): Evaluation ID for challenge
         """
         hook = SynapseHook(context["params"]["synapse_conn_id"])
-        if hook.ops.monitor_evaluation_queue("9615332"):
+        if hook.ops.monitor_evaluation_queue(
+            context["params"]["synapse_evaluation_id"]
+        ):
             return "launch_model2data_workflow"
         return "stop_dag"
 
--- a/dags/challenge-submission-dag.py
+++ b/dags/challenge-submission-dag.py
@@ -71,10 +71,12 @@ def challenge_submission_dag():
         )
         return run_id
 
-    @task()
+    @task.sensor(poke_interval=300, timeout=604800, mode="reschedule")
     def monitor_model2data_workflow(run_id: str, **context):
         hook = NextflowTowerHook(context["params"]["tower_conn_id"])
-        asyncio.run(hook.ops.monitor_workflow(run_id=run_id))
+        workflow = hook.ops.get_workflow(run_id)
+        print(f"Current workflow state: {workflow.status.state.value}")
+        return workflow.status.is_done
 
     submission_check = check_for_new_submissions()
     stop = stop_dag()
--- a/dags/challenge-submission-dag.py
+++ b/dags/challenge-submission-dag.py
@@ -1,0 +1,88 @@
+import asyncio
+
+from datetime import datetime
+
+from airflow.decorators import dag, task
+from airflow.models import Param
+
+from orca.services.nextflowtower import NextflowTowerHook
+from orca.services.nextflowtower.models import LaunchInfo
+from orca.services.synapse import SynapseHook
+
+
+dag_params = {
+    "synapse_conn_id": Param("SYNAPSE_CHALLENGE_CONN", type="string"),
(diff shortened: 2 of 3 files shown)


dags/challenge-submission-dag.py
```diff
@@ -0,0 +1,91 @@
...
+        "retries": 2,
+    },
+    "tags": ["nextflow_tower"],
+    "params": dag_params,
+}
+
+
+@dag(**dag_config)
+def challenge_submission_dag():
+    @task.branch()
```
thomasyu888, line 37: TIL: task branch. This is quite nice!

dags/challenge-submission-dag.py
```diff
@@ -0,0 +1,91 @@
...
+    @task()
+    def stop_dag():
+        pass
+
+    @task()
+    def launch_model2data_workflow(**context):
+        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
+        info = LaunchInfo(
+            run_name=context["params"]["tower_run_name"],
+            pipeline="https://github.com/Sage-Bionetworks-Workflows/nf-model2data",
```
thomasyu888, line 61: So this is the beginning of whether we want airflow to manage the logic of "invalid docker submission" or if we want nextflow to do that.

thomasyu888 (approved):
🪐 This looks great!  I'm excited about this work!

Pipfile
```diff
@@ -7,7 +7,8 @@ name = "pypi"
 apache-airflow = "~=2.5"
 apache-airflow-providers-amazon = "~=7.2"
 sagetasks = "~=0.4"
-py-orca = "~=1.0"
+py-orca = ">1.3.0"
+fs-synapse = "~=1.0"
```
BrunoGrandePhD: I don't see you using fs-synapse, so I'm guessing you installed it because py-orca complained when you tried using the Synapse hook. Instead of including it here, I suggest that you specify the `all` extra for py-orca as follows. This will install the non-essential dependencies, including fs-synapse. We should update py-orca to inform users to do this if they run into an import error so this feature is more discoverable. 

```suggestion
py-orca = {version = "~=1.3", extras = ["all"]}
```

By the way, if you want to update the `Pipfile.lock` file, you can run `pipenv lock --dev`, which will install the latest version of py-orca and its dependencies. Using `~=1.3` means that it will allow py-orca to progress as long as the major version remains 1. Once version 2.0 is released, pipenv will continue to install version 1.X.

dags/challenge-submission-dag.py
```diff
@@ -0,0 +1,91 @@
+import asyncio
+
+from datetime import datetime
+
+from airflow.decorators import dag, task
+from airflow.models import Param
+
+from orca.services.nextflowtower import NextflowTowerHook
+from orca.services.nextflowtower.models import LaunchInfo
```
BrunoGrandePhD, line 9: Because `LaunchInfo` is a user-facing model, we can import it into the `nextflowtower` module (in the `__init__.py` file) so it doesn't require its own import line. Just an idea for the next time that we update py-orca.

dags/challenge-submission-dag.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+        run_id = hook.ops.launch_workflow(
+            info, context["params"]["tower_compute_env_type"]
+        )
+        return run_id
+
+    @task()
+    def monitor_model2data_workflow(run_id: str, **context):
+        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
+        asyncio.run(hook.ops.monitor_workflow(run_id=run_id))
```
BrunoGrandePhD: You can consider making this a sensor instead using the same approach I used [here](https://github.com/Sage-Bionetworks-Workflows/orca-recipes-airflow/blob/0ef9ccbd958e691f37f2d99bc19b10c7661fa00a/dags/cavatica-launch-poc-v2.py#L50-L61). There's no practical difference though, so I wouldn't worry about it for this PR. Just mentioning it for awareness.

dags/challenge-submission-dag.py
```diff
@@ -0,0 +1,91 @@
...
+    monitor = monitor_model2data_workflow(run_id=run_id)
+
+    submission_check >> [
+        run_id,
+        stop,
+    ]
+    run_id >> monitor
+
+
+challenge_submission_dag = challenge_submission_dag()
```
BrunoGrandePhD: I would avoid creating a DAG instance that uses the same variable name as the DAG function. According to the [Airflow docs](https://airflow.apache.org/docs/apache-airflow/stable/tutorial/taskflow.html#instantiate-a-dag):

> It’s no longer required to “register” the DAG into a global variable for Airflow to be able to detect the dag if that DAG is used inside a with block, or if it is the result of a @dag decorated function.

So, I think the following should suffice: 

```suggestion
challenge_submission_dag()
```

BrunoGrandePhD (approved):
I'm approving because nothing is a dealbreaker. 🚀 

It's so cool to see multiple py-orca hooks being used in practice!

BrunoGrandePhD (approved):
I'm approving because nothing is a dealbreaker. 🚀

BrunoGrandePhD (approved):
I'm approving because nothing is a dealbreaker. 🚀

dags/challenge-submission-dag.py
thomasyu888: +1 Totally spaced on this!

dags/challenge-submission-dag.py
BWMac: I guess the question is what determines an invalid docker submission. Is that something that becomes apparent when you try to execute the container (Nextflow), or is it something that you catch with data/metadata when you retrieve the submission from Synapse (maybe airflow)

dags/challenge-submission-dag.py
BWMac: doesn't the `monitor_workflow` async function do everything that is encapsulated in your example function by itself though? It already checks every 5 minutes by default. Wasn't this more necessary before adding `monitor_workflow` to `py-orca`?

BrunoGrandePhD: I think the main difference is that Airflow sensors can be configured in `reschedule` mode, which allows the worker slot to be freed up between checks ([docs](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/sensors.html)). This would matter more if we waited longer between checks and when we have more concurrent monitoring tasks. We don't want dozens of worker slots being taken up by jobs running `sleep()`. 

In my example, I went with the `poke` mode because I don't think `reschedule` has been implemented yet with the `@task.sensor()` decorator, but that's a temporary limitation.

BWMac: I played around for a while and converted `monitor_model2data_workflow` to a sensor using the reschedule mode. I tested it and it is set up to reschedule 5 minutes out when the status it receives back is not terminal.

PR #22, opened 2023-07-13
Refactor with `py-orca`

BWMac:
This PR refactors existing DAGs to use the hooks we have created in `py-orca` and generally updates the repository. It also removes two DAGs that we no longer need. Detailed list of changes:

- Bump `py-orca` version
- Delete `dags/cavatica-launch-poc.py` and `dags/nf-validate-test.py` DAGs
- Refactor `dags/dag_content/tower_metrics_content.py` to use `SynapseHook`
- Refactor `dags/nf-hello-test.py`, `dags/htan-nf_dcqc_dag.py`, and `dags/genie-nf-validate.py` to use `NextflowTowerHook` and `LaunchInfo`
- Delete `dags/dag_content/utils.py` and `dags/dag_content/nextflow_tower_functions.py` because refactoring makes them redundant
- added needed connection URIs to the AWS secrets backend
    - For Tower connections, I created one per workspace
    
All DAGs have been thoroughly tested, with the exception of `dags/genie-nf-validate.py` because I do not have access to the Genie Tower workspace. I would really appreciate it if someone with access could help with the testing.

I am also still thinking about how we should increment `run_name`'s for Nextflow Tower DAGs that run on a schedule like the genie one. If we don't increment, `py-orca` will think that we are trying to resume the run every time (this makes me think that the incrementation should maybe be part of `py-orca` itself. In past example scripts, the increment has been linked to a specific dataset that was being run through a workflow, but this is not always an available option so we should try to come up with a standard way to do this. For now, I have appended "_<today's-date>" to the run name so that it can at least distinguish between runs on different days, but this could have unintended consequences for particularly long runs that are retried.


thomasyu888 (commented):
Thanks @BWMac ! I'll take a look today and test the GENIE dag.

I like the date timestamp, but I thought I remember a parameter we can set to false to not resume workflow executions by default.

We can probably use that for the GENIE dag.

BWMac:
@thomasyu888 Ah, that's right. I will make that change!

dags/genie-nf-validate.py
thomasyu888, line 1: @rxu17 can you help me test this DAG?

thomasyu888 (approved):
🔥 I'm going to pre-approve, but I'd wait for rixing's test.

Thanks Brad!

rxu17 (approved):
The genie DAG ran successfully (tower workflow was triggered and finished) after testing it on the local airflow server. Got the validation emails.

dags/dag_content/tower_metrics_content.py
```diff
@@ -4,12 +4,13 @@
 from itertools import chain
 from typing import Any
 
-import synapseclient
+from synapseclient import File
```
philerooski, line 7: fwiw, there are some downsides to importing this way. The biggest ones imo are:

* No performance benefit from doing "from ... import ...". The entire `synapseclient` module is still imported.
* Need to update import statement each time you need to use something else from `synapseclient`.
* It's less explicit in the code where a module/class/method is coming from. `File` is generic enough that it could be mistaken for something from the python standard library.

More info: 
* https://stackoverflow.com/questions/710551/use-import-module-or-from-module-import
* https://google.github.io/styleguide/pyguide.html#22-imports

PR #23, opened 2023-07-18
[GEN-782] Fix genie validation dag

rxu17:
**Purpose:** This fixes the genie DAG so the parameters passed through the DAG are getting propagated through to the Nextflow workflow call. Previously, I believe it was because the `params_yaml` doesn't exist in the [LaunchInfo's specification](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/main/src/orca/services/nextflowtower/models.py#L153-L169), and only `params` does, the Nextflow workflow behavior in light of not receiving the defined params from Airflow would default to the default param values in the main genie nextflow pipeline.

Did test runs in the local Airflow UI, and the latest nextflow tower run reflects the param values passed from Airflow correctly:
image

Also the [params for nf-genie](https://github.com/Sage-Bionetworks-Workflows/nf-genie/blob/main/nextflow_schema.json#L14-L24) had been updated, so this DAG has been updated to reflect that.

### DIFF
--- a/dags/genie-nf-validate.py
+++ b/dags/genie-nf-validate.py
@@ -13,8 +13,7 @@ dag_params = {
     "pipeline": Param("Sage-Bionetworks-Workflows/nf-genie", type="string"),
     "revision": Param("main", type="string"),
     "profile": Param("aws_prod", type="string"),
-    "only_validate": Param("true", type="string"),
-    "production": Param("true", type="string"),
+    "process_type": Param("only_validate", type="string"),
     "release": Param("13.3-consortium", type="string"),
     "work_dir": Param("s3://genie-bpc-project-tower-scratch/1days", type="string"),
 }
@@ -49,11 +48,10 @@ def genie_nf_validate_dag():
             work_dir=context["params"]["work_dir"],
             profiles=[context["params"]["profile"]],
             workspace_secrets=["SYNAPSE_AUTH_TOKEN"],
-            params_yaml=f"""
-                only_validate: {context["params"]["only_validate"]}
-                production: {context["params"]["production"]}
-                release: {context["params"]["release"]}
-                """,
+            params={
+                "process_type": context["params"]["process_type"],
+                "release": context["params"]["release"]
+            },
         )
         run_id = hook.ops.launch_workflow(
             info, context["params"]["tower_compute_env_type"], ignore_previous_runs=True
--- a/dags/genie-nf-validate.py
+++ b/dags/genie-nf-validate.py
@@ -8,7 +8,7 @@ from orca.services.nextflowtower.models import LaunchInfo
 
 dag_params = {
     "tower_conn_id": Param("GENIE_BPC_PROJECT_TOWER_CONN", type="string"),
-    "tower_compute_env_type": Param("spot", type="string"),
+    "tower_compute_env_type": Param("ondemand", type="string"),
     "tower_run_name": Param("airflow-genie-validate", type="string"),
     "pipeline": Param("Sage-Bionetworks-Workflows/nf-genie", type="string"),
     "revision": Param("main", type="string"),


rxu17:
It seems like nextflow takes in parameters that doesn't exist in the `nextflow.config` and throws no error? Because `production` was no longer a parameter for `nf-genie` for a while but our genie airflow DAG was still passing it to nextflow tower and tower was still taking it in (not doing anything with it though).

Wonder if we need something in either in `nf-genie` or the `py-orca` or here in this repo to alert us to unused `params` in tower hmm...

thomasyu888 (approved):
🔥 LGTM! 

As for using parameters that no longer exist... What happens when you do that locally?  I wonder why the schema validation doesn't cause the workflow to error out.

dags/genie-nf-validate.py
```diff
@@ -49,11 +48,10 @@ def launch_nf_genie_on_tower(**context):
             work_dir=context["params"]["work_dir"],
             profiles=[context["params"]["profile"]],
             workspace_secrets=["SYNAPSE_AUTH_TOKEN"],
-            params_yaml=f"""
-                only_validate: {context["params"]["only_validate"]}
-                production: {context["params"]["production"]}
-                release: {context["params"]["release"]}
-                """,
+            params={
+                "process_type": context["params"]["process_type"],
```
thomasyu888, line 52: Nit: By default `process_type` is `only_validation`, but i think it's ok to explicitly specify it here.

rxu17: Yeah I chose to specify it here in case the default changes in the future

rxu17:
When I run `nf-genie` locally with parameters that are not part of the valid set of params, it throws no errors or warnings:

 
image

rxu17:
`genie-bpc-pipeline` does throw out a warning when specifying nonexisting parameters like here:


image


I wonder if we should just have it throw an error instead of warning by modifying the code here: https://github.com/Sage-Bionetworks/genie-bpc-pipeline/blob/f3c010d8f7188db86ae7e1559c82547e12c9f7c3/lib/NfcoreSchema.groovy#L169-L179. Anyways, it seems like something we have to add to the nextflow side of our pipelines (and probably easier to do it from the nextflow side)

rxu17:
Should [LaunchInfo](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/main/src/orca/services/nextflowtower/models.py#L153-L169) have something to catch when a LaunchInfo object is created and things get defined that are not part of the launch specification?

thomasyu888:
@rxu17 all great ideas.

1. I think we can have a ticket to add in an error if invalid parameters are specified for both main and bpc nextflow pipelines.
2. When you say launchInfo, are you taking about the class itself? I think there might be pydantic validation on the class ( I could be wrong)?

rxu17:
> When you say launchInfo, are you taking about the class itself? I think there might be pydantic validation on the class ( I could be wrong)?

Yes. Ah I see ... I just tested out a sample data class to mimic LaunchInfo, and it looks like can take extra class attributes that's not defined in the class itself without throwing a warning or error. And since `params` is optional in `LaunchInfo`, it's okay it was not defined which is why this just slipped by.

dags/genie-nf-validate.py
```diff
@@ -13,8 +13,7 @@
     "pipeline": Param("Sage-Bionetworks-Workflows/nf-genie", type="string"),
     "revision": Param("main", type="string"),
     "profile": Param("aws_prod", type="string"),
-    "only_validate": Param("true", type="string"),
-    "production": Param("true", type="string"),
+    "process_type": Param("only_validate", type="string"),
     "release": Param("13.3-consortium", type="string"),
     "work_dir": Param("s3://genie-bpc-project-tower-scratch/1days", type="string"),
```
thomasyu888, line 18: For some reason, we are also using the spot instance as the default compute env?  can we use on-demand for robustness?

PR #24, opened 2023-07-28
[ORCA-260] Adds External Recipes

BWMac:
This PR adds the following recipes from external repositories:

- NTAP from [`ntap-add5-scripts`](https://github.com/Sage-Bionetworks-Workflows/ntap-add5-scripts)
- iAtlas from [`iatlas-scripts`](https://github.com/Sage-Bionetworks-Workflows/iatlas-scripts)

They are organized within the existing `local` directory by project. The LENS recipe (work in progress) was already present within this directory and is now in the `local/iatlas` directory along with `immune_subtype_classifier`. I have also set up these directories so that each project shares a `README.md` file.

I gave the workflows org a once-over and did not see any other obvious candidates to be moved over here at this time. I would be happy to take suggestions for any further recipes that should be present in this repo.


thomasyu888 (approved):
💯 LGTM!

PR #26, opened 2023-09-29
[ORCA-273] Adds `nf-dynamic-challenge` Nextflow Tower DAG

BWMac:
This PR contributes `nf-dynamic-challenge-dag.py` which follows a similar structure to other Nextflow Tower DAGs in the repo.

This new DAG does the following:
- Takes parameters defined by the `nf-dynamic-challenge` workflow (`view_id`, `cpus`, and `memory`)
- Connects to the production instance of Nextflow Tower (`dynamic-challenge-project`)
- Launches `nf-dynamic-challenge` using an example [file view](https://www.synapse.org/#!Synapse:syn52576179/tables/) I created for testing
- Monitors the workflow until its completion

This DAG has been tested in GitHub Codespaces and the Nextflow Workflow completed successfully in Tower.

### DIFF
--- a/dags/nf-dynamic-challenge-dag.py
+++ b/dags/nf-dynamic-challenge-dag.py
@@ -1,0 +1,54 @@
+from datetime import datetime
+
+from airflow.decorators import dag, task
+from airflow.models import Param
+
+from orca.services.nextflowtower import NextflowTowerHook
+from orca.services.nextflowtower.models import LaunchInfo
+
+
+dag_params = {
+    "tower_conn_id": Param("EXAMPLE_DEV_PROJECT_TOWER_CONN", type="string"),
+    "tower_run_name": Param("nf-dynamic-challenge-test", type="string"),
+    "tower_compute_env_type": Param("spot", type="string"),
+}
+
+dag_config = {
+    "schedule_interval": None,
+    "start_date": datetime(2023, 6, 1),
+    "catchup": False,
+    "default_args": {
+        "retries": 2,
+    },
+    "tags": ["nextflow_tower"],
+    "params": dag_params,
+}
+
+
+@dag(**dag_config)
+def nf_dynamic_challenge_dag():
+    @task()
+    def launch_nf_dynamic_challenge_on_tower(**context):
+        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
+        info = LaunchInfo(
+            run_name=context["params"]["tower_run_name"],
+            pipeline="Sage-Bionetworks-Workflows/nf-dynamic-challenge",
+            revision=" bwmac/orca-273/update_workflow_inputs",
+        )
+        run_id = hook.ops.launch_workflow(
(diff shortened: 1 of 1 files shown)


dags/nf-dynamic-challenge-dag.py
```diff
@@ -0,0 +1,61 @@
...
+        run_id = hook.ops.launch_workflow(
+            info, context["params"]["tower_compute_env_type"]
+        )
+        return run_id
+
+    @task.sensor(poke_interval=300, timeout=604800, mode="reschedule")
+    def monitor_nf_hello_workflow(run_id: str, **context):
+        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
+        workflow = hook.ops.get_workflow(run_id)
+        print(f"Current workflow state: {workflow.status.state.value}")
```
BryanFauble, line 54: Would it make sense to swap to using the logging package to log into at an info level? For example this is similar to format's I've used before when we ingested logs into Elasticsearch and wanted to filter logs nicely:

```
import logging
import json

# Create a logger and set log level
logger = logging.getLogger("My new logger")
logger.setLevel(logging.DEBUG)

# Create a console handler and set its log level to INFO
console_handler = logging.StreamHandler()
console_handler.setLevel(logging.DEBUG)

# Create a formatter (JSON) and add it to the console handler
formatter = logging.Formatter("%(message)s")
json_formatter = logging.Formatter(json.dumps({
    "timestamp": "%(asctime)s",
    "logger": "%(name)s",
    "level": "%(levelname)s",
    "message": "%(message)s"
}))
console_handler.setFormatter(json_formatter)

# Add the console handler to the logger to allow printing message to console
logger.addHandler(console_handler)


logger.info("My first log message")
logger.debug("My second log message")
```

Output:
```
{"timestamp": "2023-09-29 12:38:58,655", "logger": "My new logger", "level": "INFO", "message": "My first log message"}
{"timestamp": "2023-09-29 12:38:58,656", "logger": "My new logger", "level": "DEBUG", "message": "My second log message"}
```

BryanFauble: Ah - I think I found my answer in the logs you provided you ORCA-273:
```
[2023-09-29T18:00:01.930+0000] {logging_mixin.py:150} INFO - Current workflow state: RUNNING
```
It looks like it is already handling the pretty-print of the messages :D

BryanFauble (approved):
LGTM!

dags/nf-dynamic-challenge-dag.py
BWMac: Right, the logging is mostly handled by logic in our orchestrator package, [`py-orca`](https://github.com/Sage-Bionetworks-Workflows/py-orca). IMHO, if we want to add more detailed logging it should also live there so that it is seamlessly applied to all DAGs which use that package.

thomasyu888 (approved):
🔥 LGTM! Thanks for doing this.

PR #27, opened 2023-10-18
[ORCA-281] Updates Dynamic Challenge DAG

BWMac:
This PR updates the `nf-dynamic-challenge` DAG to use our new challenge monorepo Nextflow workflow, `nf-synapse-challenge` and configures it to use the `DATA_TO_MODEL_CHALLENGE` entrypoint. It also adds the submission view provided by the challenge team.

In order to enable this work, I needed to upgrade `py-orca` to its latest version so that we could leverage the `entry_name` parameter in airflow/metaflow DAGs. 

I also pinned the version of `apache-airflow` we are using to `2.7.2` and added the `apache-airflow-providers-celery` dependency which is needed for the newer airflow version.

### DIFF
--- a/dags/nf-dynamic-challenge-dag.py
+++ b/dags/nf-dynamic-challenge-dag.py
@@ -9,9 +9,9 @@ from orca.services.nextflowtower.models import LaunchInfo
 
 dag_params = {
     "tower_conn_id": Param("DYNAMIC_CHALLENGE_PROJECT_TOWER_CONN", type="string"),
-    "tower_run_name": Param("nf-dynamic-challenge-test", type="string"),
+    "tower_run_name": Param("dynamic-challenge-evaluation", type="string"),
     "tower_compute_env_type": Param("spot", type="string"),
-    "challenge_view_id": Param("syn52576179", type="string"),
+    "challenge_view_id": Param("syn52658661", type="string"),
     "tower_cpus": Param("4", type="string"),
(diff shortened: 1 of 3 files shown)


dags/nf-dynamic-challenge-dag.py
```diff
@@ -35,12 +35,13 @@ def launch_nf_dynamic_challenge_on_tower(**context):
...
         info = LaunchInfo(
             run_name=context["params"]["tower_run_name"],
-            pipeline="Sage-Bionetworks-Workflows/nf-dynamic-challenge",
+            pipeline="Sage-Bionetworks-Workflows/nf-synapse-challenge",
             revision="main",
             params={
                 "view_id": context["params"]["challenge_view_id"],
             },
             workspace_secrets=["SYNAPSE_AUTH_TOKEN"],
+            entry_name="DATA_TO_MODEL_CHALLENGE",
```
thomasyu888, line 44: Should this be dynamic challenge?

thomasyu888 (approved):
🔥 Going to pre-approve, but LGTM! Just a small comment

dags/nf-dynamic-challenge-dag.py
BWMac: This is the entrypoint for all data-to-model challenges. My hope is that we can use the same subworkflow for more than one individual challenge. This is just the name of that subworkflow.

PR #30, opened 2023-11-21
[ORCA-291] LENS DAG

BWMac:
***This PR is a WIP***

**Problem:**

We need a way to reproducibly run the LENS workflow for the iAtlas project.

**Solution:**

A Metaflow DAG to be run locally that will follow a flow of `nf-synstage` -> `lens_for_nf_tower` -> `nf-synindex`.

This Metaflow DAG is based on the [example script](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/main/demo.py) in `py-orca`.


BryanFauble (approved):
Changes make sense to me!

If you run into issues with the synindex & the DLS1/2 issue this is what I did in another flow: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/local/ntap/workflows-538-rnaseq-mouse.py#L115

PR #31, opened 2023-12-04
Adding single-cell rnaseq workflow

BryanFauble:
**Problem:**
We did not have a single-cell RNASEQ pipeline created for human.

**Solution:**

1. Implementing the `nf-core/scrnaseq` workflow. 
2. Additionally there was some complexity around the cellranger pipeline due to file names. Because of this it requires that the boto3 client is installed and configured on your local machine. This allows us to rename files in S3 and modify the staged CSV file.

**Testing:**
I ran the workflow up until we found out that the data was not single-cell. However, this should work for single-cell data and is a good starting point when we implement this for a single-cell set of data.
https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/ntap-add5-project/watch/4QUHCFXd9vKhlh

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,6 +16,7 @@ s3fs = "~=2023.5"
 metaflow = "~=2.9"
 fs-synapse = "~=1.0"
 synapseclient = "~=2.7.2"
+boto3 = ">=1.7.0,<2.0"
 
 [requires]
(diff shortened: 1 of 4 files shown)


BWMac (approved):
LGTM

PR #33, opened 2023-12-08
[ORCA-292] Pins `connexion` Version

BWMac:
**Problem:**

It was recently reported through many issues on the `airflow` GitHub repo ([example](https://github.com/apache/airflow/issues/35434)) that the release of `connexion` 3 has broken airflow deployments. This was causing our "production" airflow instance to fail to spin up in EC2, and also prevented development versions from running in GH codespaces.

**Solution:**

Pin the version of `connexion` in our `pipfile`. This is not a perfect long-term solution, but it should enable our airflow instance to run again. To better prevent this from happening in the future, we should implement constraints as described [here](https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html). I will create a separate ticket for adding such constraints to our airflow deployment.

Notes:
- I also went ahead and bumped up the airflow version used in our `Dockerfile` to match the version that we are actually spinning up and running for consistency.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE_IMAGE=apache/airflow:2.5.1-python3.10
+ARG BASE_IMAGE=apache/airflow:2.7.2-python3.10
 FROM $BASE_IMAGE
 
 RUN pip install --no-cache-dir pipenv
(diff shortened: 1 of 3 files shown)


BryanFauble (approved):
LGTM!

PR #34, opened 2024-01-10
[ORCA-290] Synapse Portal Data to Snowflake DAG

BWMac:
**Problem:**

We want to upload data from the Synapse entities that function as the backend for many data portals to Snowflake tables, but currently, we have to do so manually.

**Solution:**

Create an Airflow DAG that downloads portal data from Synapse, and upserts it to the appropriate database tables in Snowflake.

Notes:
- Much of this logic was adapted from this [script](https://github.com/Sage-Bionetworks/snowflake/blob/dev/sage/portal_elt.py).
- I tried to have `portal_dict` included as an Airflow `Param` using the `object` type, but it led to instability with the `scheduler`. We should look into JSON configurations for this DAG and others in the future.
-  The DAG is currently written to be triggered every Monday at midnight server time, I am happy to change this if there is a more preferred day/time/cadence.

Edit: I have updated this PR to use an overwrite method for the database tables instead of an upsert. This is because if any schema changes were to occur, it would break the pipeline. A future task will need to be completed to allow for the upsert method while allowing for schema changes ([ticket](https://sagebionetworks.jira.com/browse/ORCA-298)).

### DIFF
--- a/dags/portal-data-to-snowflake-dag.py
+++ b/dags/portal-data-to-snowflake-dag.py
@@ -1,0 +1,58 @@
+from datetime import datetime
+
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
+
+dag_params = {
+    "snowflake_conn_id": Param(
+        "SNOWFLAKE_DATA_ENGINEER_PORTAL_RAW_CONN", type="string"
+    ),
+    # "portal_dict": Param(
+    #     {
+    #         "AD": "syn11346063",
+    #         "PSYCHENCODE": "syn20821313.16",
+    #         "NF": "syn16858331",
+    #         "ELITE": "syn51228429",
+    #         "HTAN": "syn52748752",
+    #         "GENIE": "syn52794526",
+    #     },
+    #     type="dict",
+    # ),
+}
+
+portal_dict = {
+    "AD": "syn11346063",
+    "PSYCHENCODE": "syn20821313.16",
+    "NF": "syn16858331",
+    "ELITE": "syn51228429",
+    "HTAN": "syn52748752",
+    "GENIE": "syn52794526",
+}
+
+dag_config = {
+    "schedule_interval": "0 0 * * 1",
+    "start_date": datetime(2023, 2, 21),
+    "catchup": False,
+    "default_args": {
+        "retries": 1,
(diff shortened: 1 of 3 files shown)


dags/portal-data-to-snowflake-dag.py
```diff
@@ -0,0 +1,132 @@
...
+                overwrite=True,
+                quote_identifiers=False,
+            )
+            merge_sql = prepare_merge_sql(
+                portal_name=portal_name,
+                portal_df=info["data"],
+                temp_table=temp_table,
+                id_col=info["id_col"],
+            )
+            # TODO account for schema changes
```
BryanFauble: Does anything need to be done about this at the moment, or is there a follow up jira to address this?

BryanFauble (approved):
LGTM!

dags/portal-data-to-snowflake-dag.py
BWMac: This is a note left by @thomasyu888 from the original script that he was running manually to perform this work. I left this in the DAG to be sure it was addressed if needed in the future. I don't know if there is a Jira ticket.

thomasyu888: @BWMac I think this script will currently fail if there are column changes in any of the tables.  

I forgot but thats actually the reason why I switched to just overwrite the tables in snowflake instead of an upsert.  Lets create a ticket to either switch to overwriting the table or supporting schema changes.

BWMac: @thomasyu888 Should this new ticket be completed as part of this PR before merging?

BryanFauble: Probably afterword? But we would need to be clear that the created ticket would be a blocker for any future schema changes.

BWMac: Per slack DMs with @thomasyu888 I am going to go ahead and do the overwrite method for now since it is simple and then we can create a ticket for supporting upserts + schema changes in the future.

BWMac: I have updated the DAG to use overwriting instead, and I created a [ticket](https://sagebionetworks.jira.com/browse/ORCA-298) for updating this to upsert and allow for schema changes.

thomasyu888 (approved):
🔥 LGTM!

PR #36, opened 2024-02-20
[ORCA-301] Top 10 project downloads

BryanFauble:
**Problem:**

1. We want to report out the top downloaded public synapse projects to slack.

**Solution:**

1. Create an efficient SQL query to execute on snowflake, and then post the results to slack.

**Testing:**

1. See the #topcharts slack channel

**Additional notes:**

1. I tried to use the official airflow slack operator, but it never worked properly and would never tell me the exact reason why it couldn't work properly. I ended up just using the slack SDK instead and it worked perfectly.

### DIFF
--- a/dags/top-downloads-from-sowflake.py
+++ b/dags/top-downloads-from-sowflake.py
@@ -1,5 +1,6 @@
 """This script is used to execute a query on Snowflake and report the results to a
-slack channel. See ORCA-301 for more context."""
+slack channel. This retrieved the top X publicly downloaded Synapse projects.
+See ORCA-301 for more context."""
 
 from datetime import datetime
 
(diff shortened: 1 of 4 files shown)


dags/top-downloads-from-sowflake.py
```diff
@@ -0,0 +1,117 @@
...
+def snowflake_top_downloads_to_slack():
+    @task
+    def get_top_downloads_from_snowflake(**context) -> str:
+        snow_hook = SnowflakeHook(context["params"]["snowflake_conn_id"])
+        ctx = snow_hook.get_conn()
+        cs = ctx.cursor()
+        cs.execute(QUERY)
+        top_downloaded_df = cs.fetch_pandas_all()
+
+        message = ":synapse: Top Downloaded Public Synapse Projects Yesterday!\n\n"
```
thomasyu888: Nit: did we want to include the fact that it's yesterday UTC time?

BryanFauble: Like:
```suggestion
        message = ":synapse: Top Downloaded Public Synapse Projects Yesterday (UTC)!\n\n"
```
?

BWMac (approved):
LGTM

PR #38, opened 2024-02-26
[ORCA-303] Push results to synapse table

BryanFauble:
**Problem:**

1. We want to push a non-limited query to a Synapse table so that it can be used elsewhere.

**Solution:**

1. Created a table in Synapse: https://www.synapse.org/#!Synapse:syn53696951/tables/ - With this schema:
image

2. Created a query that retrieves a non-limited query from Snowflake and total data size in bytes - I chose bytes in this case because GiB was causing a lot of data to be truncated to `0` automatically by Synapse
3. I also needed to update the API Token of the ORCA service account with modify - I did this and pushed the new token into the AWS secrets

**Testing:**

1. Verified through a GH Codespace that I could run the Airflow DAG and push results into the table:

image

image

### DIFF
--- a/dags/top-downloads-from-sowflake.py
+++ b/dags/top-downloads-from-sowflake.py
@@ -3,18 +3,20 @@ slack channel. This retrieved the top X publicly downloaded Synapse projects.
 See ORCA-301 for more context."""
 
 from dataclasses import dataclass
-from datetime import datetime
+from datetime import date, datetime, timedelta
 from typing import List
 
+import synapseclient
 from airflow.decorators import dag, task
+from airflow.models import Variable
 from airflow.models.param import Param
 from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
-from airflow.models import Variable
+from orca.services.synapse import SynapseHook
 from slack_sdk import WebClient
-
 
 dag_params = {
     "snowflake_conn_id": Param("SNOWFLAKE_SYSADMIN_PORTAL_RAW_CONN", type="string"),
+    "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
 }
 
 dag_config = {
@@ -33,7 +35,9 @@ BYTE_STRING = "GiB"
 # 30 is the power of 2 for GiB, 40 is the power of 2 for TiB
 POWER_OF_TWO = 30
 
-QUERY = f"""
+SYNAPSE_RESULTS_TABLE = "syn53696951"
+
+QUERY_WITH_LIMIT = f"""
 WITH PUBLIC_PROJECTS AS (
     SELECT
         node_latest.project_id,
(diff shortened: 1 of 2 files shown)


dags/top-downloads-from-sowflake.py
```diff
@@ -163,11 +233,37 @@ def post_top_downloads_to_slack(message: str) -> str:
...
+                    metric.data_download_size,
+                ]
+            )
+
+        syn_hook = SynapseHook(context["params"]["synapse_conn_id"])
+        syn_hook.client.store(
+            synapseclient.Table(schema=SYNAPSE_RESULTS_TABLE, values=data)
+        )
+
+    top_downloads = get_top_downloads_from_snowflake(include_limit=True)
```
thomasyu888: Do you think it's worthwhile to get all the top results then filter for the top 10 when generating the slack message and upload all results the synapse table?  This way you just run one snow query.

That said, I'm also ok with this.

dags/top-downloads-from-sowflake.py
```diff
@@ -163,11 +233,37 @@ def post_top_downloads_to_slack(message: str) -> str:
...
 
     top_downloads >> slack_message >> post_to_slack
 
+    top_downloads = get_top_downloads_from_snowflake(include_limit=False)
+    push_to_synapse_table = push_results_to_synapse_table(metrics=top_downloads)
+
+    top_downloads >> push_to_synapse_table
+
 
 snowflake_top_downloads_to_slack()
```
BWMac: You might want to change this name to match the name of the script or be more indicative of what this DAG does now because this is the name that shows up in the Airflow UI.

BWMac (approved):
Pre-approving, but one comment

thomasyu888 (commented):
🔥 LGTM!

dags/top-downloads-from-sowflake.py
```diff
@@ -92,6 +96,64 @@
...
+    INNER JOIN
+        PUBLIC_PROJECTS
+    ON
+        filedownload.project_id = PUBLIC_PROJECTS.project_id
+    INNER JOIN
+        synapse_data_warehouse.synapse.file_latest
+    ON
+        filedownload.file_handle_id = file_latest.id
+    WHERE
+        filedownload.record_date = DATEADD(HOUR, -24, CURRENT_DATE)
```
thomasyu888: Just thinking ahead, but if the DAG ever fails or if Airflow goes down, do we want a "more robust" way to backfill information?

dags/top-downloads-from-sowflake.py
```diff
@@ -163,11 +233,37 @@ def post_top_downloads_to_slack(message: str) -> str:
         result = client.chat_postMessage(channel="topcharts", text=message)
         return result
 
-    top_downloads = get_top_downloads_from_snowflake()
```
thomasyu888, line 166: Not related to this line, but let's remove the redundant byte string on the slack message

dags/top-downloads-from-sowflake.py
BryanFauble: I updated this to be `top_public_synapse_projects_from_snowflake `

dags/top-downloads-from-sowflake.py
BryanFauble: Yes - Great idea. We can execute a single query this way. I verified that with the change I still generated the correct download stats:
https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/38/commits/701f9c0d0762bb0c8e09cdedefb2afa2ae7f3628

```
1. <https://www.synapse.org/#!Synapse:syn2759792|CommonMind Consortium Knowledge Portal> - 6093 downloads, 3 unique users, 9257.699 GiB egressed
2. <https://www.synapse.org/#!Synapse:syn51364943|UKB-PPP> - 5621 downloads, 16 unique users, 2725.256 GiB egressed
3. <https://www.synapse.org/#!Synapse:syn2580853|AD Knowledge Portal - backend> - 2500 downloads, 15 unique users, 6851.613 GiB egressed
4. <https://www.synapse.org/#!Synapse:syn4993293|mPower Public Researcher Portal> - 2009 downloads, 3 unique users, 2.658 GiB egressed
5. <https://www.synapse.org/#!Synapse:syn4921369|PsychENCODE Knowledge Portal - backend> - 1930 downloads, 4 unique users, 4064.707 GiB egressed
6. <https://www.synapse.org/#!Synapse:syn3193805|Multi-Atlas Labeling Beyond the Cranial Vault - Workshop and Challenge> - 1008 downloads, 20 unique users, 310.150 GiB egressed
7. <https://www.synapse.org/#!Synapse:syn49637038|SINGLE-CELL ATLAS OF HUMAN BLOOD DURING HEALTHY AGING> - 178 downloads, 2 unique users, 7681.545 GiB egressed
8. <https://www.synapse.org/#!Synapse:syn20446507|HTAN All Projects> - 92 downloads, 3 unique users, 1.244 GiB egressed
9. <https://www.synapse.org/#!Synapse:syn1773109|Progenitor Cell Biology Consortium (PCBC)> - 47 downloads, 4 unique users, 0.455 GiB egressed
10. <https://www.synapse.org/#!Synapse:syn50896946|FAIR Demo All Projects> - 39 downloads, 1 unique users, 0.001 GiB egressed
One download is a user downloading an entity (File, Table, Views, etc) once
```

image

dags/top-downloads-from-sowflake.py
BryanFauble: Good idea - I added in a time delta parameter that can be passed into the DAG with 24 as the default. This allows us to back-fill previous days.

image

Since this would also post to slack (Maybe we want to disable in the future) I updated the title of the message to be either `Yesterday` or the specific date if > 24 hours time delta.
```
Top Downloaded Public Synapse Projects Yesterday!
Top Downloaded Public Synapse Projects 2024-03-02!
```

BWMac (approved):

PR #39, opened 2024-03-07
[ORCA-302] Updates Dynamic Challenge DAG for Full Automation

BWMac:
**Description:**

To provide automated challenge submission evaluation for the Dynamic challenge, we needed an Airflow DAG that would run at a regular cadence and send new submissions to Nextflow Tower for evaluation. The `dynamic-challenge-dag` has now been updated to perform the following tasks:

1. `get_new_submissions`: Retrieve a list of the new submissions with a table query.
2. `update_submission_statuses`: Update submission statuses to `EVALUATION_IN_PROGRESS`. If the list of submissions is empty (no new submissions), the DAG stops here.
3. `stage_submissions_manifest`: Create a CSV file manifest with the submission IDs. Stage it in an S3 bucket accessible by Tower.
4. `launch_data_to_model_on_tower`: Launch the `nf-synapse-challenge` workflow run on Tower by passing it the manifest's S3 path.
5.  `monitor_workflow`: Monitor workflow run on Tower.

**Notes:**
- I was able to bring the scheduling interval down from 2 minutes to 1 minute after we eliminated the need for the second table query.

**Testing:**

- [x] [Run](https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/dynamic-challenge-project/watch/1EuByn4b8vV2GX) on 100 submissions at once
- [x] [Run](https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/dynamic-challenge-project/watch/1YAypr9m5EaJvL) on 10 Batches of 10 submissions. Batches were submitted every 120 seconds.
- [x] [Run](https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/dynamic-challenge-project/watch/28ZkiiTyAlwwHI) on 20 batches with random numbers of submissions. Batches were submitted at random intervals between 15 and 120 seconds.


Pipfile
```diff
@@ -10,10 +10,11 @@ apache-airflow-providers-celery = ">=3.0"
 apache-airflow-providers-snowflake = "~=5.2"
 sagetasks = "~=0.4"
 connexion = "==2.14.2"
-py-orca = {version = "==1.3.3", extras = ["all"]}
+py-orca = {version = "==1.3.4", extras = ["all"]}
```
BryanFauble, line 13: I think something might be wrong here - Can you double check that you have main pulled into your branch. The pipfile in main looks like:
```
[packages]
apache-airflow = "==2.7.2"
apache-airflow-providers-amazon = "~=7.2"
apache-airflow-providers-celery = ">=3.0"
apache-airflow-providers-snowflake = "~=5.2"
sagetasks = "~=0.4"
connexion = "==2.14.2"
py-orca = {version = "==1.3.3", extras = ["all"]}
slack-sdk = ">=3.27"
# See https://stackoverflow.com/questions/77846645/airflow-db-migrate-is-giving-typeerror-sqlalchemysessioninterface-init-mi
flask-session = "==0.5.0"
```

Which doesn't match your before/after

dags/dynamic-challenge-dag.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@dag(**dag_config)
+def dynamic_challenge_dag():
+    @task.branch()
+    def check_for_new_submissions(**context):
+        hook = SynapseHook(context["params"]["synapse_conn_id"])
+        # Unfortunately, you cannot use XCom in a branching task
+        # So we have to call `get_submissions_with_status` twice in the DAG
+        submissions = hook.ops.get_submissions_with_status(
```
BryanFauble: @jaymedina Something I noticed in https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/main/src/orca/services/synapse/ops.py#L97 - The way this is set up:
`f"select * from {submission_view} where status = '{submission_status}'"` means that you are grabbing all columns of the table, but right after this you are only getting the ID column in the DF:

`submission_ids = query_results.asDataFrame()["id"].tolist()`

The select statement should change to:

`f"select id from {submission_view} where status = '{submission_status}'"`

The reason is that only the ID would then be transferred over HTTP, instead of all columns - It's minor, but best practice to only transfer the minimum amount of data required. Do you mind creating a tech debt item to tackle this change?

dags/dynamic-challenge-dag.py
```diff
@@ -0,0 +1,141 @@
...
+                submission_status="EVALUATION_IN_PROGRESS",
+            )
+
+    @task()
+    def stage_submissions_manifest(submissions: list, **context):
+        s3_hook = S3Hook(
+            aws_conn_id=context["params"]["aws_conn_id"], region_name=REGION_NAME
+        )
+        df = pd.DataFrame({"submission_id": submissions})
+        df.to_csv(FILE_NAME, index=False)
```
BWMac, line 81: I am realizing that this should probably be a temp file.

dags/dynamic-challenge-dag.py
```diff
@@ -0,0 +1,141 @@
...
+def dynamic_challenge_dag():
+    @task.branch()
+    def check_for_new_submissions(**context):
+        hook = SynapseHook(context["params"]["synapse_conn_id"])
+        # Unfortunately, you cannot use XCom in a branching task
+        # So we have to call `get_submissions_with_status` twice in the DAG
+        submissions = hook.ops.get_submissions_with_status(
+            context["params"]["view_id"], "RECEIVED"
+        )
+        if submissions:
```
BryanFauble: You should change the logic here - Return back the submissions in this step to prevent needing to call:
```
        submissions = hook.ops.get_submissions_with_status(
            context["params"]["view_id"], "RECEIVED"
        )
```

Twice. You are executing a second query to get data you've already retrieved.

BWMac: I can't do that. See the comment directly above this.

dags/dynamic-challenge-dag.py
BryanFauble: Alternatively after upload delete the local copy

dags/dynamic-challenge-dag.py
```diff
@@ -0,0 +1,141 @@
...
+    def update_submission_statuses(submissions: list, **context):
+        hook = SynapseHook(context["params"]["synapse_conn_id"])
+        for submission in submissions:
+            hook.ops.update_submission_status(
+                submission_id=submission,
+                submission_status="EVALUATION_IN_PROGRESS",
+            )
+
+    @task()
+    def stage_submissions_manifest(submissions: list, **context):
```
BryanFauble: nit: Return types of all functions

dags/dynamic-challenge-dag.py
BWMac: Maybe I could move the `stop_dag` task to after `update_submission_statuses` since it does not need to return anything. Then it would run a couple tasks with an empty list but we might be able to avoid the double query

BryanFauble: Gotcha - Does it prevent you from explicitly stashing the values and then pulling the values later on? I see these docs:
https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/xcoms.html#xcoms
```
# pushes data in any_serializable_value into xcom with key "identifier as string"
task_instance.xcom_push(key="identifier as a string", value=any_serializable_value)

# pulls the xcom variable with key "identifier as string" that was pushed from within task-1
task_instance.xcom_pull(key="identifier as string", task_ids="task-1")
```

BWMac: Yes, I tried that and it did not work.

BryanFauble (approved):
Awesome changes - Going to pre-approve this.

dags/dynamic-challenge-dag.py
BryanFauble: That.... just seems wrong and something that should work - Maybe we should put in an issue for what you found.

BWMac: I'll find out if there is one already

BWMac: > Maybe I could move the stop_dag task to after update_submission_statuses since it does not need to return anything. Then it would run a couple tasks with an empty list but we might be able to avoid the double query

Fortunately, this does work. I'll update my description to reflect this change

dags/dynamic-challenge-dag.py
jaymedina: Made the ticket, thanks @BryanFauble : https://sagebionetworks.jira.com/browse/IBCDPE-848

jaymedina (approved):

rrchai (approved):
LGTM 🔥

PR #41, opened 2024-04-03
[ORCA-316] Fix `snowflake-connector-python` version

BWMac:
**Problem:**

We recently updated our dependency management strategy for this repository. When we made this update, it appears that it caused and older version of `snowflake-connector-python` to be installed in the Airflow deployment which led to the top charts DAG hanging mid-run.

**Solution:**

Explicitly set the version of `snowflake-connector-python` in `requirements-airflow.txt`. I also went ahead and pinned the provider versions in `requirements-airflow.txt` like we used to in the old `pipfile`.

**Testing:**

- I ran the top charts DAG in the development environment and it [worked](https://sagebionetworks.slack.com/archives/C06KHDY55QR/p1712184891139389). 
- I also manually ran through all of our dependencies (`apache-airflow` + `requirements-airflow.txt` + `requirements-dev.txt`) installed in the Airflow deployment and local development environment to ensure that compatible package versions are being installed.

### DIFF
--- a/dev_setup.sh
+++ b/dev_setup.sh
@@ -7,7 +7,7 @@ source venv/bin/activate
 # Upgrade pip to latest version
 pip install --upgrade pip
 # Install airflow with constraints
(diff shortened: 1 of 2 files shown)


BryanFauble (approved):
LGTM!

PR #42, opened 2024-04-08
[ORCA-317] Final DAG updates for  AI Institute in Dynamic Systems Challenge

BWMac:
This PR makes a couple of final changes before the launch of the AI Institute in Dynamic Systems Challenge including:
1. Pin the version of `nf-synapse-challenge` used.
     - It occurred to me that there is a possibility for new changes to the `main` branch of `nf-synapse-challenge` to be introduced mid-challenge and for consistency purposes this seemed not great. If we needed to update `nf-synapse-challenge` for this challenge we would need to update the revision value here.
2. Update the DAG to use the `dynamic_challenge` profile recently introduced to `nf-synapse-challenge`. This configures all of the python scripts needed for the workflow run.
3. Add the `send_email` parameter to the run configuration. `send_email` is a boolean set to `true` in the workflow by default, but making this explicit in the DAG is good for transparency and also adds the ability to toggle this from the DAG configuration.


jaymedina (approved):
LGTM 🚀

PR #43, opened 2024-04-09
[SNOW-66] All Time Download Metrics DAG

BWMac:
**Description:**

This PR adds a new DAG `top_public_synapse_projects_all_time_from_snowflake` which leverages Snowflake to get all time download metrics for all public Synapse projects and load them into a Synapse [Table](https://www.synapse.org/#!Synapse:syn55259224/tables/). The DAG is scheduled to run on the first day of each month.

This DAG and the Snowflake query it uses are adapted from the existing `top_public_synapse_projects_from_snowflake` DAG.


BryanFauble (approved):
🔥 LGTM!

PR #44, opened 2024-04-09
[SNOW-65] Past 30 Days Download Metrics DAG

BWMac:
**Description:**

This PR adds a new DAG `top_public_synapse_projects_30_days_from_snowflake` which leverages Snowflake to get download metrics from the past 30 days for all public Synapse projects and load them into a Synapse [Table](https://www.synapse.org/#!Synapse:syn55382267/tables/). The DAG is scheduled to run every day.

This DAG and the Snowflake query it uses are adapted from the existing `top_public_synapse_projects_from_snowflake` DAG.

### DIFF
--- a/dags/top-public-synapse-projects-30-days-from-snowflake.py
+++ b/dags/top-public-synapse-projects-30-days-from-snowflake.py
@@ -16,6 +16,7 @@ from orca.services.synapse import SynapseHook
 dag_params = {
     "snowflake_conn_id": Param("SNOWFLAKE_SYSADMIN_PORTAL_RAW_CONN", type="string"),
     "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
+    "current_date": Param(date.today().strftime("%Y-%m-%d"), type="string"),
 }
 
 dag_config = {
@@ -62,7 +63,7 @@ def top_public_synapse_projects_30_days_from_snowflake() -> None:
         snow_hook = SnowflakeHook(context["params"]["snowflake_conn_id"])
         ctx = snow_hook.get_conn()
         cs = ctx.cursor()
-        query = """
+        query = f"""
             WITH PUBLIC_PROJECTS AS (
                 SELECT
                     node_latest.project_id,
@@ -92,9 +93,9 @@ def top_public_synapse_projects_30_days_from_snowflake() -> None:
                 ON
                     filedownload.file_handle_id = file_latest.id
                 WHERE
-                    filedownload.record_date > DATEADD(DAY, -30, CURRENT_DATE)
+                    filedownload.record_date > DATEADD(DAY, -30, '{context["params"]["current_date"]}')
                 AND 
-                    filedownload.record_date <= CURRENT_DATE
+                    filedownload.record_date <= '{context["params"]["current_date"]}'
             ),
 
             DOWNLOAD_STAT AS (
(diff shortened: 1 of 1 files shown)


dags/top-public-synapse-projects-30-days-from-snowflake.py
```diff
@@ -0,0 +1,170 @@
...
+                INNER JOIN
+                    PUBLIC_PROJECTS
+                ON
+                    filedownload.project_id = PUBLIC_PROJECTS.project_id
+                INNER JOIN
+                    synapse_data_warehouse.synapse.file_latest
+                ON
+                    filedownload.file_handle_id = file_latest.id
+                WHERE
+                    filedownload.record_date > DATEADD(DAY, -30, CURRENT_DATE)
```
BryanFauble: It might be worthwhile to allow "current_date" to be passed into this DAG. If this misses a day and we want to back-fill the data for the day that was missed we currently do not have a way of doing this.

If we passed the configuration into the DAG we could execute it for any arbitrary 30 day period. It would need to be a change here, and the today var you defined in `push_results_to_synapse_table`

dags/top-public-synapse-projects-30-days-from-snowflake.py
```diff
@@ -0,0 +1,170 @@
...
+    name: str
+    project: str
+    downloads_per_project: int
+    number_of_unique_users_downloaded: int
+    data_download_size: float
+
+
+@dag(**dag_config)
+def top_public_synapse_projects_30_days_from_snowflake() -> None:
+    """Execute a query on Snowflake and report the results to a slack channel and
```
BryanFauble: remove comment about slack

dags/top-public-synapse-projects-30-days-from-snowflake.py
BryanFauble: I suppose the query would also need to be updated to only select those records `<= current date` and `> -30 days`

BryanFauble (approved):
Just 1 comment about back-filling data. Will go ahead and pre-approve here though.

dags/top-public-synapse-projects-30-days-from-snowflake.py
BWMac: I updated the DAG to have a parameterized `CURRENT_DATE`. 

> I suppose the query would also need to be updated to only select those records `<= current date` and `> -30 days`

Isn't that what I am already [doing](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/0f4324c58afd27c19bfdb3a275b8272c19b06d17/dags/top-public-synapse-projects-30-days-from-snowflake.py#L95-L99)?

BryanFauble: > I updated the DAG to have a parameterized `CURRENT_DATE`.
> 
> > I suppose the query would also need to be updated to only select those records `<= current date` and `> -30 days`
> 
> Isn't that what I am already [doing](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/0f4324c58afd27c19bfdb3a275b8272c19b06d17/dags/top-public-synapse-projects-30-days-from-snowflake.py#L95-L98)?

Not quite - In your query you are looking at all records that are: `filedownload.record_date > DATEADD(DAY, -30, '{context["params"]["current_date"]}')`

Suppose you run this query for last month (03/09/2024). Your query would answer the question:
"Get me all filedownloads that have a record date greater than 02/09/2024"

The data you are retrieving is going to include 2 months worth of data since it would include everything also collected between 03/09/2024 -> 04/09/2024.

By including a `<= current date` we are setting an upper (today) and lower (-30 days) bound on the data being selected.

BryanFauble: Ignore my comment - I looked back at the code and saw you are already doing this. I missed the `AND`:
```
                WHERE
                    filedownload.record_date > DATEADD(DAY, -30, '{context["params"]["current_date"]}')
                AND 
                    filedownload.record_date <= '{context["params"]["current_date"]}'
```

PR #45, opened 2024-04-10
Enable automated execution of PEGS challenge DAG

jaymedina:
Automatically run the PEGS challenge DAG every minute 🚀


dags/pegs-challenge-submission-dag.py
```diff
@@ -10,17 +10,17 @@
...
-    "tower_view_id": Param("syn53239158", type="string"),
+    "tower_view_id": Param("syn55253884", type="string"),
     "tower_input_id": Param("syn53239289", type="string"),
     "tower_compute_env_type": Param("spot", type="string"),
 }
 
 dag_config = {
-    "schedule_interval": None,
-    "start_date": datetime(2023, 6, 1),
+    "schedule_interval": "*/1 * * * *",
```
BWMac, line 32: I used `* * * * *` for my "run every minute" chron expression. I'm assuming this does the same thing, but just wanted to note it.

BWMac (changes requested):
This DAG won't work for running the workflow (other than with mostly default params including the default manifest) as it is currently written. We removed the part of `nf-synapse-challenge` that gets the new submissions for a challenge with the intention of having that become part of the automating Airflow DAGs, but you haven't added that here. See the dynamic challenge [DAG](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/dags/dynamic-challenge-dag.py) for an example of doing this by generating and providing a manifest.

dags/pegs-challenge-submission-dag.py
BryanFauble: @BWMac This is a site I reference often: `https://crontab.guru/#*/1_*_*_*_*`

BryanFauble (approved):
LGTM!

dags/pegs-challenge-submission-dag.py
BWMac: @BryanFauble Thanks for the link, that's a good reference to have in the back pocket. And yup it confirmed that both `*/1 * * * *` and `* * * * *` mean the same thing.

dags/pegs-challenge-submission-dag.py
```diff
@@ -1,26 +1,34 @@
...
+from airflow.providers.amazon.aws.hooks.s3 import S3Hook
 
 from orca.services.nextflowtower import NextflowTowerHook
 from orca.services.nextflowtower.models import LaunchInfo
 from orca.services.synapse import SynapseHook
 
+UUID = uuid.uuid4()
+REGION_NAME = "us-east-1"
+BUCKET_NAME = "pegs-challenge-project-tower-scratch"
+FILE_NAME = f"submissions_{UUID}.csv"
```
BryanFauble, line 17: nit: Just use UUID directly in this:
```suggestion
FILE_NAME = f"submissions_{uuid.uuid4()}.csv"
```

jaymedina: Thanks Bryan, just need to clarify something: @BWMac was this written as an environment variable in case it changes from `uuid4` in the future, or can it just be folded into the `FILE_NAME` string?

BWMac: I think I wrote it that way so I could reuse the `UUID` constant, but then I ended up only using it once

jaymedina: Got it, thanks for the quick reply

dags/pegs-challenge-submission-dag.py
```diff
@@ -1,26 +1,35 @@
...
+FILE_NAME = f"submissions_{uuid.uuid4()}.csv"
+KEY = "work"
 
 dag_params = {
     "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
-    "synapse_evaluation_id": Param("9615531", type="string"),
+    "synapse_evaluation_id": Param("9615511", type="string"),
+    "aws_conn_id": Param("AWS_TOWER_PROD_S3_CONN", type="string"),
     "tower_conn_id": Param("PEGS_CHALLENGE_PROJECT_TOWER_CONN", type="string"),
     "tower_run_name": Param("pegs_model_submission_evaluation", type="string"),
```
BWMac: Ah, sorry for the bad info yesterday. I forgot that I also use `UUID` [here](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/e6048d0989a6d413cc52f63a9d5f243466ebfc1b/dags/dynamic-challenge-dag.py#L26) so that each run has a unique run name which matches its manifest. That's why I had it set as a constant, I was using it twice. 

You should probably do the same or something similar so that runs don't get conflated with one another.

dags/pegs-challenge-submission-dag.py
```diff
@@ -32,25 +41,43 @@
...
+        df.to_csv(FILE_NAME, index=False)
+        s3_hook.load_file(
+            filename=FILE_NAME, key=f"{KEY}/{FILE_NAME}", bucket_name=BUCKET_NAME
+        )
+        os.remove(FILE_NAME)
+        return f"s3://{BUCKET_NAME}/{KEY}/{FILE_NAME}"
+
     @task()
     def launch_model2data_workflow(**context):
         hook = NextflowTowerHook(context["params"]["tower_conn_id"])
```
BWMac, line 84: You might also want to pass the Python scripts for validation, scoring, and emailing as parameters to the workflow run, or use the `pegs_challenge` profile just so its clear exactly what the workflow run is doing when the DAG triggers it.

dags/pegs-challenge-submission-dag.py
```diff
@@ -1,26 +1,35 @@
...
+from airflow.providers.amazon.aws.hooks.s3 import S3Hook
 
 from orca.services.nextflowtower import NextflowTowerHook
 from orca.services.nextflowtower.models import LaunchInfo
 from orca.services.synapse import SynapseHook
 
+REGION_NAME = "us-east-1"
+BUCKET_NAME = "pegs-challenge-project-tower-scratch"
+FILE_NAME = f"submissions_{uuid.uuid4()}.csv"
+KEY = "work"
```
BWMac: Using `work` is fine, but for my DAG I used `10days/<challenge_name>`. I liked this approach because then the manifests would be cleared after 10 days instead of 30 (the default for `scratch` bucket storage) and also it would be extra easy to find the manifests in one place if need be.

BWMac (commented):
One other thing I just thought of is that you might want to decrease the `poke_interval` in `monitor_model2data_workflow` so that submitters get their feedback a little quicker. I used [60](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/e6048d0989a6d413cc52f63a9d5f243466ebfc1b/dags/dynamic-challenge-dag.py#L109-L110) in my DAG.

dags/pegs-challenge-submission-dag.py
jaymedina: Right now the profile is configured to the real eval queue (not test) so I've given the individual parameters

BWMac: If you give it the profile and these parameters it will override the ones set in the profile. That being said, just passing all of the parameters is fine too

dags/pegs-challenge-submission-dag.py
```diff
@@ -1,26 +1,36 @@
...
 
 dag_params = {
     "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
-    "synapse_evaluation_id": Param("9615531", type="string"),
+    "synapse_evaluation_id": Param("9615511", type="string"),
+    "aws_conn_id": Param("AWS_TOWER_PROD_S3_CONN", type="string"),
     "tower_conn_id": Param("PEGS_CHALLENGE_PROJECT_TOWER_CONN", type="string"),
-    "tower_run_name": Param("pegs_model_submission_evaluation", type="string"),
-    "tower_view_id": Param("syn53239158", type="string"),
+    "tower_run_name": Param("pegs_model_submission_evaluation_{UUID}", type="string"),
```
BWMac:
```suggestion
    "tower_run_name": Param(f"pegs_model_submission_evaluation_{UUID}", type="string"),
```

dags/pegs-challenge-submission-dag.py
jaymedina: Noted: Once we're ready to go to prod, I won't be using parameters and will just be using the profile.

BWMac (approved):
LGTM 🔥

PR #46, opened 2024-04-11
hotfix: Fix KeyError in PEGS DAG

jaymedina:
The PEGS Challenge DAG was failing at the start due to a KeyError (no such key as 'view_id')


BryanFauble (approved):
LGTM!

PR #47, opened 2024-04-11
[IBCDPE-831] Updates to PEGS DAG

jaymedina:
### problem
A few parameters were missing in the params configuration, and `manifest_path` needed to be declared

### solution
Updated accordingly and tested on a codespace instance

### testing

PEGS challenge DAG is running successfully 

image

On the Nextflow side, here's an example of successful/running executions

image


BWMac (approved):
Pre-approving, but feel free to re-request review if any changes need to be made during stress testing!

jaymedina:
Leaving this in draft until stress test is over (April 19)

BWMac:
@jaymedina I realized while working on [ORCA-300](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/50) that using the UUIDs this way won't work exactly as intended. In the Dynamic Challenge DAG, my hope was to have the manifest file and workflow run ID have matching UUIDs so that it is easy to tell which file goes with which run, but it turns out that the way Airflow executes the DAGs causes different UUIDs to be used in different tasks sometimes. I fixed this in the linked PR by making the UUID a parameter of the DAG run so that it is always the same across tasks. I'll be updating the Dynamic Challenge DAG, but I wanted to give you a heads-up while this PR is still open for PEGS.

Edit: I made the update to the Dynamic Challenge DAG in #51, you should be able to do something similar here.

[ORCA-300]: https://sagebionetworks.jira.com/browse/ORCA-300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jaymedina:
Sounds good, thanks for the heads up @BWMac. Addressed in the latest commit.

dags/pegs-challenge-submission-dag.py
```diff
@@ -11,21 +11,22 @@
 from orca.services.nextflowtower.models import LaunchInfo
 from orca.services.synapse import SynapseHook
 
-UUID = uuid.uuid4()
 REGION_NAME = "us-east-1"
 BUCKET_NAME = "pegs-challenge-project-tower-scratch"
-FILE_NAME = f"submissions_{UUID}.csv"
+FILE_NAME = f"submissions.csv"
```
BWMac: NIT: no need for the `f`

dags/pegs-challenge-submission-dag.py
```diff
@@ -11,21 +11,22 @@
...
 KEY = "10days/pegs_challenge"
 
 dag_params = {
```
BWMac, line 19: There are a couple of parameters that aren't actually used (synapse_evaluation_id and tower_input_id). I'm assuming that's because they are included in the profile within the workflow. Should these be removed?

jaymedina: Yep thanks!

jaymedina:
Update: DAG is running successfully with the latest branch

image

dags/pegs-challenge-submission-dag.py
```diff
@@ -11,21 +11,20 @@
...
 BUCKET_NAME = "pegs-challenge-project-tower-scratch"
-FILE_NAME = f"submissions_{UUID}.csv"
+FILE_NAME = "submissions.csv"
 KEY = "10days/pegs_challenge"
 
 dag_params = {
     "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
-    "synapse_evaluation_id": Param("9615511", type="string"),
     "aws_conn_id": Param("AWS_TOWER_PROD_S3_CONN", type="string"),
+    "revision": Param("e19ca1dab3a85d77b62e8e00481d7291c19a0048", type="string"),
```
BryanFauble, line 22: Is there documentation around where this revision string is coming from?

BryanFauble (approved):
Just 1 comment, otherwise looks good to me!

dags/pegs-challenge-submission-dag.py
jaymedina: No, but that's a good point. I don't think there is any comprehensive piece of documentation outlining how `orca-recipes` and `nf-synapse-challenge` go together. I do have a ticket with a scope that this falls into, and I'll link this thread convo to that ticket so I don't forget to talk about params.

To keep it brief, this basically points to a branch, a version, or a commit ID of the pipeline that executes the NF workflow (in this case, `nf-synapse-challenge`)

jaymedina:
Here is the latest workflow run that runs off the latest ver of this branch. It's a successful workflow run following the handling of `profiles`. Apparently that's supposed to be a list, even if it's just a single profile. I also added the `tower` profile to work with `RUN_DOCKER` and use appropriate resource amounts.

https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/pegs-challenge-project/watch/1andK1qSuuzkQH

dags/pegs-challenge-submission-dag.py
BWMac: The parameter could also be renamed something more descriptive so its clear what it is referring to

PR #50, opened 2024-05-01
[ORCA-300] DCQC POC DAG

BWMac:
**Description:**

This PR adds a DAG that automatically runs DCQC on a manifest provided in Synapse. The DAG works by:

- Getting a list of all file entities in a Synapse parent location that match a manifest-specific suffix uploaded or modified in the past 24 hours.
- Checking how many new manifest entities were discovered. If 0, the DAG stops, if 1, the DAG continues, if more than 1, an error is raised.
- Staging the manifest in a Tower S3 bucket with a path specific to the DCQC run.
- Launching the DCQC workflow run with the staged manifest as input.
- Monitoring the workflow run.
- Retrieving the output file from S3, renaming it, and uploading it to the same Synapse location originally provided. This output file will always have the same name as written.

**Testing:**

- Successful Tower [run](https://tower-dev.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/example-dev-project/watch/yDZR2Lpoo6bbK)
- Synapse [location](https://www.synapse.org/#!Synapse:syn58807287) used

### DIFF
--- a/dags/dcqc-poc-dag.py
+++ b/dags/dcqc-poc-dag.py
@@ -84,7 +84,6 @@ def dcqc_poc_dag():
         s3_hook.load_file(
             filename=manifest_file.path, key=f"{KEY}/{run_uuid}/{manifest_file.name}", bucket_name=BUCKET_NAME
         )
-        print(f"s3://{BUCKET_NAME}/{KEY}/{run_uuid}/{manifest_file.name}")
         return f"s3://{BUCKET_NAME}/{KEY}/{run_uuid}/{manifest_file.name}"
 
     @task()
@@ -93,10 +92,8 @@ def dcqc_poc_dag():
         Launches nf-dcqc tower workflow
         """
         hook = NextflowTowerHook(context["params"]["tower_conn_id"])
-        run_name = context["params"]["tower_run_name"]
         parent = context["params"]["synapse_container"]
         run_uuid = context["params"]["uuid"]
-        print(f"{run_name}_{parent}_{run_uuid}")
         info = LaunchInfo(
             run_name=f"DCQC_{parent}_{run_uuid}",
             pipeline="Sage-Bionetworks-Workflows/nf-dcqc",
--- a/dags/dcqc-poc-dag.py
+++ b/dags/dcqc-poc-dag.py
@@ -35,9 +35,10 @@ dag_config = {
     "default_args": {
         "retries": 2,
     },
-    "tags": ["nextflow_tower_metrics"],
+    "tags": ["nextflow_tower"],
     "params": dag_params,
 }
+
 
 @dag(**dag_config)
 def dcqc_poc_dag():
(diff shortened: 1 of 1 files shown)


dags/dcqc-poc-dag.py
```diff
@@ -0,0 +1,182 @@
...
+def dcqc_poc_dag():
+    @task()
+    def get_new_manifest_entities(**context):
+        """
+        Gets a list of all manifest files in the synapse_container that have been uploaded in the last 24 hours.
+        """
+        syn_hook = SynapseHook(context["params"]["synapse_conn_id"])
+        container = syn_hook.client.get(context["params"]["synapse_container"])
+
+        if container.concreteType not in [
```
thomasyu888, line 54: nit: You can also do `isinstance(container, (Folder, Project))`, but this is the same thing while saving two imports.

dags/dcqc-poc-dag.py
```diff
@@ -0,0 +1,182 @@
...
+        ]:
+            raise ValueError(
+                f"Provided Synapse entity must be a Folder or Project, entity provided is '{container.concreteType}'"
+            )
+
+        child_entities = syn_hook.client.getChildren(container, includeTypes=["file"])
+        manifest_entities = [
+            child
+            for child in child_entities
+            if child["name"].endswith(context["params"]["manifest_suffix"])
```
thomasyu888, line 66: I like this, as long as people name their manifest: `{x}_dcqc_manifest.csv`, it will then pick it up

dags/dcqc-poc-dag.py
```diff
@@ -0,0 +1,182 @@
...
+        Check that there is only one manifest file in the synapse_container.
+        If 0, continue to the stop_dag task.
+        If 1, continue to the stage_manifest task.
+        If >1, raise an error.
+        """
+        if len(manifest_entities) == 1:
+            return "stage_manifest"
+        if len(manifest_entities) == 0:
+            return "stop_dag"
+        if len(manifest_entities) > 1:
```
thomasyu888, line 86: We may want to get the input of FAIR at sprint review about this, since it may not be ideal to limit to 1 manifest entity.

We can take that opportunity to get some feedback and create tickets

thomasyu888 (approved):
🔥 wow, a culmination of over a year's worth of work: piecing everything together in one ticket.  ORCA, DCQC, Airflow, what an accomplishment. Kudos!

dags/dcqc-poc-dag.py
```diff
@@ -0,0 +1,182 @@
...
+                f"Provided Synapse entity must be a Folder or Project, entity provided is '{container.concreteType}'"
+            )
+
+        child_entities = syn_hook.client.getChildren(container, includeTypes=["file"])
+        manifest_entities = [
+            child
+            for child in child_entities
+            if child["name"].endswith(context["params"]["manifest_suffix"])
+            and datetime.strptime(child["modifiedOn"], "%Y-%m-%dT%H:%M:%S.%fZ")
+            > datetime.now() - timedelta(days=1)
```
BryanFauble: You may need to get the `.now()` time in the UTC timezone for this to be exactly correct. Otherwise I believe this would use the timezone of the machine running the code, which may not be in UTC.

https://docs.python.org/3/library/datetime.html#datetime.datetime.now

dags/dcqc-poc-dag.py
BryanFauble: +1

dags/dcqc-poc-dag.py
```diff
@@ -0,0 +1,182 @@
...
+        """
+        s3_hook = S3Hook(
+            aws_conn_id=context["params"]["aws_conn_id"], region_name=REGION_NAME
+        )
+        run_uuid = context["params"]["uuid"]
+        output_file = s3_hook.download_file(
+            key=f"{KEY}/{run_uuid}/output.csv", bucket_name=BUCKET_NAME
+        )
+
+        os.rename(output_file, "dcqc_output.csv")
```
BryanFauble, line 160: If it ends up that you'll allow multiple manifest files to be present in the container this logic might need to change depending on how you run the DAG tasks

BryanFauble (approved):
Great work! Just a few comments on areas to doublecheck.

dags/dcqc-poc-dag.py
BWMac: Yeah, and that suffix is parameterized in the DAG so you could change the naming convention for a run if you wanted.

dags/dcqc-poc-dag.py
BWMac: I had the same thought, but since this is a POC I went ahead and limited it. We haven't had a DAG that triggers more than one Tower job before so I decided not to take on figuring out the best way to do so for this PR.

dags/dcqc-poc-dag.py
BWMac: The machine that airflow runs on does use UTC time, but I will specify this so that it doesn't matter where this is run.

dags/dcqc-poc-dag.py
BWMac: The other reason I didn't do this is to avoid confusion because I am already importing the OOP File class

PR #51, opened 2024-05-01
Updates UUID Handling for `dynamic-challenge-dag.py`

BWMac:
**Problem:**

While working on #50, I realized that the way I used the UUIDs in `dynamic-challenge-dag` doesn't work exactly as intended. I intended to have the manifest file path and tower workflow run ID have matching UUIDs in them so that it would be easy to tell which file goes with which run for debugging purposes, but it turns out that the way Airflow executes the DAGs causes different UUIDs to be used in different tasks. The DAG is working otherwise and challenge submissions are being evaluated correctly, but the UUIDs do not match.

**Solution:**

I fixed this by making the UUID a parameter of the DAG run so that the same value can be pulled into any task.

**Testing:**

- Successful tower [run](https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/dynamic-challenge-project/watch/22gQxq5ujjuFLp)
- Screenshot of staged submissions manifest location with the same UUID in the path as in the tower run name:
Screenshot 2024-05-01 at 2 37 35 PM

### DIFF
--- a/dags/dynamic-challenge-dag.py
+++ b/dags/dynamic-challenge-dag.py
@@ -12,28 +12,25 @@ from orca.services.nextflowtower import NextflowTowerHook
 from orca.services.nextflowtower.models import LaunchInfo
 from orca.services.synapse import SynapseHook
 
-UUID = uuid.uuid4()
 REGION_NAME = "us-east-1"
 BUCKET_NAME = "dynamic-challenge-project-tower-scratch"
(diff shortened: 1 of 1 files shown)


jaymedina (approved):

thomasyu888 (approved):
🔥 LGTM!  great catch!

PR #52, opened 2024-05-09
[IBCDPE-918] update revision for PEGS DAG

jaymedina:
### problem
One [more feature](https://github.com/Sage-Bionetworks-Workflows/nf-synapse-challenge/pull/29) needed to be added to `nf-synapse-challenge`, so the revision that the PEGS DAG is pointed to needed to be updated.

### solution
Update the revision to the latest commit to reflect new features

### testing & preview

The DAG has been running successfully

image

With no issues on the tower end (i.e. manifest and other params are correctly being passed on through `LaunchInfo`)

image


BryanFauble (approved):
LGTM

PR #54, opened 2024-05-31
[IBCDPE-907] Sort by unique user count instead of downloads

BryanFauble:
Problem:

1. Downloads are sorted by number of files being downloaded

Solution:

1. Sorting by number of unique users instead

Testing:

1. Verified that the query executes as expected in snowflake:
image


BWMac (approved):
🚀

PR #59, opened 2024-07-09
[IBCDPE-1003] Adds backfill option to TopCharts DAG

BWMac:
**Problem:**
Due to some recent stability issues with our Airflow deployment, the `top_public_synapse_projects_from_snowflake` has failed to run the past two Saturdays. This causes missing data in the "Top Downloaded Public Synapse Projects" Synapse [table](https://www.synapse.org/Synapse:syn53696951/tables/#query/eyJzcWwiOiJTRUxFQ1QgKiBGUk9NIHN5bjUzNjk2OTUxIFdIRVJFIHByb2plY3RfaWQgPSAnc3luMjU4MDg1MyciLCAiaW5jbHVkZUVudGl0eUV0YWciOnRydWUsICJsaW1pdCI6MjUsICJzb3J0IjpbeyJjb2x1bW4iOiJyZWNvcmRfZGF0ZSIsICJkaXJlY3Rpb24iOiJERVNDIn1dfQ==).

**Solution:**
Add an option to run the DAG in "backfill mode". This was achieved by adding two new DAG parameters, `backfill` (boolean, `False` by default) and `backfill_date` (string date, YYYY-MM-DD format). When data is missing, we can go into the Airflow UI and manually trigger a DAG run where `backfill` is toggled on and `backfill_date` is set to the date of the failed DAG run.

**Notes:**
- When the DAG is run in backfill mode, a branching task prevents the part of the DAG that creates and posts a message to Slack from running. This is important to avoid confusing messages with incorrect information from being pushed to the topcharts channel.
- New DAG flow for a run where `backfill` is `True`:
Screenshot 2024-07-09 at 10 25 20 AM

*Side note, should we disable SonarCloud in this repo? It doesn't understand the structure of a DAG and so most of what it flags is not helpful.

### DIFF
--- a/dags/top-public-synapse-projects-from-snowflake.py
+++ b/dags/top-public-synapse-projects-from-snowflake.py
@@ -157,7 +157,7 @@ def top_public_synapse_projects_from_snowflake() -> None:
         return metrics
     
     @task.branch()
-    def check_backfill(**context) -> None:
+    def check_backfill(**context) -> str:
         """Check if the backfill is enabled. When it is, do not post to Slack."""
         if context["params"]["backfill"]:
             return "stop_dag"
@@ -165,6 +165,7 @@ def top_public_synapse_projects_from_snowflake() -> None:
     
     @task()
     def stop_dag() -> None:
+        """Stop the DAG."""
         pass
 
     @task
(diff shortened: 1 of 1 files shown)


BryanFauble:
> *Side note, should we disable SonarCloud in this repo? It doesn't understand the structure of a DAG and so most of what it flags is not helpful.

@BWMac 
What we could do is for the rules that don't make sense for airflow script we can turn those rules off.

I created a new sonarcloud Quality Profile: https://sonarcloud.io/organizations/sage-bionetworks-workflows/quality_profiles/show?name=Sage+way+-+Workflows&language=py

`Sage way - Workflows`

I also have you permissions on the profile, so I think that gives you the right to modify which rules are active. Would you mind creating a tech debt ticket to look back at this. We'll want to turn off or modify the rules that don't make sense in the context of airflow/nextflow scripts that are contained within this repo.

I have already updated orca-recipes in sonarcloud to use this new profile, so any changes to the active rules will be used on the next analysis.

BryanFauble (approved):
Great changes thanks for making this

PR #64, opened 2024-07-30
[SNOW-97] Adds "Synapse by the numbers" DAG

BWMac:
**Description:**

In order to enable the "Synapse by the numbers" section of the new Synapse homepage, we need to have a Synapse table which contains metrics describing the total amount of data hosted in Synapse, the number of active users last month, and the number of downloads last month which should be updated on the 1st of each month. 

The DAG implemented in this PR does exactly that. It has been tested by uploading to a test table [here](https://www.synapse.org/Synapse:syn61915256/tables/). I also tested it in the official table [here](https://www.synapse.org/Synapse:syn61588123/tables/), but subsequently deleted the record created so that the next record will be added on 8/1.

### DIFF
--- a/dags/synapse-by-the-numbers-dag.py
+++ b/dags/synapse-by-the-numbers-dag.py
@@ -1,0 +1,159 @@
+"""This script is used to execute a query on Snowflake and report the results to a 
+Synapse table. This retrieves data describing the total amount of data hosted in Synapse,
+the number of active users last month, and the number of downloads last month."""
+
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import List
+
+import synapseclient
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
+from orca.services.synapse import SynapseHook
+
+dag_params = {
+    "snowflake_conn_id": Param("SNOWFLAKE_SYSADMIN_PORTAL_RAW_CONN", type="string"),
+    "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
+    "current_date": Param(date.today().strftime("%Y-%m-%d"), type="string"),
+}
+
+dag_config = {
+    # run on the 1st of every month at midnight
+    "schedule_interval": "0 0 1 * *",
+    "start_date": datetime(2024, 7, 1),
+    "catchup": False,
+    "default_args": {
+        "retries": 1,
+    },
+    "tags": ["snowflake"],
+    "params": dag_params,
+}
+
+SYNAPSE_RESULTS_TABLE = "syn61588123"
+
+
+@dataclass
+class DownloadMetric:
+    """Dataclass to hold the download metrics from Synapse.
+
+    Attributes:
+        total_data_size_in_pib: The size of the data hosted on Synapse in PiB
+        active_users_last_month: The number of active users last month
+        total_downloads_last_month: The total number of downloads by users last month
+    """
+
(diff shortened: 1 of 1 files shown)


dags/synapse-by-the-numbers-dag.py
```diff
@@ -0,0 +1,159 @@
...
+                        content_size
+                    FROM
+                        synapse_data_warehouse.synapse.node_latest
+                    JOIN
+                        synapse_data_warehouse.synapse.file_latest
+                        ON node_latest.file_handle_id = file_latest.id
+                    UNION
+                    SELECT DISTINCT
+                        file_latest.id,
+                        content_size
```
thomasyu888: Just a quick note: Are there duplicated file handle ids that is generated from this union?

BryanFauble: No:
https://docs.snowflake.com/en/sql-reference/operators-query#union-all

`UNION combines with duplicate elimination.`

dags/synapse-by-the-numbers-dag.py
```diff
@@ -0,0 +1,159 @@
...
+        """Execute the query on Snowflake and return the results."""
+        snow_hook = SnowflakeHook(context["params"]["snowflake_conn_id"])
+        ctx = snow_hook.get_conn()
+        cs = ctx.cursor()
+        query = f"""
+            WITH TOTAL_SIZE AS (
+                SELECT
+                    SUM(content_size) / POWER(2, 50) AS SIZE_IN_PETABYTES
+                FROM (
+                    SELECT DISTINCT
```
thomasyu888: Nit: in the future, split off the subquery into a CTE

thomasyu888 (approved):
🔥 excellent!

dags/synapse-by-the-numbers-dag.py
BWMac: Like [this](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/64/commits/fa58b604780e4bcee27d34737a1f0f20940336f0)?

dags/synapse-by-the-numbers-dag.py
```diff
@@ -0,0 +1,160 @@
...
+import synapseclient
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
+from orca.services.synapse import SynapseHook
+
+dag_params = {
+    "snowflake_conn_id": Param("SNOWFLAKE_SYSADMIN_PORTAL_RAW_CONN", type="string"),
+    "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
+    "current_date": Param(date.today().strftime("%Y-%m-%d"), type="string"),
```
BryanFauble: I might make a small modification here to change this to be the month that this is being run for. I'm thinking to make it easier if we need to manually run this we don't have to think about putting the month after the one we want the data for.

So get today's date and subtract a month in lieu of it being passed in, then you won't need to use the DATEADD procedure in the sql query.

BryanFauble: It's a little unfortunate since I set the standard for the other queries, but thinking back I didn't really like that design so much

BryanFauble (approved):
Changes look great! Just 1 suggestion that your free to take or leave depending on if it'll make things simpler.

dags/synapse-by-the-numbers-dag.py
BWMac: Agreed. Will do!

PR #65, opened 2024-07-31
[SNOW-98] Add Trending Snapshots DAG

jaymedina:
This DAG will automate populating the [Trending Projects Snapshots table](https://www.synapse.org/Synapse:syn61597055/tables/) at the users' desired cadence.

- [x] Successful initial run of DAG on Trending Snapshots Test
- [x] Fix calculation of `TOTAL_DATA_SIZE_IN_GIB`
- [x] Revert changes made for debugging ( date, time window )
- [x] Clean up DAG script ( refactor, document, etc )

### 1 : Test query efficacy
Testing the query on [this test Project](https://www.synapse.org/Synapse:syn57428232/files/) ensures that the project size (i.e. the total number of File entities stored within a project) is being calculated correctly:

image

----

### 2 : Test query efficacy
Confirming the query results for the 30 days leading up to 6/29/2024 approach the results from the initial query

** Current query results **
image

** Initial query results **
image

----

### 3 : Test query efficacy
This query also accounts for cases where there are no new downloads / users within the time range specified (thanks to `LEFT JOIN`)

image

----

### 4 : Test DAG functionality
[Trending Snapshots Test](https://www.synapse.org/Synapse:syn61932294/tables/)

----

### 5 : Test DAG functionality

I changed `time_window` in the `dag_config` to 2 days to ensure the `n_unique_users` reflects this shorter time window. Results:

[Trending Snapshots Test 2 days](https://www.synapse.org/Synapse:syn61942766/tables/)

### DIFF
--- a/dags/trending-projects-snapshot-dag.py
+++ b/dags/trending-projects-snapshot-dag.py
@@ -30,7 +30,7 @@ dag_params = {
     "snowflake_conn_id": Param("SNOWFLAKE_SYSADMIN_PORTAL_RAW_CONN", type="string"),
     "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
     "current_date": Param(date.today().strftime("%Y-%m-%d"), type="string"),
-    "month_to_run": Param((date.today() - relativedelta(months=1)).strftime("%Y-%m-%d"), type="string")
+    "month_to_run": Param((date.today() - relativedelta(months=0)).strftime("%Y-%m-%d"), type="string")
     }
 
 dag_config = {
@@ -55,13 +55,13 @@ class SnapshotMetrics:
         project_id: The ID of the Project
         n_unique_users: The number of unique users who downloaded from the Project
         last_download_date: The date of the last download from the Project
-        total_data_size_in_gib: The current total size of the Project in GiB
+        estimated_project_size_in_gib: The current total size of the Project in GiB
     """
 
     project_id: float
     n_unique_users: int
     last_download_date: int
-    total_data_size_in_gib: float
+    estimated_project_size_in_gib: float
 
 
 @dag(**dag_config)
@@ -112,11 +112,12 @@ def trending_projects_snapshot() -> None:
                     SELECT ID, PROJECT_ID, FILE_HANDLE_ID
                     FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
                     WHERE 1=1
+                    AND DATE_TRUNC('MONTH', CHANGE_TIMESTAMP) <= DATE_TRUNC('MONTH', DATE('{context["params"]["month_to_run"]}'))
                     AND NODE_TYPE = 'file'
                     AND PROJECT_ID IN (SELECT PROJECT_ID FROM TOP_10_PUBLIC_PROJECTS)
                 ),
                 FILE_SIZES AS (
-                    SELECT LATEST_FILE_HANDLES.PROJECT_ID, SUM(FILE_LATEST.CONTENT_SIZE) / POWER(1024, 3) AS TOTAL_DATA_SIZE_IN_GIB
+                    SELECT LATEST_FILE_HANDLES.PROJECT_ID, SUM(FILE_LATEST.CONTENT_SIZE) / POWER(1024, 3) AS ESTIMATED_PROJECT_SIZE_IN_GIB
                     FROM LATEST_FILE_HANDLES
                     JOIN SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILE_LATEST
                     ON LATEST_FILE_HANDLES.FILE_HANDLE_ID = FILE_LATEST.ID
@@ -127,7 +128,7 @@ def trending_projects_snapshot() -> None:
                     TOP_10_PUBLIC_PROJECTS.PROJECT_ID,
                     TOP_10_PUBLIC_PROJECTS.N_UNIQUE_USERS,
                     COALESCE(TO_CHAR(MAX(RECENT_DOWNLOADS.RECORD_DATE), 'YYYY-MM-DD'), 'N/A') AS LAST_DOWNLOAD_DATE,
-                    ROUND(FILE_SIZES.TOTAL_DATA_SIZE_IN_GIB, 3) AS TOTAL_DATA_SIZE_IN_GIB
+                    ROUND(FILE_SIZES.ESTIMATED_PROJECT_SIZE_IN_GIB, 3) AS ESTIMATED_PROJECT_SIZE_IN_GIB
                 FROM 
                     TOP_10_PUBLIC_PROJECTS
 
@@ -140,7 +141,7 @@ def trending_projects_snapshot() -> None:
                 GROUP BY
                     TOP_10_PUBLIC_PROJECTS.PROJECT_ID,
                     TOP_10_PUBLIC_PROJECTS.N_UNIQUE_USERS,
-                    FILE_SIZES.TOTAL_DATA_SIZE_IN_GIB
+                    FILE_SIZES.ESTIMATED_PROJECT_SIZE_IN_GIB
 
                 ORDER BY 
                     TOP_10_PUBLIC_PROJECTS.N_UNIQUE_USERS DESC
@@ -156,7 +157,7 @@ def trending_projects_snapshot() -> None:
                     project_id=row["PROJECT_ID"],
                     n_unique_users=row["N_UNIQUE_USERS"],
                     last_download_date=row["LAST_DOWNLOAD_DATE"],
-                    total_data_size_in_gib=row["TOTAL_DATA_SIZE_IN_GIB"],
+                    estimated_project_size_in_gib=row["ESTIMATED_PROJECT_SIZE_IN_GIB"],
                 )
             )
         return metrics
@@ -172,7 +173,7 @@ def trending_projects_snapshot() -> None:
                     metric.project_id,
                     metric.n_unique_users,
                     metric.last_download_date,
-                    metric.total_data_size_in_gib,
+                    metric.estimated_project_size_in_gib,
                     export_date
                 ]
             )
--- a/dags/trending-projects-snapshot-dag.py
+++ b/dags/trending-projects-snapshot-dag.py
@@ -93,48 +93,56 @@ def trending_projects_snapshot() -> None:
                     AND NODE_TYPE = 'project'
                     AND IS_PUBLIC = TRUE
                 ),
-                LATEST_FILE_HANDLES AS (
-                    SELECT DISTINCT ID, PROJECT_ID, FILE_HANDLE_ID
-                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
-                    WHERE 1=1
-                    AND NODE_TYPE = 'file'
-                    AND PROJECT_ID IN (SELECT PROJECT_ID FROM PUBLIC_PROJECTS)
-                ),
-                FILE_SIZES AS (
-                    SELECT fh.PROJECT_ID, SUM(f.CONTENT_SIZE) / POWER(1024, 3) AS TOTAL_DATA_SIZE_IN_GIB
-                    FROM LATEST_FILE_HANDLES fh
-                    JOIN SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILE_LATEST f
-                    ON fh.FILE_HANDLE_ID = f.ID
-                    GROUP BY fh.PROJECT_ID
-                ),
                 RECENT_DOWNLOADS AS (
                     SELECT PROJECT_ID, RECORD_DATE, USER_ID
                     FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILEDOWNLOAD
-                    WHERE 1=1
-                    AND DATE_TRUNC('MONTH', RECORD_DATE) = DATE_TRUNC('MONTH', DATE({context["params"]["month_to_run"]}))
+                    WHERE DATE_TRUNC('MONTH', RECORD_DATE) = DATE_TRUNC('MONTH', DATE({context["params"]["month_to_run"]}))
                     AND STACK = 'prod'
                     AND PROJECT_ID IN (SELECT PROJECT_ID FROM PUBLIC_PROJECTS)
+                ),
+                TOP_10_PUBLIC_PROJECTS AS (
+                    SELECT PROJECT_ID, COUNT(DISTINCT USER_ID) AS N_UNIQUE_USERS
+                    FROM RECENT_DOWNLOADS
+                    GROUP BY PROJECT_ID
+                    ORDER BY N_UNIQUE_USERS DESC
+                    LIMIT 10
+                ),
+                LATEST_FILE_HANDLES AS (
+                    SELECT ID, PROJECT_ID, FILE_HANDLE_ID
+                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
+                    WHERE 1=1
+                    AND NODE_TYPE = 'file'
+                    AND PROJECT_ID IN (SELECT PROJECT_ID FROM TOP_10_PUBLIC_PROJECTS)
+                ),
+                FILE_SIZES AS (
+                    SELECT LATEST_FILE_HANDLES.PROJECT_ID, SUM(FILE_LATEST.CONTENT_SIZE) / POWER(1024, 3) AS TOTAL_DATA_SIZE_IN_GIB
+                    FROM LATEST_FILE_HANDLES
+                    JOIN SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILE_LATEST
+                    ON LATEST_FILE_HANDLES.FILE_HANDLE_ID = FILE_LATEST.ID
+                    GROUP BY LATEST_FILE_HANDLES.PROJECT_ID
                 )
 
                 SELECT 
-                    PUBLIC_PROJECTS.PROJECT_ID,
-                    COUNT(DISTINCT RECENT_DOWNLOADS.USER_ID) AS N_UNIQUE_USERS,
+                    TOP_10_PUBLIC_PROJECTS.PROJECT_ID,
+                    TOP_10_PUBLIC_PROJECTS.N_UNIQUE_USERS,
                     COALESCE(TO_CHAR(MAX(RECENT_DOWNLOADS.RECORD_DATE), 'YYYY-MM-DD'), 'N/A') AS LAST_DOWNLOAD_DATE,
                     ROUND(FILE_SIZES.TOTAL_DATA_SIZE_IN_GIB, 3) AS TOTAL_DATA_SIZE_IN_GIB
                 FROM 
-                    PUBLIC_PROJECTS
+                    TOP_10_PUBLIC_PROJECTS
 
                 LEFT JOIN RECENT_DOWNLOADS
-                ON PUBLIC_PROJECTS.PROJECT_ID = RECENT_DOWNLOADS.PROJECT_ID
+                ON TOP_10_PUBLIC_PROJECTS.PROJECT_ID = RECENT_DOWNLOADS.PROJECT_ID
 
-                LEFT JOIN FILE_SIZES
-                ON PUBLIC_PROJECTS.PROJECT_ID = FILE_SIZES.PROJECT_ID
+                JOIN FILE_SIZES
+                ON TOP_10_PUBLIC_PROJECTS.PROJECT_ID = FILE_SIZES.PROJECT_ID
 
                 GROUP BY
-                    PUBLIC_PROJECTS.PROJECT_ID, FILE_SIZES.TOTAL_DATA_SIZE_IN_GIB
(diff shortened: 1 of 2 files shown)


dags/trending-snapshots-dag.py
```diff
@@ -0,0 +1,166 @@
...
+                    SUM(file_latest.CONTENT_SIZE) / POWER(1024, 3) AS TOTAL_DATA_SIZE_IN_GIB
+
+                -- Select the table you want to curate from
+                FROM 
+                    RECENT_DOWNLOADS recent_downloads
+
+                -- Join FILE_LATEST
+                JOIN 
+                    SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILE_LATEST file_latest
+                ON
```
thomasyu888: What is the purpose of this join?

jaymedina: `FILE_LATEST` is used to get the sum of the `CONTENT_SIZE` for each File entity within a given project, to get the aggregate size of the project (L98 currently)

dags/trending-snapshots-dag.py
```diff
@@ -0,0 +1,166 @@
...
+        """Execute the query on Snowflake and return the results."""
+        snow_hook = SnowflakeHook(context["params"]["snowflake_conn_id"])
+        ctx = snow_hook.get_conn()
+        cs = ctx.cursor()
+        query = f"""
+                WITH RECENT_DOWNLOADS AS (
+                    SELECT *
+                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILEDOWNLOAD
+                    WHERE 1=1
+                    AND RECORD_DATE > DATEADD(DAY, -28, '2024-06-26')
```
jaymedina: I'm changing this later. I'm trying to re-create the numbers in the proof-of-concept schema in the Google Sheets. Right now the numbers are _slightly_ different, if you run this on Snowsight, and my goal is to figure that out tomorrow.

dags/trending-snapshots-dag.py
thomasyu888: That makes sense, but I think you are joining against the filedownload table which would give you total data transferred in bytes

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,183 @@
...
+    def get_trending_project_snapshot(**context) -> List[SnapshotMetrics]:
+        """Execute the query on Snowflake and return the results."""
+        snow_hook = SnowflakeHook(context["params"]["snowflake_conn_id"])
+        ctx = snow_hook.get_conn()
+        cs = ctx.cursor()
+        query = f"""
+                WITH PUBLIC_PROJECTS AS (
+                    SELECT PROJECT_ID
+                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
+                    WHERE 1=1
```
jaymedina, line 92: This is for code styling to make things easier to read. We can assume anything with the `AND` operator is an actual filter.

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,183 @@
...
+                    AND RECORD_DATE > DATEADD(DAY, -{context["params"]["time_window"]}, CURRENT_DATE()) 
+                    AND STACK = 'prod'
+                    AND PROJECT_ID IN (SELECT PROJECT_ID FROM PUBLIC_PROJECTS)
+                )
+
+                SELECT 
+                    pp.PROJECT_ID,
+                    COUNT(DISTINCT rd.USER_ID) AS N_UNIQUE_USERS,
+                    COALESCE(TO_CHAR(MAX(rd.RECORD_DATE), 'YYYY-MM-DD'), 'N/A') AS LAST_DOWNLOAD_DATE,
```
jaymedina: In the event that we have no new downloads within the requested time window for a given project, we `LEFT JOIN` onto `PUBLIC_PROJECTS` and fill `NULL` values with `N/A`.

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,183 @@
...
+                    FROM LATEST_FILE_HANDLES fh
+                    JOIN SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILE_LATEST f
+                    ON fh.FILE_HANDLE_ID = f.ID
+                    GROUP BY fh.PROJECT_ID
+                ),
+                RECENT_DOWNLOADS AS (
+                    SELECT PROJECT_ID, RECORD_DATE, USER_ID
+                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.FILEDOWNLOAD
+                    WHERE 1=1
+                    AND RECORD_DATE > DATEADD(DAY, -{context["params"]["time_window"]}, CURRENT_DATE())
```
thomasyu888: See Brad's PR on getting last month and not last 30 days.

dags/trending-projects-snapshot-dag.py
thomasyu888: Nice!

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,183 @@
...
+                    pp.PROJECT_ID,
+                    COUNT(DISTINCT rd.USER_ID) AS N_UNIQUE_USERS,
+                    COALESCE(TO_CHAR(MAX(rd.RECORD_DATE), 'YYYY-MM-DD'), 'N/A') AS LAST_DOWNLOAD_DATE,
+                    ROUND(fs.TOTAL_DATA_SIZE_IN_GIB, 3) AS TOTAL_DATA_SIZE_IN_GIB
+                FROM 
+                    PUBLIC_PROJECTS pp
+
+                -- Join RECENT_DOWNLOADS
+                LEFT JOIN RECENT_DOWNLOADS rd
+                ON pp.PROJECT_ID = rd.PROJECT_ID
```
thomasyu888: Nit: this is just personal preference, but short hand is never easier to read than just writing the whole table out:


```
FROM 
    PUBLIC_PROJECTS
LEFT JOIN
    RECENT_DOWNLOADS
    ON PUBLIC_PROJECTS.PROJECT_ID = RECENT_DOWNLOADS.PROJECT_ID
```

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,183 @@
...
+
+                SELECT 
+                    pp.PROJECT_ID,
+                    COUNT(DISTINCT rd.USER_ID) AS N_UNIQUE_USERS,
+                    COALESCE(TO_CHAR(MAX(rd.RECORD_DATE), 'YYYY-MM-DD'), 'N/A') AS LAST_DOWNLOAD_DATE,
+                    ROUND(fs.TOTAL_DATA_SIZE_IN_GIB, 3) AS TOTAL_DATA_SIZE_IN_GIB
+                FROM 
+                    PUBLIC_PROJECTS pp
+
+                -- Join RECENT_DOWNLOADS
```
thomasyu888: Nit: The comment can seem a bit redundant here.

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,183 @@
...
+                    WHERE 1=1
+                    AND NODE_TYPE = 'project'
+                    AND IS_PUBLIC = TRUE
+                ),
+                LATEST_FILE_HANDLES AS (
+                    SELECT DISTINCT ID, PROJECT_ID, FILE_HANDLE_ID
+                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
+                    WHERE 1=1
+                    AND NODE_TYPE = 'file'
+                    AND PROJECT_ID IN (SELECT PROJECT_ID FROM PUBLIC_PROJECTS)
```
thomasyu888: NIt: I was going to say you could combine the PUBLIC_PROJECTS / LATEST_FILE_HANDLES / FILE_SIZES into one CTE, but I understand why you have PUBLIC_PROJECTS as a separate CTE.

dags/trending-projects-snapshot-dag.py
thomasyu888: What is your rationale here for shifting this to a character?  the expected output is what we have for "export_date" so keep that in mind.  Nice use of Coalese though for those projects that don't have download dates at all. 

One query optimization is to first get the top 10 downloaded public projects by user, then only grab the data sizes and activity for those projects.

jaymedina: > What is your rationale here for shifting this to a character? 

It's because `COALESCE` needs all the arguments to be the same data type (in this case, string). I was getting a `Date 'N/A' is not recognized` error otherwise.

>  the expected output is what we have for "export_date" so keep that in mind

Noted - it looks like the format on Synapse is determined by how the schema is formatted on Synapse itself. I did a test run using the date formatting as it is (`'YYYY-MM-DD'`) but it shows up like `MM/DD/YYYY` on Synapse, same formatting as `export_date`: 
image

jaymedina: > One query optimization is to first get the top 10 downloaded public projects by user, then only grab the data sizes and activity for those projects.

[Updated ✔️ ](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/65/commits/cc55b4e92d33609233282495f50416abe37d11c7)

`RECENT_DOWNLOADS` still works with `PUBLIC_PROJECTS` (not `TOP_10`) because the downloads are what's used to determine what the top 10 projects for that month are. After that, the remaining CTEs use the results from the `TOP_10` project CTE to get the file handles and the corresponding total size for each project, which minimizes calculations. 

Here are my results for the month of June:

image

I tried comparing the Query duration between the old and new version of this query, but tbh the calculation time is so small it's hard to tell. I think logically, though, this change does minimize compute.

One other thing that came to mind as I was working on this is we should consider expanding the scope of the trending project snapshots to also include upload activity. In other words `N_UNIQUE_USERS` would include users both uploading and downloading, and we'd have another column for `LAST_UPLOAD_DATE`.

dags/trending-projects-snapshot-dag.py
jaymedina: While I was working on the optimization suggestion from [this thread](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/65#discussion_r1701338028), I considered having the `RECENT_DOWNLOADS` CTE join `FILEDOWNLOAD` and `NODE_LATEST` and getting rid of `PUBLIC_PROJECTS` altogether, but I decided against it because:

- I don't believe it would minimize compute ( I tried it out a few times and found no significant difference in Query duration between the version with `PUBLIC_PROJECTS` )
- Smaller CTEs are easier to read and makes the query easier to understand (from my perspective)

But lmk what you think!

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,190 @@
...
+        snow_hook = SnowflakeHook(context["params"]["snowflake_conn_id"])
+        ctx = snow_hook.get_conn()
+        cs = ctx.cursor()
+        query = f"""
+                WITH PUBLIC_PROJECTS AS (
+                    SELECT PROJECT_ID
+                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
+                    WHERE 1=1
+                    AND NODE_TYPE = 'project'
+                    AND IS_PUBLIC = TRUE
```
thomasyu888:
```suggestion
                    AND IS_PUBLIC
```

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,190 @@
...
+                    GROUP BY PROJECT_ID
+                    ORDER BY N_UNIQUE_USERS DESC
+                    LIMIT 10
+                ),
+                LATEST_FILE_HANDLES AS (
+                    SELECT ID, PROJECT_ID, FILE_HANDLE_ID
+                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
+                    WHERE 1=1
+                    AND NODE_TYPE = 'file'
+                    AND PROJECT_ID IN (SELECT PROJECT_ID FROM TOP_10_PUBLIC_PROJECTS)
```
thomasyu888, line 117: We should probably add in a `is_public` check here as well because there can be private files within a public project that we want to filter activity out for.

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,190 @@
...
+                FROM 
+                    TOP_10_PUBLIC_PROJECTS
+
+                LEFT JOIN RECENT_DOWNLOADS
+                ON TOP_10_PUBLIC_PROJECTS.PROJECT_ID = RECENT_DOWNLOADS.PROJECT_ID
+
+                JOIN FILE_SIZES
+                ON TOP_10_PUBLIC_PROJECTS.PROJECT_ID = FILE_SIZES.PROJECT_ID
+
+                GROUP BY
```
thomasyu888, line 141: Just some food for thought as there are many ways to solve a problem.  Another approach is to do this in two parts

1. Calculate the top 10 downloaded projects and their last download activity (`DL_INFO_CTE`)
2. Calculate the data size of those top 10 downloaded project (`DATA_SIZE_CTE`)
3. join `DL_INFO_CTE` with `DATA_SIZE_CTE` on `project_id` column.

That's effectively what you're doing here, but this `group by` clause can be a bit confusing at first glance

thomasyu888 (approved):
🔥 LGTM!  I'm going to pre-approve, but please see some of my comments.  Excellent work here!

dags/trending-projects-snapshot-dag.py
jaymedina: I decided against this because calculating a project size based off only the public files wouldn't truly reflect the size of the project itself. It would also mislead users into thinking that they have access to all of a project, when in reality they might only have access to a portion of it, if that makes sense. So right now I'm still leaning against filtering for public files. What do you think?

jaymedina:
Thanks for the review @thomasyu888! I'll wait until after Sage Week before merging.

Some final to-do items for me:

* Re-run DAG against `Trending Snapshots Test` table to make sure results look as expected
* Revert back to point at the official table
* Merge

Quick thought:
I have the DAG scheduled to run on the 2nd of each month, so with our current merge timeline, the next run would be September 2nd. If we need an updated version of the table before then, I can run it manually. Just want to run this by you so we're on the same page.

Thanks!

thomasyu888:
> Quick thought: I have the DAG scheduled to run on the 2nd of each month, so with our current merge timeline, the next run would be September 2nd. If we need an updated version of the table before then, I can run it manually. Just want to run this by you so we're on the same page.

Great work here! Yes, please run it manually this time, and September will be automated.

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,190 @@
...
+                    total_data_size_in_gib=row["TOTAL_DATA_SIZE_IN_GIB"],
+                )
+            )
+        return metrics
+
+    @task
+    def push_results_to_synapse_table(metrics: List[SnapshotMetrics], **context) -> None:
+        """Push the results to a Synapse table."""
+        data = []
+        export_date = date.today()
```
BryanFauble, line 169: This `export_date` may lead to some issues if we are backfilling data. I am thinking of the situation where "I want to load the data from Jan 2024". In that case it is correct that I would be exporting the data today, however, the data being exported represents data for the month of Jan 2024.

Maybe this would need to change to `month_to_run` instead of `date.today()`?

BryanFauble: Maybe the table might need a column for when the data was exported. And a column for which month the data represents?

BryanFauble (approved):
1 comment around export date. But everything else looks good. I will go ahead and pre-approve, please re-request review if there are sizable changes.

dags/trending-projects-snapshot-dag.py
jaymedina: Good point. `LAST_DOWNLOAD_DATE` shares the date of the last download within the designated month for the query, so technically this is the column that shares which month the data pertains to: If you take a look at my [Trending Snapshots Test](https://www.synapse.org/Synapse:syn61932294/tables/) table you'll see that I exported data from both June and July on the same day, as denoted by `LAST_DOWNLOAD_DATE` and `EXPORT_DATE`.

One solution we can go with is renaming `LAST_DOWNLOAD_DATE` to something more specific like `LAST_DOWNLOAD_OF_MONTH`. What do you think of moving forward this way?

BryanFauble: `LAST_DOWNLOAD_DATE` should be sufficient for what you need here. So then I think you'd be able to query for a month worth of data like:


`SELECT * FROM syn61932294 where LAST_DOWNLOAD_DATE >= '1717200000000' AND  LAST_DOWNLOAD_DATE < '1719792000000'`

1717200000000 = `Sat Jun 01 2024 00:00:00.000` UTC
1719792000000 = `Mon Jul 01 2024 00:00:00.000` UTC

jaymedina: The way we are querying for data is with the `DATE_TRUNC` operator on Snowsight's SQL. So it retrieves downloads for a given month based on the month_to_download given in the dag params:

image

BryanFauble: Right, that's on the snowflake side. 

The query I gave would be what you might use to select data from the Synapse table

jaymedina: I see what you mean. Yes, that's right! 
image

jaymedina:
Before I merge this in, I'm having second thoughts about what's in `TOTAL_DATA_SIZE` column.

The screenshot below shows a backfill of the top 10 trending Projects in February 2024. While `N_UNIQUE_USERS` and `LAST_DOWNLOAD_DATE` are a snapshot of those metrics of the Project within that month, `TOTAL_DATA_SIZE` still reflects the most _recent_ size of the project, rather than what the size was on February.

I think the important metric is the number of unique users, as that is what determines what the "top 10" trending projects are, but I wonder if that project size column is misleading.

If we want `TOTAL_DATA_SIZE` to reflect what the project size was _for that month_ and not what it is currently, the change would be simple. I would just add another WHERE clause to the `LATEST_FILE_HANDLES` CTE, like so (with `DATE()` being feed programmatically rather than hardcoded like it is below:

```
AND DATE_TRUNC('MONTH', CHANGE_TIMESTAMP) <= DATE_TRUNC('MONTH', DATE('2024-2-12'))
```

But I'd like to know your thoughts on this before I move forward @BryanFauble @thomasyu888 

image

The other solution (simpler) is just change the column name to `CURRENT_TOTAL_DATA_SIZE` but I wonder if users would find it useful to know how big the project is at the time that the export happened.

thomasyu888:
> 
> But I'd like to know your thoughts on this before I move forward @BryanFauble @thomasyu888 
> 

@jaymedina this is a great catch! I think what we probably care about is when the entity was created on, so if we filter out all entities created before the runtime, it would give us a more accurate representation.  The issue with this is that we don't capture deletions so the data won't be fully accurate.

jaymedina:
> The issue with this is that we don't capture deletions so the data won't be fully accurate.

I see. Then I'll move forward with modifying the total project size calculations and can fold the caveat you mention into the column name itself by modifying the name to be `ESTIMATED_PROJECT_SIZE_IN_GIB`, what do you think?

thomasyu888:
> modifying the name to be ESTIMATED_PROJECT_SIZE_IN_GIB, what do you think?

Great idea - before doing that change, let's let the frontend team know we are going to make a schema change. Can you post on #synapse? This will impact the homepage.

jaymedina:
After Jay's feedback I've moved forward with the schema modifications and calculating the Project size based off the files available in `FILE_LATEST` that land within the given time window.

[Here](https://www.synapse.org/Synapse:syn61932294/tables/) is a test I just ran with the newest DAG that updates the table for July 2024 and December 2022. Take a look at the new Project size column and notice the difference between Project sizes for the same Projects (notably: `AD Knowledge Portal`).

I'm re-requesting a review since this is a structural change.

Thank you both for your patience and for your feedback!

BryanFauble (approved):
Great change!

dags/trending-projects-snapshot-dag.py
```diff
@@ -0,0 +1,192 @@
...
+                    FROM RECENT_DOWNLOADS
+                    GROUP BY PROJECT_ID
+                    ORDER BY N_UNIQUE_USERS DESC
+                    LIMIT 10
+                ),
+                LATEST_FILE_HANDLES AS (
+                    SELECT ID, PROJECT_ID, FILE_HANDLE_ID
+                    FROM SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
+                    WHERE 1=1
+                    AND DATE_TRUNC('MONTH', CHANGE_TIMESTAMP) <= DATE_TRUNC('MONTH', DATE('{context["params"]["month_to_run"]}'))
```
thomasyu888: This will be an issue because sometimes the CHANGE_TIMESTAMP will be greater than the CREATEDON date. Let's say an entity was created in Feb but updated in April, by doing this query, when you do the calculation of data sizes for Feb, you're going to leave this entity out.

jaymedina: Got it, I've updated it so that as long as the file was created <= the `month_to_run`, it will be used in the calculation of the total Project size.

thomasyu888 (approved):
🎉 LGTM!

PR #66, opened 2024-08-01
Update synapse-by-the-numbers-dag.py to run on the 2nd of every month

BWMac:
**Problem:**

Our DAG running on the 1st of every month at midnight UTC may cause the queried results to be inaccurate and it is confusing because the data appears under a record for the last day of the prior month (when viewed in a local timezone).

**Solution:**

Change this so that the DAG runs on the 2nd of every month at midnight instead.

### DIFF
--- a/dags/synapse-by-the-numbers-dag.py
+++ b/dags/synapse-by-the-numbers-dag.py
@@ -21,7 +21,7 @@ dag_params = {
 }
 
 dag_config = {
-    # run on the 1st of every month at midnight
+    # run on the 2nd of every month at midnight
     "schedule_interval": "0 0 2 * *",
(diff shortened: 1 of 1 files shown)


dags/synapse-by-the-numbers-dag.py
```diff
@@ -22,7 +22,7 @@
 
 dag_config = {
     # run on the 1st of every month at midnight
```
jaymedina: nit: Update comment to reflect this runs the 2nd

jaymedina (approved):

dags/synapse-by-the-numbers-dag.py
BWMac: Whoops, thanks for catching that!

PR #70, opened 2024-10-23
[SNOW-139] Filters Homepage Metrics In Synapse Download Metrics DAGS

BWMac:
**Description:**

This PR adds updates the queries in 4 Airflow DAGs which pull project-specific metrics from Snowflake and pushes those metrics into Synapse to exclude the "[Synapse Homepage Data](https://www.synapse.org/#!Synapse:syn23593546)" project. This will prevent metrics from the homepage from appearing in the "Trending Projects" section displayed on the Homepage itself.

### DIFF
--- a/dags/top-public-synapse-projects-30-days-from-snowflake.py
+++ b/dags/top-public-synapse-projects-30-days-from-snowflake.py
@@ -1,5 +1,5 @@
 """This script is used to execute a query on Snowflake and report the results to a 
-Synapse table. This retrieves download metrics for Public Synapse Projects for the past 30 days.
+Synapse table. This retrieves download metrics for Public Synapse Projects excluding the Synapse Homepage Project for the past 30 days.
 It is scheduled to run at 00:00 every day."""
 
 from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


dags/top-public-synapse-projects-30-days-from-snowflake.py
```diff
@@ -31,6 +31,7 @@
 }
 
 SYNAPSE_RESULTS_TABLE = "syn55382267"
+SYNAPSE_HOMEPAGE_PROJECT_ID = 23593546
```
BryanFauble, line 34: Would it make sense to make this an array, in the event we need to add other projects it would be a simple add?

BryanFauble (approved):
This is great! Thanks for making the change. This will be a great test when we merge it in that the python code is updated on airflow in prod.

dags/top-public-synapse-projects-30-days-from-snowflake.py
BWMac: I think that would be easy enough to do in the future if we needed to, but for now I like the specificity of the named constant so it's clear what is being filtered out.

BryanFauble: Works for me!

PR #71, opened 2024-10-28
[IBCDPE-1111] Update dockerfile install for latest airflow instance

BryanFauble:
**Problem:**

1. We are 1 year behind on airflow updates.
2. The current version of the airflow release has a few bugs in it's deployment via helm to kubernetes that cause the scheduler to fail and not recover.

**Solution:**

1. Update to 2.9.3 of airflow.
2. Install py-orca from a git hash since it previously had a dependency of 2.7.2 of airflow and was causing version downgrades.
3. Pin Pendulum dependency to <3 as it was causing startup issues.

**Testing:**

1. I deployed this out to the sandbox kubernetes cluster and verified that DAGs could run as expected.

**TODO:**

1. When/if we release https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/45 we should remove the need to install the dependency via git hash and instead pull the version from PYPI.

Related PRs:

1. https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/42
2. https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/45

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,22 @@
-ARG BASE_IMAGE=apache/airflow:2.7.2-python3.10
+ARG BASE_IMAGE=apache/airflow:2.9.3-python3.10
 FROM $BASE_IMAGE
 
 RUN pip install --upgrade pip
 
-RUN pip install apache-airflow[amazon,celery,snowflake]==2.7.2 --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-2.7.2/constraints-3.10.txt"
+RUN pip install apache-airflow[amazon,celery,snowflake]==2.9.3 --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-2.9.3/constraints-3.10.txt"
 
 COPY requirements-airflow.txt /tmp/requirements-airflow.txt
 RUN pip install --no-cache-dir -r /tmp/requirements-airflow.txt
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,12 +1,22 @@
...
+USER root
+RUN apt-get update
+RUN apt-get -y install git
+
+# change back to base image user
+USER 50000
+
+# Installed from git a version of py-orca that is compatible with the version of the airflow image
+RUN pip install -e git+https://github.com/Sage-Bionetworks-Workflows/py-orca@7c4781c4750ca8025deb1c6665730d7d46844fbc#egg=py-orca[all]
+
```
BryanFauble: We should probably update py-orca so we don't need to do this. @BWMac Are you aware of the implications of upgrading py-orca with this and what/if any additional testing we'll need to do?

Related PR: https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/45

thomasyu888 (approved):
🔥 LGTM!  Will wait for Brad's return

Dockerfile
BWMac: With the testing that you did in the py-orca PR I don't expect any breaking changes here (basically we just need to make sure DAGs can run).

BWMac (approved):

PR #72, opened 2025-02-06
Updates topcharts DAG docstring

BWMac:
There was some recent confusion about the backfill functionality of this DAG and the docstring doesn't currently provide adequate detail.


philerooski:
This explains well how to configure `backfill_date`, although ideally all the top-level parameter's behavior would be expounded upon. The 'Attributes' format you used in the `DownloadMetric` class docstring does this nicely.

philerooski (approved):
See previous comment.

PR #73, opened 2025-02-11
Updates docker-compose.yaml for airflow 2.9.3

BWMac:
When I was testing the newly rotated AWS credentials used in codespaces on this repo for Airflow development, I ran into an error with our `docker-compose` setup. It is explained in an issue on the airflow repo [here](https://github.com/apache/airflow/discussions/31521). TLDR is that we were using an outdated `docker-compose.yaml` file so I updated it to be on par with the version published with v2.9.3 of `apache-airflow`.

This only impacts airflow development, because our production airflow deployment is using kubernetes and the [helm chart](https://airflow.apache.org/docs/helm-chart/stable/index.html) instead.

I tested that this new docker compose setup works by running Airflow in a codespace and successfully executing the `test-secrets` dag which ensures that the secrets backend config is working.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -24,9 +24,11 @@
 # The following variables are supported:
 #
 # AIRFLOW_IMAGE_NAME           - Docker image name used to run Airflow.
-#                                Default: apache/airflow:2.4.2
+#                                Default: apache/airflow:2.9.3
 # AIRFLOW_UID                  - User ID in Airflow containers
 #                                Default: 50000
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 @BWMac is the production deployed version of Airflow the same as this one?

BWMac:
@thomasyu888 Yes https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/f1f901d195d19036c0a5fb8334bd29e54439563d/modules/apache-airflow/templates/values.yaml#L77

PR #75, opened 2025-02-25
[Snow-228] Created a DAG to get projects, datasets that were created in the past seven days 

linglp:
## Problem
As requested in SNOW-228, a Synapse table needs to be created to store information about projects and datasets created in the past seven days. Additionally, this includes folders that have the annotation "content type = dataset".

## Solution
A DAG was implemented along with a Synapse table to store the retrieved data.
Slack message can be sent to a slack channel

## Test
A table can be successfully updated using airflow: https://www.synapse.org/Synapse:syn64919238/tables/
Made sure that slack messages can be sent to "hotdrop_test" channel

### DIFF
--- a/dags/synapse-datasets-created-7-days.py
+++ b/dags/synapse-datasets-created-7-days.py
@@ -1,5 +1,5 @@
-"""This script executes a query on Snowflake to retrieve data about entities created in the past few days and stores the results in a Synapse table. 
-It is scheduled to run daily at 00:00."""
+"""This script executes a query on Snowflake to retrieve data about entities created in the past 7 days and stores the results in a Synapse table. 
+It is scheduled to run every Monday at 00:00 UTC"""
 
 from dataclasses import dataclass
 from datetime import date, datetime
@@ -7,9 +7,11 @@ from typing import List
 
 import synapseclient
 from airflow.decorators import dag, task
+from airflow.models import Variable
 from airflow.models.param import Param
 from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
 from orca.services.synapse import SynapseHook
+from slack_sdk import WebClient
 
 dag_params = {
     "snowflake_conn_id": Param("SNOWFLAKE_SYSADMIN_PORTAL_RAW_CONN", type="string"),
@@ -18,7 +20,7 @@ dag_params = {
 }
 
 dag_config = {
-    "schedule_interval": "0 0 * * *",
+    "schedule_interval": "0 0 * * 1",
     "start_date": datetime(2025, 2, 1),
     "catchup": False,
     "default_args": {
@@ -109,6 +111,29 @@ def datasets_or_projects_created_7_days() -> None:
                 )
             )
         return entity_created
+    
+    @task
+    def generate_slack_message(entity_created: List[EntityCreated], **context) -> str:
+        """Generate the message to be posted to the slack channel."""
+        message = f":synapse: Datasets or projects created in the last 7 days \n\n"
+        for index, row in enumerate(entity_created):
+            if row.content_type:
+                type = row.content_type.strip("[] \n").strip('"')
+            else:
+                type = row.node_type
+            message += f"{index+1}. <https://www.synapse.org/#!Synapse:syn{row.id}|*{row.name}*> (Type: {type}, Created on: {row.created_on}, Created by: <https://www.synapse.org/Profile:{row.created_by}/profile|this user>, Public: {row.is_public})\n\n"
+            if row.content_type:
+                logger.info("before cleanning", row.content_type)
+                logger.info("after stripping", type)
+        return message
+
+    @task
+    def post_slack_messages(message:str) -> bool:
+        """Post the top downloads to the slack channel."""
+        client = WebClient(token=Variable.get("SLACK_DPE_TEAM_BOT_TOKEN"))
+        result = client.chat_postMessage(channel="hotdrop_test", text=message)
+        print(f"Result of posting to slack: [{result}]")
+        return result is not None
 
     @task
     def push_results_to_synapse_table(entity_created: List[EntityCreated], **context) -> None:
@@ -135,10 +160,14 @@ def datasets_or_projects_created_7_days() -> None:
             synapseclient.Table(schema=SYNAPSE_RESULTS_TABLE, values=data)
         )
 
-    get_entity_created = get_datasets_projects_created_7_days()
-    push_to_synapse_table = push_results_to_synapse_table(entity_created=get_entity_created)
+    entity_created = get_datasets_projects_created_7_days()
+    # push_to_synapse_table = push_results_to_synapse_table(entity_created=entity_created)
+    slack_message = generate_slack_message(entity_created=entity_created)
+    post_to_slack = post_slack_messages(message=slack_message)
 
-    get_entity_created >> push_to_synapse_table
+    entity_created >> slack_message
+    slack_message >> post_to_slack
+    # entity_created >> push_to_synapse_table
 
 
 datasets_or_projects_created_7_days()
--- a/dags/synapse-datasets-created-7-days.py
+++ b/dags/synapse-datasets-created-7-days.py
@@ -60,7 +60,7 @@ class EntityCreated:
     Attributes:
         name: The name of this entity.
         id: The unique immutable ID for this entity.
-        parent_id: parent id of the entity
+        project_id: project id of the entity
         node_type: Type of Node
         content_type: Content type of folder annotated,
         created_on: The date this entity was created.
@@ -71,7 +71,7 @@ class EntityCreated:
 
     name: str
     id: int
-    parent_id: int
+    project_id: int
     node_type: str
     content_type: str
     created_on: str
@@ -113,7 +113,7 @@ def datasets_or_projects_created_7_days() -> None:
         SELECT 
             name,
             n.id,
-            parent_id,
+            project_id,
             node_type,
             ARRAY_TO_STRING(annotations:annotations:contentType:value, ', ') as content_type,
             TO_DATE(n.created_on) as entity_created_date,
@@ -149,7 +149,7 @@ def datasets_or_projects_created_7_days() -> None:
                     name=row["NAME"],
                     id=row["ID"],
                     node_type=row["NODE_TYPE"],
-                    parent_id=row["PARENT_ID"],
+                    project_id=row["PROJECT_ID"],
                     content_type=row["CONTENT_TYPE"],
                     created_on=row["ENTITY_CREATED_DATE"],
                     user_name_full_name=row["USER_NAME_FULL_NAME"],
@@ -198,7 +198,7 @@ def datasets_or_projects_created_7_days() -> None:
                 [
                     row.name,
                     row.id,
-                    row.parent_id,
+                    row.project_id,
                     row.node_type,
                     row.content_type,
                     row.created_on,
--- a/dags/synapse-datasets-created-7-days.py
+++ b/dags/synapse-datasets-created-7-days.py
@@ -64,6 +64,7 @@ class EntityCreated:
         node_type: Type of Node
         content_type: Content type of folder annotated,
         created_on: The date this entity was created.
+        full_name: Full name of the user that created this entity
         created_by: The ID of the user that created this entity.
         is_public: True to indicate if this entity is public.
     """
--- a/dags/synapse-datasets-created-7-days.py
+++ b/dags/synapse-datasets-created-7-days.py
@@ -119,10 +119,10 @@ def datasets_or_projects_created_7_days() -> None:
         message = f":synapse: Datasets or projects created in the last 7 days \n\n"
         for index, row in enumerate(entity_created):
             if row.content_type:
-                type = row.content_type.strip("[] \n").strip('"')
+                data_type = row.content_type.strip("[] \n").strip('"')
             else:
-                type = row.node_type
-            message += f"{index+1}. <https://www.synapse.org/#!Synapse:syn{row.id}|*{row.name}*> (Type: {type}, Created on: {row.created_on}, Created by: <https://www.synapse.org/Profile:{row.created_by}/profile|this user>, Public: {row.is_public})\n\n"
+                data_type = row.node_type
+            message += f"{index+1}. <https://www.synapse.org/#!Synapse:syn{row.id}|*{row.name}*> (Type: {data_type}, Created on: {row.created_on}, Created by: <https://www.synapse.org/Profile:{row.created_by}/profile|this user>, Public: {row.is_public})\n\n"
         return message
 
     @task
--- a/dags/synapse-datasets-created-7-days.py
+++ b/dags/synapse-datasets-created-7-days.py
@@ -175,7 +175,7 @@ def datasets_or_projects_created_7_days() -> None:
     def post_slack_messages(message: str) -> bool:
         """Post the top downloads to the slack channel."""
         client = WebClient(token=Variable.get("SLACK_DPE_TEAM_BOT_TOKEN"))
-        result = client.chat_postMessage(channel="hotdrop_test", text=message)
+        result = client.chat_postMessage(channel="hotdrops", text=message)
         print(f"Result of posting to slack: [{result}]")
         return result is not None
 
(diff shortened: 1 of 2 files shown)


dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,173 @@
...
+def datasets_or_projects_created_7_days() -> None:
+    """Execute a query on Snowflake and report the results to a Synapse table."""
+
+    @task
+    def get_datasets_projects_created_7_days(**context) -> List[EntityCreated]:
+        """Execute the query on Snowflake and return the results."""
+        snow_hook = SnowflakeHook(context["params"]["snowflake_conn_id"])
+        ctx = snow_hook.get_conn()
+        cs = ctx.cursor()
+        query = f"""
```
linglp, line 112: @BWMac I saw that other queries like [this one](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/dags/top-public-synapse-projects-all-time-from-snowflake.py#L65) so I also kept my f string here. Is there any reason that we should use a f string in query? I am not inserting anything in the query.

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,173 @@
...
+    @task
+    def generate_slack_message(entity_created: List[EntityCreated], **context) -> str:
+        """Generate the message to be posted to the slack channel."""
+        message = f":synapse: Datasets or projects created in the last 7 days \n\n"
+        for index, row in enumerate(entity_created):
+            if row.content_type:
+                data_type = row.content_type.strip("[] \n").strip('"')
+            else:
+                data_type = row.node_type
+            message += f"{index+1}. <https://www.synapse.org/#!Synapse:syn{row.id}|*{row.name}*> (Type: {data_type}, Created on: {row.created_on}, Created by: <https://www.synapse.org/Profile:{row.created_by}/profile|this user>, Public: {row.is_public})\n\n"
```
linglp: Note: the result of this message is something like: 
[SMARTER](https://www.synapse.org/#!Synapse:syn64831041) (Type: project, Created on: 2025-02-20, Created by: [this user](https://www.synapse.org/Profile:3531708/profile), Public: False)

dags/synapse-datasets-created-7-days.py
BWMac: If you aren't inserting anything, then no.

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,173 @@
...
+                    created_by=row["CREATED_BY"],
+                    is_public=row["IS_PUBLIC"],
+                )
+            )
+        return entity_created
+
+    @task
+    def generate_slack_message(entity_created: List[EntityCreated], **context) -> str:
+        """Generate the message to be posted to the slack channel."""
+        message = f":synapse: Datasets or projects created in the last 7 days \n\n"
```
BWMac: NIT: This seems like it doesn't need to be an f-string

dags/synapse-datasets-created-7-days.py
BWMac: The vast majority of created projects being exposed in the Slack [messages](https://sagebionetworks.slack.com/archives/C08F6SKK6Q3/p1740601878184249) are not public. Is this intentional? Should this only be limited to public datasets?

dags/synapse-datasets-created-7-days.py
BWMac: You aren't inserting anything at the moment, but you should be inserting the `current_date` into the query like the other DAGs do. That way we can run this DAG for past dates if we need to backfill data.

dags/synapse-datasets-created-7-days.py
linglp: hmm based on my conversation with @thomasyu888 , we just need to get all the data for the first iteration. Maybe for the next iteration, we will only look at public projects/datasets

dags/synapse-datasets-created-7-days.py
linglp: Good catch. Thank you

dags/synapse-datasets-created-7-days.py
linglp: In the `push_results_to_synapse_table` function, I already added today's date as 'export date.' Is that sufficient, or do we still need to insert today's date in the query here?

BWMac: No we should insert the date into the query so that we can pull data for dates other than just today.

BWMac: and the date in the Synapse table should match the date that we are pulling data for.

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,173 @@
...
+
+SYNAPSE_RESULTS_TABLE = "syn64919238"
+
+
+@dataclass
+class EntityCreated:
+    """Dataclass to record
+
+    Attributes:
+    name: The name of this entity.
```
BryanFauble: Nit. These should be tabbed in once more under Atributes

BryanFauble (changes requested):
Only issue i see is this comment chain here:

https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/75/files#r1972395892

Once you fix feel free to re-request the review

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,173 @@
...
+            TO_DATE(created_on) as entity_created_date,
+            created_by,
+            is_public,
+        FROM 
+            synapse_data_warehouse.synapse.node_latest
+        WHERE 
+            (
+                node_type IN ('dataset', 'project') 
+                OR 
+                (content_type LIKE '%dataset%' AND content_type IS NOT NULL)
```
jaymedina: nit: you can remove this `AND content_type IS NOT NULL` as it's redundant when following the LIKE statement

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,173 @@
...
+            )
+        AND 
+            (
+                entity_created_date >= CURRENT_TIMESTAMP - INTERVAL '7 days' 
+            )
+        ORDER BY entity_created_date;
+        """
+        print(query)
```
jaymedina, line 141: Was this print statement for troubleshooting?

dags/synapse-datasets-created-7-days.py
jaymedina: For the query to be more robust, I would use `ILIKE` instead of `LIKE` so the string isn't case-sensitive. [Here](https://docs.snowflake.com/en/sql-reference/functions/ilike_any) is documentation for `ILIKE`.

dags/synapse-datasets-created-7-days.py
BryanFauble: IMO this is probably fine. The consequence is that it'll be a message that gets printed to the Airflow logs, but there isn't anything sensitive here. It can be the first part of investigating an issue if something pops up (making sure the query is correct).

linglp: I saw some other files, such as synapse-by-the-numbers-dag have the print statement so I kept it here. I think this might help with trouble shooting if we are going to insert backfill date in the query

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,173 @@
...
+        )
+
+    entity_created = get_datasets_projects_created_7_days()
+    push_to_synapse_table = push_results_to_synapse_table(entity_created=entity_created)
+    slack_message = generate_slack_message(entity_created=entity_created)
+    post_to_slack = post_slack_messages(message=slack_message)
+
+    entity_created >> slack_message
+    slack_message >> post_to_slack
+    entity_created >> push_to_synapse_table
```
jaymedina, line 170: typo: `push_results_to_synapse_table`*

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,173 @@
...
+            synapse_data_warehouse.synapse.node_latest
+        WHERE 
+            (
+                node_type IN ('dataset', 'project') 
+                OR 
+                (content_type LIKE '%dataset%' AND content_type IS NOT NULL)
+            )
+        AND 
+            (
+                entity_created_date >= CURRENT_TIMESTAMP - INTERVAL '7 days'
```
jaymedina: Did you want to modify this so it uses the date from the `dag_params` dictionary?

jaymedina (commented):
Looks great! Just a few comments, and I agree with @BWMac in replacing `CURRENT_TIMESTAMP` in the query so you can input any date, not just the current day.

dags/synapse-datasets-created-7-days.py
jaymedina: Fine with me!

dags/synapse-datasets-created-7-days.py
linglp: Okay I will do that. Thanks for the suggestion.

dags/synapse-datasets-created-7-days.py
linglp: Yes. Let me do that. Thank you

dags/synapse-datasets-created-7-days.py
linglp: Thanks for the suggestion, Jenny. I decided to just modify the query by doing something like: `ARRAY_TO_STRING(annotations:annotations:contentType:value, ', ') as content_type` and I could just do `content_type = 'dataset'`

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,221 @@
...
+    def push_results_to_synapse_table(
+        entity_created: List[EntityCreated], **context
+    ) -> None:
+        """Push the results to a Synapse table."""
+        data = []
+        # convert context["params"]["backfill_date_time"] to date in same format as date.today()
+        today = (
+            date.today()
+            if not context["params"]["backfill"]
+            else datetime.strptime(
```
linglp, line 192: @BWMac if I set the back fill date parameter in airflow like how I set it up now, by default is using the UTC time zone. But when the table gets uploaded to synapse, synapse would use the user's local time zone. Is that okay?

BWMac: Yes, that is how the other DAGs work. See this explanation in one of the module docstrings [here](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/57a03fa128d9365be239d0ceda4076fc7647b135/dags/top-public-synapse-projects-from-snowflake.py#L5-L10)

linglp: Got it. Thanks. Should I also add that docstring that you link to this file?

BWMac: Yeah that's probably a good idea!

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,221 @@
+"""This script executes a query on Snowflake to retrieve data about entities created in the past 7 days and stores the results in a Synapse table.
+It is scheduled to run every Monday at 00:00 UTC.
```
BWMac, line 2: NIT: Add info about the slack-posting functionality of the DAG

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,221 @@
...
+        is_public: True to indicate if this entity is public.
+    """
+
+    name: str
+    id: int
+    parent_id: int
+    node_type: str
+    content_type: str
+    created_on: str
+    user_name: str
```
BryanFauble: `user_name` isn't included in the docstring of the class.

Also, this is a little mis-leading. It isn't actually the username, it's their `full_name`, or `first_last_name`. This, along with the query should probably be updated (And maybe the Synapse table column name?)

Relevant line of SQL:
`CONCAT(first_name,' ',last_name) AS user_name,`

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,221 @@
...
+                    row.name,
+                    row.id,
+                    row.node_type,
+                    row.parent_id,
+                    row.content_type,
+                    row.created_on,
+                    row.created_by,
+                    row.is_public,
+                    today,
+                ]
```
BryanFauble: Should the Synapse table include all attributes from the `EntityCreated`? Totally fine if not, but if it should, the first name/last name is missing.

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,221 @@
...
+
+    entity_created = get_datasets_projects_created_7_days()
+    check = check_backfill()
+    stop = stop_dag()
+    push_to_synapse_table = push_results_to_synapse_table(entity_created=entity_created)
+    slack_message = generate_slack_message(entity_created=entity_created)
+    post_to_slack = post_slack_messages(message=slack_message)
+
+    entity_created >> check >> [stop, slack_message]
+    entity_created >> slack_message
```
BryanFauble: I think this line (`entity_created >> slack_message`) should be removed. Since above you are already creating a directed edge to call `slack_message` based off the conditional present in `check`: `entity_created >> check >> [stop, slack_message]`

BryanFauble (changes requested):
There are a few more items I flagged

dags/synapse-datasets-created-7-days.py
linglp: Thanks Bryan! `full_name` is reasonable.

dags/synapse-datasets-created-7-days.py
linglp: Got it. I will add the docstring.

dags/synapse-datasets-created-7-days.py
linglp: I added full_name so that the slack message would be easier to read. But I don't think it matters for the synapse table?

BryanFauble: It depends on what the requirements of the ticket are

linglp: Okay let me ask that in the Jira ticket.

dags/synapse-datasets-created-7-days.py
linglp: Thanks for the comment, Brad! I will mark this as resolved now.

dags/synapse-datasets-created-7-days.py
linglp: Have gotten the clarification from Tom, and he said that we don't need to include that in the synapse table. I will mark this as "resolved" then.

dags/synapse-datasets-created-7-days.py
linglp: Got it. Thanks

BryanFauble (approved):
LGTM!

BWMac (approved):
🔥 🔥 🔥 🔥 🔥 🔥

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,228 @@
...
+            date.today()
+            if not context["params"]["backfill"]
+            else datetime.strptime(
+                context["params"]["backfill_date_time"], "%Y-%m-%d  %H:%M:%S"
+            ).date()
+        )
+        for row in entity_created:
+            data.append(
+                [
+                    row.name,
```
thomasyu888, line 199: This can be excluded

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,228 @@
...
+                    row.id,
```
thomasyu888, line 200: If you add "syn" in front you can render the entityid column.

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,228 @@
...
+                    row.parent_id,
```
thomasyu888, line 201: Instead of parent_id, can we use project_id?

dags/synapse-datasets-created-7-days.py
```diff
@@ -0,0 +1,228 @@
...
+        backfill_date_time = context["params"]["backfill_date_time"]
+        query_date = (
+            None if not context["params"]["backfill"] else f"{backfill_date_time}"
+        )
+
+        query = f"""
+        SELECT 
+            name,
+            n.id,
+            parent_id,
```
thomasyu888: Instead of parent_id, can we use project_id?

thomasyu888 (approved):
🔥 LGTM!  Just one small comment, and please switch to the #hotdrops channel instead of the test channel!

dags/synapse-datasets-created-7-days.py
linglp: Sure. I could update to use project_id.

PR #77, opened 2025-03-14
[DPE-1234] Create synapse dataset collection to croissant DAG

BryanFauble:
# **Problem:**

- There is a need as discussed in this design doc to create an automated process for converting a Synapse dataset/dataset collection into the croissant format

# **Solution:**

- Creating an airflow DAG that will run weekly on Mondays that will query for datasets, and the files contains within those datasets in order to create the croissant file format.

# **Testing:**

- I ran a codespace instance in this github repo/branch and verified that data could be queried, and created
- I verified that I could upload and update existing files in a test S3 bucket:
image

- I also verified that I was able to push telemetry (Trace/log) data out to SigNoz:

image

### DIFF
--- a/dags/synapse-dataset-to-croissant.py
+++ b/dags/synapse-dataset-to-croissant.py
@@ -324,6 +324,10 @@ def dataset_to_croissant() -> None:
             croissant_json_ld_for_dataset = cs.fetch_pandas_all()
             cs.close()
 
+            if croissant_json_ld_for_dataset.empty:
+                print(f"No content found for {dataset_id}")
+                return
+
             synapse_id = croissant_json_ld_for_dataset["DATASET_SYNAPSE_ID"][0]
             name = croissant_json_ld_for_dataset["NAME"][0]
             push_to_s3 = context["params"]["push_results_to_s3"]
@@ -347,6 +351,7 @@ def dataset_to_croissant() -> None:
                 s3_hook.load_file_obj(file_obj=metadata_file,
                                       key=f"{name}_{synapse_id}_croissant.jsonld",
                                       bucket_name=BUCKET_NAME,
+                                      replace=True,
                                       )
 
         except Exception as e:
--- a/dags/synapse-dataset-to-croissant.py
+++ b/dags/synapse-dataset-to-croissant.py
@@ -1,0 +1,363 @@
+"""
+This DAG queries snowflake for the given dataset collections to look for datasets.
+
+For each dataset in the dataset collections, it queries the dataset in Snowflake and 
+pushes the results to a Croissant file in GitHub.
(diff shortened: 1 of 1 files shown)


dags/synapse-dataset-to-croissant.py
```diff
@@ -0,0 +1,563 @@
...
+            WITH entity_metadata AS (
+                SELECT
+                    items,
+                    name,
+                    ANNOTATIONS,
+                    CREATED_ON,
+                    VERSION_NUMBER,
+                    CONCAT('syn', id) as DATASET_SYNAPSE_ID
+                FROM synapse_data_warehouse.synapse.node_latest
+                WHERE id = {dataset_id} -- This is the ID of a Dataset
```
BWMac: NIT: query comment is redundant to the variable name

dags/synapse-dataset-to-croissant.py
```diff
@@ -0,0 +1,563 @@
...
+        with otel_tracer.start_as_current_span("query_snowflake_and_push_croissant_file", context=TraceContextTextMapPropagator().extract(root_carrier_context)) as span:
+            if not dataset_id:
+                otel_logger.info("No dataset found")
+                otel_tracer.span_processor.force_flush()
+                otel_logger.handlers[0].flush()
+                return
+
+            span.set_attribute("airflow.dataset_id", f"syn{dataset_id}")
+
+            query = f"""
```
BWMac, line 318: I wonder as we continue performing data operations with Airflow + Snowflake if there is a better pattern for storing/using queries that we should consider. This one is pretty complex and having it just be a long string block within Python code feels kind of brittle.

BWMac (approved):
LGTM

dags/synapse-dataset-to-croissant.py
jaymedina: We could abstract recurring queries/CTEs into [user-defined functions](https://docs.snowflake.com/en/sql-reference/sql/create-function) or stored procedures, although it wouldn't be much help for queries with very specific use-cases like this one, and I'm not a fan of abstracting things if it makes the query harder to debug. But still this is a good point and I have an [exploration ticket](https://sagebionetworks.jira.com/browse/SNOW-234) on this where we can add more robust DAGs as a motivator.

BryanFauble: Thanks for the comments. To me, it seems to come down to whether we need to be able to run the queries outside the context on this DAG.

At the moment I do not see that need, however, based on Jenny's ticket we could start to define the standard for introducing snowflake functions in a standardized way. At that point I think we could then port the queries over and refactor the DAGs to call them.

PR #83, opened 2025-04-02
[DPE-1275] Move to Synapse API, Automate deletion of S3 objects/Synapse Rows, File/Dataset versioning corrections

BryanFauble:
# **Problem:**

- We were using the Snowflake datawarehouse to source information in order to build this process. Per [these guidelines](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4043079690/Data+Warehouse+Usage+Guidelines+policies) we want to limit usage of the warehouse.
- Deleting S3 objects that were no longer present in the dataset collections was a manual process
- We did not capture version information for the file entities
- We did not persist S3 object links back to a Synapse table to use for downstream processes

# **Solution:**

- Moving all (Except MD5 retrieval) reads to occur via unauthenticated/anonymous Synapse APIs
- Adding logic to delete out S3 objects which are no longer present in the dataset collections given.
- Adding version information for the croissant file
- Adding support to push s3 object links to a Synapse table
- Support deleting rows out of the Synapse table which should no longer show

# **Testing:**

- I verified that I could run this DAG in github codespaces with all params set to True
- Data is pushed to this (INTERNAL TO SAGE FOR NOW) table: https://www.synapse.org/Synapse:syn65903895/tables/

TODO: 

- [x] Naming convention changes for the croissant spec file are ok
- [x] The JSON LD file contains all expected changes
- [x] Push the links to the S3 objects back into a Synapse table that has columns for: Synapse ID of the Dataset collection, Synapse ID of the Dataset, Version of the Dataset, Link to the S3 object

### DIFF
--- a/dags/synapse-dataset-to-croissant.py
+++ b/dags/synapse-dataset-to-croissant.py
@@ -487,6 +487,12 @@ def extract_s3_objects_to_delete(bucket_objects: List[str], dataset_collections_
         dataset_collections_to_consider_for_deletion: The dataset collections to
             consider for deletion. This is passed in from the
             `dataset_collections_for_cleanup` DAG parameter.
+        combined_dataset_collection_and_datasets: The combined list of dataset IDs,
+            dataset collections, and dataset versions to use to determine which
+            datasets to delete.
+
+    Returns:
+        The list of objects to delete from the S3 bucket.
     """
     objects_to_delete = []
 
--- a/dags/synapse-dataset-to-croissant.py
+++ b/dags/synapse-dataset-to-croissant.py
@@ -523,10 +523,17 @@ def extract_s3_objects_to_delete(bucket_objects: List[str], dataset_collections_
         match_found_for_object = False
 
         for combined_dataset_collection_and_dataset in combined_dataset_collection_and_datasets:
-            if (dataset_collection == combined_dataset_collection_and_dataset["dataset_collection"] and
-                    dataset_id == combined_dataset_collection_and_dataset["dataset_id"] and
-                    dataset_version == str(combined_dataset_collection_and_dataset["dataset_version"]) and
-                    dataset_name == combined_dataset_collection_and_dataset["dataset_name"]):
+
+            matches_dataset_collection = dataset_collection == combined_dataset_collection_and_dataset[
+                "dataset_collection"]
+            matches_dataset_id = dataset_id == combined_dataset_collection_and_dataset[
+                "dataset_id"]
+            matches_dataset_version = dataset_version == str(
+                combined_dataset_collection_and_dataset["dataset_version"])
+            matches_dataset_name = dataset_name == combined_dataset_collection_and_dataset[
+                "dataset_name"]
+
+            if matches_dataset_collection and matches_dataset_id and matches_dataset_version and matches_dataset_name:
                 match_found_for_object = True
                 break
         if not match_found_for_object:
@@ -636,7 +643,7 @@ def execute_push_to_s3(dataset: Entity, dataset_id: str, dataset_version: str, s
 def execute_push_to_synapse(push_to_synapse: bool, dataset: Entity, dataset_id: str, dataset_version: str, dataset_collection: str, s3_url: str, **context) -> None:
     """
     Handle the push to Synapse of the croissant file link. This is done by using
-    an unauthenticated Synapse client to first query the table to determine if an
+    an authenticated Synapse client to first query the table to determine if an
     update is needed. If the link already exists with the expected S3 URL, then
     skip the update. If the link does not exist or the S3 URL is different, then
     update the link with the new S3 URL using the authenticated Synapse client.
@@ -993,7 +1000,14 @@ def dataset_to_croissant() -> None:
         transformation and push the results to S3.
 
         Arguments:
-            dataset_id: The ID of the dataset to query in Snowflake.
+            root_carrier_context: The root carrier context to use for the trace context.
+            dataset_id: The ID of the dataset to query.
+            dataset_version: The version of the dataset to query.
+            dataset_collection: The ID of the dataset collection to query.
+            context: The context of the DAG run.
+
+        Returns:
+            None
         """
         with otel_tracer.start_as_current_span("query_and_push_croissant_file", context=TraceContextTextMapPropagator().extract(root_carrier_context)) as span:
             push_to_s3 = context["params"]["push_results_to_s3"]
@@ -1017,11 +1031,15 @@ def dataset_to_croissant() -> None:
                     f"{dataset_id}.{dataset_version}", downloadFile=False)
                 span.set_attribute("airflow.dataset_name", dataset.name)
 
+                # The dataset entity in synapse returned a list of files that are
+                # attached to the dataset. This is used to get the file IDs and
+                # versions of the files that are attached to the dataset.
                 if not hasattr(dataset, "datasetItems") or not dataset.get("datasetItems"):
                     span.set_attribute("airflow.croissant_result", False)
                     otel_logger.warning(
                         f"No files found for dataset {dataset_id}.{dataset_version}")
                     return
+
                 dataset_items = dataset.get("datasetItems")
                 for dataset_item in dataset_items:
                     syn_id = dataset_item["entityId"]
@@ -1039,6 +1057,7 @@ def dataset_to_croissant() -> None:
                 if path_to_remove:
                     os.remove(path_to_remove)
 
+            # Sort so we have a consistent order for the data shown in the croissant file
             file_ids_and_versions_attached_to_dataset.sort(
                 key=lambda x: x["file_id"])
 
@@ -1054,6 +1073,10 @@ def dataset_to_croissant() -> None:
 
             span.set_attribute("airflow.croissant_result", True)
 
+            # The distribution section of the croissant file is used to describe the
+            # files that are attached to the dataset. The first `metadata` file is
+            # used to describe the metadata associated with the dataset and where to find
+            # it on the Synapse server (The Dataset view)
             distribution_files = [{
                 "@type": "FileObject",
                 "@id": "metadata",
--- a/dags/synapse-dataset-to-croissant.py
+++ b/dags/synapse-dataset-to-croissant.py
@@ -1,7 +1,23 @@
 """
-This DAG queries snowflake for the given dataset collections to look for datasets.
+This DAG interacts with a few different services to accomplish the following:
 
-For each dataset in the dataset collections, it queries the dataset in Snowflake and
+- (Synapse, Unauthenticated) Given list of dataset collections query Synapse to retrieve Anonymously accessible datasets
+- (Synapse, Unauthenticated) For each dataset in the dataset collections, query Synapse Anonymously to retrieve metadata about each dataset
+- (Synapse, Unauthenticated) For each Anonymously accessible dataset, query Synapse Anonymously to retrieve the files attached to the dataset
+- (Snowflake, Authenticated) For each file which is not Anonymously downloadable, query Snowflake to retrieve the content_md5 (Authenticated calls to the datawarehouse)
+- (S3, Authenticated) For each dataset push an object a public S3 bucket in the `org-sagebase-dpe-prod` AWS account
+- (Synapse, Unauthenticated) For each dataset Anonymously query a Synapse table which contains links to the S3 object
+- (Synapse, Authenticated) Delete rows from the Synapse table which are not in a dataset collection
+- (Synapse, Aauthenticated) For each dataset push a link for the S3 object to a Synapse table via the authenticated Synapse client
+- (S3, Authenticated) Delete S3 objects from the S3 bucket which are not in a dataset collection
+
+
+To add more datasets to be proceed by this dag:
+Simply add the dataset collections to the `dataset_collections` parameter in the DAG to
+run this process for the given dataset collections.
+
+
+For each dataset in the dataset collections, it queries the dataset in Synapse and
 pushes the results to a Croissant file in S3. The S3 bucket is stored in the
 `org-sagebase-dpe-prod` AWS account. That S3 bucket is deployed to AWS through the code
 in this PR: https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/57 .
@@ -11,38 +27,30 @@ log in as an AWS user to accomplish the required work. In order for this DAG to 
 user that is running the DAG must have the correct permissions to access the S3 bucket
 and write to it. This was set up on the `airflow-secrets-backend` user in the
 `org-sagebase-dpe-prod` AWS account. The user has an inline policy to grant PutObject,
-GetObject, and ListBucket permissions to the S3 bucket.
-
-Simply add the dataset collections to the `dataset_collections` parameter in the DAG to
-run this process for the given dataset collections.
+GetObject, DeleteObject, and ListBucket permissions to the S3 bucket.
 
 In addition this DAG has been set up with the first iteration of the OpenTelemetry
 integration for Apache Airflow. OTEL support is officially added in 2.10.0, but requires
-that we also upgrade the production Airflow server.
+that we also set up the production Airflow server to support OTEL.
 
 DAG Parameters:
-- `snowflake_developer_service_conn`: A JSON-formatted string containing the connection details required to authenticate and connect to Snowflake.
-- `synapse_conn_id`: The connection ID for the Synapse connection.
-- `dataset_collections`: The dataset collections to query for datasets.
-- `push_results_to_s3`: A boolean to indicate if the results should be pushed to S3.
-                        When set to `False`, the results will be printed to the logs.
-- `delete_out_of_date_from_s3`: A boolean to indicate if the old files should be 
-                        deleted from S3. When set to `False`, the old files will not 
-                        be deleted, but a message will be logged to indicate that
-                        the files will be deleted when set to `True`.
-- `aws_conn_id`: The connection ID for the AWS connection. Used to authenticate with S3.
+- Review the DAG Parameters under the `@dag` decorated function
 """
 
 import json
 import os
+import re
+import tempfile
 from datetime import datetime
 from io import BytesIO
 from logging import NOTSET, Logger, getLogger
-from typing import Dict, List
+from types import MethodType
+from typing import Any, Dict, List
+from urllib.parse import quote_plus
 
 from airflow.decorators import dag, task
-from airflow.models.param import Param
 from airflow.models import Variable
+from airflow.models.param import Param
 from airflow.providers.amazon.aws.hooks.s3 import S3Hook
 from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
 from opentelemetry import context as otel_context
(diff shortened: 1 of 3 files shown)


dags/synapse-dataset-to-croissant.py
```diff
@@ -165,14 +178,248 @@ def set_up_logging() -> Logger:
...
+    if files_to_find_md5_in_snowflake:
+        snow_hook = SnowflakeHook(
+            context["params"]["snowflake_developer_service_conn"])
+        ctx = snow_hook.get_conn()
+        cs = ctx.cursor()
+        ids_of_files = list(files_to_find_md5_in_snowflake.keys())
+        id_and_version_pairs = []
+        for file_id, version in files_to_find_md5_in_snowflake.items():
+            id_and_version_pairs.append(file_id)
+            id_and_version_pairs.append(version)
```
BryanFauble: TODO: This code will not work if a Dataset contains 2 versions of the same file (Synapse id). Some logical changes are required here

dags/synapse-dataset-to-croissant.py
```diff
@@ -567,8 +778,14 @@ def query_snowflake_and_push_croissant_file(root_carrier_context: Dict, dataset_
     datasets = query_synapse_dataset_collection_for_datasets.partial(root_carrier_context=root_carrier_context).expand_kwargs(
         get_dataset_collections(root_carrier_context=root_carrier_context))
 
+    combined_dataset_lists = combine_dataset_lists(
```
thomasyu888, line 1199: Should we just upload a synapse table with the mapping?

Table name: croissant-files
schema:
dataset_id, croissant_link?

That does complicate things a bit, because then you'd need authentication, we can have a separate ticket to do this.

dags/synapse-dataset-to-croissant.py
```diff
@@ -308,236 +613,142 @@ def query_snowflake_and_push_croissant_file(root_carrier_context: Dict, dataset_
...
+                    "sc": "https://schema.org/",
+                    "separator": "cr:separator",
+                    "source": "cr:source",
+                    "subField": "cr:subField",
+                    "transform": "cr:transform"
+                },
+                "@type": "Dataset",
+                "@id": f"{dataset_id}.{dataset_version}",
+                "name": f"{dataset.name}",
+                "description": f"Dataset for {dataset.name}",
```
thomasyu888: Datasets actually also have a "description" property.  it might be helpful to pull that down?

dags/synapse-dataset-to-croissant.py
BryanFauble: Yeah. I was intending to knock this out. Use the existing service account to push out data to a synapse table.

Schema would need to be:
Dataset syn id
Dataset Version
Dataset collection syn id
Croissant file s3 link

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+            WHERE nl.id IN ({', '.join(['%s'] * len(ids_of_files))})
+            AND nl.is_public = TRUE
+        )
+        SELECT
+            vd.id,
+            vd.versionNumber,
+            vd.contentMd5
+        FROM version_data AS vd
+        WHERE (vd.id, vd.versionNumber) IN ({', '.join(['(%s, %s)'] * (len(id_and_version_pairs)//2))});
+        """
```
BryanFauble: This portion is the only usage of Snowflake in this script

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+        raise ex
+
+
 @dag(**dag_config)
 def dataset_to_croissant() -> None:
-    """Execute a query on Snowflake and report the results to a Synapse table."""
+    """Execute a query on Snowflake and report the results to a Synapse table.
+
+
+    DAG Parameters:
```
BryanFauble, line 732: By Adding this to this docstring it will show up in the airflow UI now!

image

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+            vd.id,
+            vd.versionNumber,
+            vd.contentMd5
+        FROM version_data AS vd
+        WHERE (vd.id, vd.versionNumber) IN ({', '.join(['(%s, %s)'] * (len(id_and_version_pairs)//2))});
+        """
+        try:
+            cs.execute(
+                query,
+                (ids_of_files + id_and_version_pairs),
```
thomasyu888, line 370: I wonder if snowflake has limits to the number of `value, value2` in the query line

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+            )
+            file_md5s = cs.fetch_pandas_all()
+        finally:
+            cs.close()
+
+    for file in files_attached_to_dataset:
+        file: File = file
+        if file.file_handle and file.file_handle.content_md5:
+            file_md5 = file.file_handle.content_md5
```
thomasyu888: Is this just the case where we download the file?

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+            file_md5 = "unknown_md5"
+
+        distribution_files.append(
+            {
+                "@type": "FileObject",
+                "@id": f"{file.id}.{file.version_number}",
+                "name": f"{file.name}",
+                "description": f"Data file associated with {file.name}",
+                "contentUrl": f"https://www.synapse.org/Synapse:{file.id}.{file.version_number}",
+                "encodingFormat": "application/json",
```
thomasyu888: Is this encoding format for the associated file...?

dags/synapse-dataset-to-croissant.py
BryanFauble: We are not downloading the file in this case, but we can get the MD5 from synapse in some cases. Those cases are when the file can be anonymously downloaded. In other cases we need to pull the data from Snowflake

dags/synapse-dataset-to-croissant.py
BryanFauble: Good point. I added it to the ticket and will look to pull the right format based on the file.

dags/synapse-dataset-to-croissant.py
```diff
@@ -1,43 +1,56 @@
...
+- (Synapse, Unauthenticated) For each Anonymously accessible dataset, query Synapse Anonymously to retrieve the files attached to the dataset
+- (Snowflake, Authenticated) For each file which is not Anonymously downloadable, query Snowflake to retrieve the content_md5 (Authenticated calls to the datawarehouse)
+- (S3, Authenticated) For each dataset push an object a public S3 bucket in the `org-sagebase-dpe-prod` AWS account
+- (Synapse, Authenticated) For each dataset query a Synapse table which contains links to the S3 object
+- (Synapse, Authenticated) Delete rows from the Synapse table which are not in a dataset collection
+- (Synapse, Authenticated) For each dataset push a link for the S3 object to a Synapse table via the authenticated Synapse client
+- (S3, Authenticated) Delete S3 objects from the S3 bucket which are not in a dataset collection
+
+
+To add more datasets to be proceed by this dag:
```
BWMac:
```suggestion
To add more datasets to be processed by this dag:
```

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+        dataset_version = dataset_version.replace("_", "")
+        dataset_name = bucket_object.split(dataset_id_and_version.group(0))[0]
+
+        match_found_for_object = False
+
+        for combined_dataset_collection_and_dataset in combined_dataset_collection_and_datasets:
+            if (dataset_collection == combined_dataset_collection_and_dataset["dataset_collection"] and
+                    dataset_id == combined_dataset_collection_and_dataset["dataset_id"] and
+                    dataset_version == str(combined_dataset_collection_and_dataset["dataset_version"]) and
+                    dataset_name == combined_dataset_collection_and_dataset["dataset_name"]):
```
BWMac: NIT: This way might be a bit more readable
```suggestion
            if all([
                dataset_collection == combined_dataset_collection_and_dataset["dataset_collection"],
                dataset_id == combined_dataset_collection_and_dataset["dataset_id"],
                dataset_version == str(combined_dataset_collection_and_dataset["dataset_version"]),
                dataset_name == combined_dataset_collection_and_dataset["dataset_name"]
            ]):
```

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+        dataset_id = synapse_row.dataset
+        dataset_version = synapse_row.dataset_version
+        row_id = synapse_row.ROW_ID
+
+        match_found_for_object = False
+
+        for combined_dataset_collection_and_dataset in combined_dataset_collection_and_datasets:
+            if (dataset_collection == combined_dataset_collection_and_dataset["dataset_collection"] and
+                    dataset_id == combined_dataset_collection_and_dataset["dataset_id"] and
+                    dataset_version == combined_dataset_collection_and_dataset["dataset_version"]):
```
BWMac, line 581: Same as above

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+        # Warning: Using an authenticated Synapse Client during this section of code
+        syn_hook = SynapseHook(
+            context["params"]["synapse_conn_id"])
+        authenticated_syn_client: Synapse = syn_hook.client
+        authenticated_syn_client._rest_call = MethodType(
+            _rest_call_replacement, authenticated_syn_client)
+        existing_row = authenticated_syn_client.tableQuery(
+            query=f"SELECT * FROM {SYNAPSE_TABLE_FOR_CROISSANT_LINKS} WHERE dataset = '{dataset_id}' AND dataset_version = {dataset_version} AND dataset_collection = '{dataset_collection}'", resultsAs="csv")
+        existing_row_df = existing_row.asDataFrame()
+        os.remove(existing_row.filepath)
```
BWMac: Any concern about some of these temporary files not being `remove`-d if an error happens?

BWMac (approved):

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+            "Failed to query snowflake and push croissant file to S3.")
+        otel_tracer.span_processor.force_flush()
+        otel_logger.handlers[0].flush()
+        raise ex
+
+
+def execute_push_to_synapse(push_to_synapse: bool, dataset: Entity, dataset_id: str, dataset_version: str, dataset_collection: str, s3_url: str, **context) -> None:
+    """
+    Handle the push to Synapse of the croissant file link. This is done by using
+    an unauthenticated Synapse client to first query the table to determine if an
```
SageGJ: It looks like this method is currently using an authenticated client

dags/synapse-dataset-to-croissant.py
BryanFauble: I couldn't find any limit on the number of placeholders that may be used. I did find limits that the query length can be up to 1 MB, which, I don't think we would hit even if we were using a dataset that had 30k entities on it (It'd be 90k placeholder strings). It would only be around 500 KB

dags/synapse-dataset-to-croissant.py
thomasyu888: Maybe premature simplification but would it be more simple to just pull everything from snowflake instead?

(Future ticket if yes)

dags/synapse-dataset-to-croissant.py
thomasyu888: This, imo, is a good example of PR review comment that doesn't have to be in the code comment itself

dags/synapse-dataset-to-croissant.py
```diff
@@ -292,272 +867,277 @@ def combine_dataset_lists(root_carrier_context: Dict, dataset_ids: List[List[Dic
...
+
+            if objects_to_delete:
+                delete_out_of_date_from_s3 = context["params"]["delete_out_of_date_from_s3"]
+                if delete_out_of_date_from_s3:
+                    otel_logger.info(
+                        f"Deleting the following objects from S3: {objects_to_delete}")
+                    s3_hook.delete_objects(
+                        bucket=BUCKET_NAME, keys=objects_to_delete)
+                else:
+                    otel_logger.info(
```
SageGJ, line 928: What would cause the objects to not be deleted here?

dags/synapse-dataset-to-croissant.py
```diff
@@ -292,272 +867,277 @@ def combine_dataset_lists(root_carrier_context: Dict, dataset_ids: List[List[Dic
...
+
+        Returns:
+            None
+        """
+        # Warning: Using an authenticated Synapse Client during this section of code
+        with otel_tracer.start_as_current_span("delete_non_current_files_from_s3", context=TraceContextTextMapPropagator().extract(root_carrier_context)) as span:
+            syn_hook = SynapseHook(context["params"]["synapse_conn_id"])
+            syn_client: Synapse = syn_hook.client
+            syn_client._rest_call = MethodType(
+                _rest_call_replacement, syn_client)
```
SageGJ, line 965: Since these lines are used in multiple places would it be useful to create a function like `create_syn_client` to create a client from a `SynaspseHook`, replace the appropriate method, and then return the new client?

dags/synapse-dataset-to-croissant.py
BryanFauble: Completed in https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/84/files#r2031909088

dags/synapse-dataset-to-croissant.py
BryanFauble: Done in https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/84

dags/synapse-dataset-to-croissant.py
BryanFauble: I added a small try/finally in https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/84

 I don't know what would really occur in these cases if temporary storage fills up. It might be a bridge we'll need to cross later, but I wasn't going to worry about it beyond what I had put in place so far.

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,552 @@ def set_up_logging() -> Logger:
...
+        dataset_name = bucket_object.split(dataset_id_and_version.group(0))[0]
+
+        match_found_for_object = False
+
+        for combined_dataset_collection_and_dataset in combined_dataset_collection_and_datasets:
+            if (dataset_collection == combined_dataset_collection_and_dataset["dataset_collection"] and
+                    dataset_id == combined_dataset_collection_and_dataset["dataset_id"] and
+                    dataset_version == str(combined_dataset_collection_and_dataset["dataset_version"]) and
+                    dataset_name == combined_dataset_collection_and_dataset["dataset_name"]):
+                match_found_for_object = True
```
SageGJ: The number of checks and what they all mean is difficult to intuit.
Something like
```suggestion
            dataset_collection_match = (dataset_collection == combined_dataset_collection_and_dataset["dataset_collection"])
            dataset_id_match = (dataset_id == combined_dataset_collection_and_dataset["dataset_id"])
            dataset_version_match = (dataset_version == str(combined_dataset_collection_and_dataset["dataset_version"]))                
            dataset_name_match = (dataset_name == combined_dataset_collection_and_dataset["dataset_name"])
            if all([
            dataset_collection_match ,
            dataset_id_match ,
            dataset_version_match,
            dataset_name_match ,
            ]):
                   match_found_for_object = True
```
might be easier for others to understand

dags/synapse-dataset-to-croissant.py
```diff
@@ -292,272 +867,277 @@ def combine_dataset_lists(root_carrier_context: Dict, dataset_ids: List[List[Dic
...
+            otel_logger.handlers[0].flush()
+            return None
+
+    @task
+    def query_and_push_croissant_file(root_carrier_context: Dict, dataset_id: str, dataset_version: int, dataset_collection: str, **context) -> None:
         """
         Query the Snowflake database for the dataset in order to perform the Croissant
         transformation and push the results to S3.
 
         Arguments:
```
linglp, line 1004: since there are more arguments in the parameter, add more in the doc string?

dags/synapse-dataset-to-croissant.py
```diff
@@ -292,272 +867,277 @@ def combine_dataset_lists(root_carrier_context: Dict, dataset_ids: List[List[Dic
...
-            cs = ctx.cursor()
-
-            otel_logger.info(
-                f"Performing query for dataset {dataset_id}")
+            span.set_attribute("airflow.dataset_id", f"{dataset_id}")
+            path_to_remove = None
+            syn_client = create_syn_client()
+            file_ids_and_versions_attached_to_dataset = []
 
             try:
```
linglp, line 1031: starting from this try and except block, could you add more comments in the code?

dags/synapse-dataset-to-croissant.py
BryanFauble: There is a boolean `delete_out_of_date_from_s3` that can be toggled when you run the DAG. When this is false we will be running this portion of the code in "dry run" mode. So it'll print what it would do, without doing it.

dags/synapse-dataset-to-croissant.py
BryanFauble: At the moment this isn't much of a concern IMO. It'll be fixed in the next release of the SYNPY client and we can remove this hacky code at that point

dags/synapse-dataset-to-croissant.py
BryanFauble: I added a few more! [Take a look at this commit](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/83/commits/616a6e1d8cb81aa69762f59cffc6d452a0591f91)

linglp: Great. Thanks for adding them.

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,559 @@ def set_up_logging() -> Logger:
...
+    `_datasetCollection_syn####_` will be considered for deletion. The DAG param
+    `dataset_collections_for_cleanup` is used to filter the objects that will be
+    considered for deletion as the second filter.
+
+    The last filter enforces that each object in the S3 bucket must match a dataset
+    id, dataset version, dataset collection, and dataset name in the
+    `combined_dataset_collection_and_datasets` list. If the object does not match
+    any of the objects in the list, then it will be marked for deletion.
+
+    Arguments:
```
linglp, line 478: here this is also missing a docstring for parameter `combined_dataset_collection_and_datasets`?

dags/synapse-dataset-to-croissant.py
```diff
@@ -165,17 +199,559 @@ def set_up_logging() -> Logger:
...
+        bucket_objects: The list of objects in the S3 bucket.
+        dataset_collections_to_consider_for_deletion: The dataset collections to
+            consider for deletion. This is passed in from the
+            `dataset_collections_for_cleanup` DAG parameter.
+    """
+    objects_to_delete = []
```
linglp, line 490: can we also add more comments to this function?

BryanFauble: Re-read over the docstring of this function and let me know what your thoughts are. The code is relatively straight forward

linglp: Thanks for updating the docstring. I understand the most parts, but I am curious why it is important to enforce the last filter? why each object in the S3 bucket must match a dataset id, dataset version, dataset collection, and dataset name? is it for ensuring that we delete S3 objects that were no longer present in the dataset collections? I know that you have mentioned that in the PR description, but could you add a comment in the code or clarify in the docstring? I am thinking referencing the PR description might not be as convenient as looking at the code comment.

linglp: I think I understand now after reading more about the doc string in delete_non_current_croissant_file_in_s3 function. We could keep the code there as it is. Thank you

thomasyu888: @linglp thanks for the thoughtful review. My 2 cents is that module docstring covers this information on line 12 which is the requirement to remove the json from the S3 bucket for two scenarios:

1. Datasets removed from the dataset collection
2. Datasets made private from public (since these croissant files are publicly accessible)

I'll defer to Bryan if he would like to add more information but imo, we also don't need to over document.

PR #84, opened 2025-04-07
[DPE-1275] Include content type and additional fields

BryanFauble:
# **Problem:**

- We have a need to include the type of data that each of the files are in the croissant file
- There are additional fields to include at the top level dataset

# **Solution:**

- Pull the `content_type` for each of the files and include it in the Croissant JSON-LD file with a specific default
- Adding additional fields that NF uses to include in the top level dataset fields
- Updating to grab the MD5s and content types for all files from Snowflake for simpler code

# **Testing:**

- I verified that I could run the DAG in a codespaces environment. Here are example values:

`encodingFormat` and `md5` on the FileObject:

```
        {
            "@type": "FileObject",
            "@id": "syn25388376.1",
            "name": "compound_target_associations_v4.rds",
            "description": "Data file associated with compound_target_associations_v4.rds",
            "contentUrl": "https://www.synapse.org/Synapse:syn25388376.1",
            "encodingFormat": "application/octet-stream",
            "md5": "441fb230a817a40e40fed6f10e2553b1",
            "sha256": "unknown"
        },
        {
            "@type": "FileObject",
            "@id": "syn25388377.1",
            "name": "compound_target_associations_v4.csv",
            "description": "Data file associated with compound_target_associations_v4.csv",
            "contentUrl": "https://www.synapse.org/Synapse:syn25388377.1",
            "encodingFormat": "text/csv",
            "md5": "fa2118fd9209fb979720a1a5a4827a5a",
            "sha256": "unknown"
        },
```

Top level dataset keys:

```
    "dataType": [
        "drugScreen"
    ],
    "diseaseFocus": [
        "Neurofibromatosis type 1"
    ],
    "manifestation": [
        "Plexiform Neurofibroma"
    ],
    "studyId": "syn7217928",
    "title": "Drug screens with 3D Plexiform Neurofibroma Model System",
    "yearProcessed": 2016
```

### DIFF
--- a/dags/synapse-dataset-to-croissant.py
+++ b/dags/synapse-dataset-to-croissant.py
@@ -12,7 +12,7 @@ This DAG interacts with a few different services to accomplish the following:
 - (S3, Authenticated) Delete S3 objects from the S3 bucket which are not in a dataset collection
 
 
-To add more datasets to be proceed by this dag:
+To add more datasets to be processed by this dag:
 Simply add the dataset collections to the `dataset_collections` parameter in the DAG to
 run this process for the given dataset collections.
 
@@ -41,6 +41,7 @@ import json
 import os
 import re
 import tempfile
+import mimetypes
 from datetime import datetime
 from io import BytesIO
 from logging import NOTSET, Logger, getLogger
@@ -69,14 +70,13 @@ from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 from opentelemetry.trace.propagation.tracecontext import \
     TraceContextTextMapPropagator
 from orca.services.synapse import SynapseHook
-from orca.errors import ConfigError
-from orca.services.synapse.client_factory import SynapseClientFactory
 from pandas import DataFrame
 from synapseclient import Entity, Synapse, Table
 from synapseclient.core.exceptions import (SynapseAuthenticationError,
                                            SynapseHTTPError)
 from synapseclient.core.retry import with_retry
 from synapseclient.models import File
+from synapseclient.core.utils import delete_none_keys
 
 dag_params = {
     "snowflake_developer_service_conn": Param("SNOWFLAKE_DEVELOPER_SERVICE_RAW_CONN", type="string"),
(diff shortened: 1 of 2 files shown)


dags/synapse-dataset-to-croissant.py
```diff
@@ -381,37 +356,55 @@ def construct_distribution_section_for_files(files_attached_to_dataset: List[Fil
         SELECT
             vd.id,
             vd.versionNumber,
-            vd.contentMd5
+            vd.contentMd5,
+            vd.fileHandleId,
+            COALESCE(fl.content_type, 'NOT_SET') as CONTENT_TYPE,
         FROM version_data AS vd
+        LEFT JOIN synapse_data_warehouse.synapse.file_latest fl
+            ON fl.id = vd.fileHandleId
```
BryanFauble, line 364: `CONTENT_TYPE` is not available within the `VERSION_HISTORY` column, which is why there is a need here to do a join to the `file_latest` table.

dags/synapse-dataset-to-croissant.py
```diff
@@ -352,15 +327,14 @@ def construct_distribution_section_for_files(files_attached_to_dataset: List[Fil
...
-            files_to_find_md5_in_snowflake[int(file.id.replace(
-                "syn", ""))] = file.version_number
-
-    file_md5s = None
+        modified_file_id = int(file.id.replace("syn", ""))
+        ids_of_files.append(modified_file_id)
+        id_and_version_pairs.append(modified_file_id)
+        id_and_version_pairs.append(file.version_number)
+        files_to_find_md5_in_snowflake[int(file.id.replace(
+            "syn", ""))] = file.version_number
```
BryanFauble, line 335: Moving to just pull this information from snowflake. Background comment: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/83#discussion_r2031799819

thomasyu888 (approved):
🔥 LGTM!

dags/synapse-dataset-to-croissant.py
```diff
@@ -381,37 +356,55 @@ def construct_distribution_section_for_files(files_attached_to_dataset: List[Fil
...
+                # Source: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types
+                file_content_type = "application/octet-stream"
+
         distribution_files.append(
             {
                 "@type": "FileObject",
                 "@id": f"{file.id}.{file.version_number}",
                 "name": f"{file.name}",
-                "description": f"Data file associated with {file.name}",
+                "description": file.description if file.description else f"Data file associated with {file.name}",
```
linglp, line 405: can file.name be None?

dags/synapse-dataset-to-croissant.py
```diff
@@ -676,8 +669,11 @@ def execute_push_to_synapse(push_to_synapse: bool, dataset: Entity, dataset_id:
             _rest_call_replacement, authenticated_syn_client)
         existing_row = authenticated_syn_client.tableQuery(
             query=f"SELECT * FROM {SYNAPSE_TABLE_FOR_CROISSANT_LINKS} WHERE dataset = '{dataset_id}' AND dataset_version = {dataset_version} AND dataset_collection = '{dataset_collection}'", resultsAs="csv")
-        existing_row_df = existing_row.asDataFrame()
-        os.remove(existing_row.filepath)
+
```
linglp, line 672: hmm I am curious why this got turned into a try except block? is it because existing_row can be None?

dags/synapse-dataset-to-croissant.py
BryanFauble: No:

```

SELECT * FROM 
    SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
WHERE
    node_type = 'file'
    AND NAME is null
limit 1

SELECT * FROM 
    SYNAPSE_DATA_WAREHOUSE.SYNAPSE.NODE_LATEST
WHERE
    node_type = 'file'
    AND NAME = ''
limit 1
```


Both return no results from the datawarehouse

dags/synapse-dataset-to-croissant.py
BryanFauble: This is to make sure that if an exception is raised during `existing_row.asDataFrame()` that the file is always removed.

`existing_row` wouldn't be `None`, but, it could error inside the function.

PR #85, opened 2025-04-08
[DPE-1266] Challenge DAG Refactor + Introduce Olfactory Challenge DAG

jaymedina:
# **Problem:**

1. We are in the process of making iterative improvements to our challenge infrastructure so that propping up new challenges is as streamlined of a process as possible. As such, we should address any redundancies in `orca-recipes`.
2. A new DAG needs to be created for the new Olfactory challenge.

# **Solution:**

- [x] **Addresses 1**: All of the challenge DAGs share similar logic. This repeated logic can be abstracted out into a "DAG factory" without adding much complexity to our repository or the challenge infra. This also abides by the DRY practice. 

<details><summary>Click for more details & reasoning</summary>
<p>

New challenges DAGs are introduced by adding challenge configurations into a `challenge_config.yaml` file. This makes new challenge contributions into `orca-recipes` similar to the contributions in `nf-synapse-challenge` (where a challenge contribution is just a matter of adding a new profile to the `nextflow.config` file).

Further, this makes DAG contribution on `orca-recipes` easier by not having to re-write logic that already exists or bloat the repository with python scripts for every challenge DAG. This also makes cleanup easier since we can just delete "decommissioned" profiles for challenges that have already completed.

</p>
</details>

- [x] **Addresses 2**: Following the refactor above ^ a new Olfactory challenge profile is introduced into `challenge_config.yaml`

# **Testing:**

For testing, I wrote a test script to execute `challenge_dag_factory.py` and ensure that the `challenge_config.yaml` is getting picked up and creating DAGs iteratively based on the profiles in the config.

<details><summary>Test Script</summary>

```
from challenge_dag_factory import load_challenge_configs, create_challenge_dag

def main():
    # Load the YAML configuration
    configs = load_challenge_configs()
    
    # Loop over each challenge defined in the YAML
    for challenge_name, config in configs.items():
        # Create the DAG object
        dag = create_challenge_dag(challenge_name, config)
        # Print the DAG ID, which confirms registration
        print(f"Created DAG: {dag.dag_id}")
        print("DAG Params:")
        # Iterate over each parameter and print it on its own row.
        for key, obj in dag.params.items():
            print(f"  {key}: {obj.value}")
        print("-" * 40)
        
if __name__ == '__main__':
    main()
```

</details> 

Here is the state of the `challenge_config.yaml` file at the time that I executed the script above.

<details><summary>challenge_config.yaml for testing</summary>

```
pegs:
  synapse_conn_id: "SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN"
  aws_conn_id: "AWS_TOWER_PROD_S3_CONN"
  revision: "558a3786e1610a0cb802e384ab082f696030809a" # Updated to the latest revision with new params.entry for PEGS
  challenge_profile: "pegs_challenge_test"
  tower_conn_id: "PEGS_CHALLENGE_PROJECT_TOWER_CONN"
  tower_view_id: "syn58942525"
  tower_compute_env_type: "spot"
  bucket_name: "pegs-challenge-project-tower-scratch"
  key: "10days/pegs_challenge"
  dag_config:
    schedule_interval: "*/1 * * * *"
    start_date: "2024-04-09T00:00:00"  # ISO-format date string; will be converted in the DAG factory
    catchup: false
    default_args:
      retries: 2
    tags:
      - "nextflow_tower"

test:
  synapse_conn_id: "SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN"
  aws_conn_id: "AWS_TOWER_PROD_S3_CONN"
  revision: "1a2b3c4d" # Updated to the latest revision with new params.entry for PEGS
  challenge_profile: "test_challenge_profile"
  tower_conn_id: "TEST_CHALLENGE_PROJECT_TOWER_CONN"
  tower_view_id: "syn123"
  tower_compute_env_type: "spot"
  bucket_name: "test-challenge-project-tower-scratch"
  key: "10days/test_challenge"
```
</details> 

And here is the output from running `test_challenge_dag_factory`: (These values match what's in the yaml)

```
@jaymedina ➜ /workspaces/orca-recipes/dags (dpe-1266-olfactory-challenge) $ python test_challenge_dag_factory.py 
Created DAG: pegs_challenge_dag
DAG Params:
  synapse_conn_id: SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN
  aws_conn_id: AWS_TOWER_PROD_S3_CONN
  revision: 558a3786e1610a0cb802e384ab082f696030809a
  challenge_profile: pegs_challenge_test
  tower_conn_id: PEGS_CHALLENGE_PROJECT_TOWER_CONN
  tower_view_id: syn58942525
  tower_compute_env_type: spot
  bucket_name: pegs-challenge-project-tower-scratch
  key: 10days/pegs_challenge
  uuid: 32e1415a-3ab5-41f4-ae65-447554597e73
----------------------------------------
Created DAG: test_challenge_dag
DAG Params:
  synapse_conn_id: SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN
  aws_conn_id: AWS_TOWER_PROD_S3_CONN
  revision: 1a2b3c4d
  challenge_profile: test_challenge_profile
  tower_conn_id: TEST_CHALLENGE_PROJECT_TOWER_CONN
  tower_view_id: syn123
  tower_compute_env_type: spot
  bucket_name: test-challenge-project-tower-scratch
  key: 10days/test_challenge
  uuid: 8f9ced37-e211-474b-b7fe-3e7d85f37da9
----------------------------------------
```


jaymedina:
Hi @BWMac, this is still a work in progress because I don't have the Olfactory challenge params yet, but I'd like for you to take a preliminary look at the refactor here so it's ready for a final review when I update the `challenge_configs.yaml` with the new Olfactory profile.

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,173 @@
...
+            "catchup": False,
+            "default_args": {"retries": 2},
+            "tags": ["nextflow_tower"],
+            "params": dag_params,
+        }
+
+def create_challenge_dag(challenge_name: str, config: dict):
+    # Generate a new uuid if none is provided.
+    if not config.get("uuid"):
+        config["uuid"] = str(uuid.uuid4())
```
BWMac: UUID shouldn't be part of the configuration, it needs to be unique for each time a DAG runs, not unique to the challenge itself. The way this is done in current DAGs works because it generates the UUID at DAG runtime, but this would generate at DAG creation time.

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,173 @@
...
+        return custom_config
+    else:
+        return {
+            "schedule_interval": "*/1 * * * *",
+            "start_date": datetime(2024, 4, 9),
+            "catchup": False,
+            "default_args": {"retries": 2},
+            "tags": ["nextflow_tower"],
+            "params": dag_params,
+        }
```
BWMac: I think this could be simplified and more readable. I'm thinking something like

```
    # Start with default configuration
    dag_config = {
        "schedule_interval": "*/1 * * * *",
        "start_date": datetime(2024, 4, 9),
        "catchup": False,
        "default_args": {"retries": 2},
        "tags": ["nextflow_tower"],
        "params": dag_params,
    }

    # Update with any custom configuration if provided
    if config.get('dag_config'):
        dag_config.update(config['dag_config'])
        
        # Ensure start_date is a datetime object if provided
        if 'start_date' in dag_config and isinstance(dag_config['start_date'], str):
            dag_config['start_date'] = datetime.fromisoformat(dag_config['start_date'])
            
        # Ensure challenge name is in tags
        if 'tags' in dag_config:
            if challenge_name not in dag_config['tags']:
                dag_config['tags'].append(challenge_name)
        else:
            dag_config['tags'] = [challenge_name]

    return dag_config
```

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,173 @@
...
+            csv_file = "submissions.csv"
+            df.to_csv(csv_file, index=False)
+            run_uuid = context["params"]["uuid"]
+            # Create a key path that incorporates the unique run identifier.
+            s3_key = f"{key_root}/{run_uuid}/{csv_file}"
+            # Upload the file.
+            s3_hook.load_file(filename=csv_file, key=s3_key, bucket_name=bucket_name)
+            os.remove(csv_file)
+            # Return S3 URI.
+            return f"s3://{bucket_name}/{s3_key}"
```
BWMac: This has been bugging me for a while, and now seems like a good time to make this improvement. We should probably be doing something like this
```
            with tempfile.NamedTemporaryFile(suffix='.csv', delete=True) as temp_file:
                df.to_csv(temp_file.name, index=False)
                run_uuid = context["params"]["uuid"]
                # Create a key path that incorporates the unique run identifier.
                s3_key = f"{key_root}/{run_uuid}/submissions.csv"
                # Upload the file.
                s3_hook.load_file(filename=temp_file.name, key=s3_key, bucket_name=bucket_name)
                # Return S3 URI.
                return f"s3://{bucket_name}/{s3_key}"
```

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,173 @@
...
+from orca.services.synapse import SynapseHook
+
+# Define the path to your challenge configuration file.
+CONFIG_FILE = os.path.join(os.path.dirname(__file__), "challenge_configs.yaml")
+
+
+def load_challenge_configs():
+    """Load challenge configurations from a YAML file."""
+    with open(CONFIG_FILE, "r") as f:
+        return yaml.safe_load(f)
```
BWMac: This should probably take the config file as an argument, it can default to the constant though.

BWMac (commented):
This is great! Much better than duplicated DAGs. I made some comments as I read through. I'm assuming that all of the DAGs in the configuration are generated when Airflow spins up/when that configuration and/or DAG are updated, is that correct? Also will they be paused or unpaused when they are generated?

Re-request when you are ready for final review!

dags/challenge_dag_factory.py
BryanFauble: When we get this deployed out to the airflow server running in K8s this is going to be an area that could fail. I honestly do not know if the git cloning process downloads the files like this or keeps them in memory.

If this does end up becoming an issue we would change this to access the ["raw" github url](https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/2ea5cba6e0dcadb296e3330b626dfa10b541a934/dags/challenge_configs.yaml) and load the HTTP response into memory.

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,173 @@
...
+            # Create a CSV manifest from submissions.
+            df = pd.DataFrame({"submission_id": submissions})
+            csv_file = "submissions.csv"
+            df.to_csv(csv_file, index=False)
+            run_uuid = context["params"]["uuid"]
+            # Create a key path that incorporates the unique run identifier.
+            s3_key = f"{key_root}/{run_uuid}/{csv_file}"
+            # Upload the file.
+            s3_hook.load_file(filename=csv_file, key=s3_key, bucket_name=bucket_name)
+            os.remove(csv_file)
```
BryanFauble: You should not need to write this to a `csv` that is written to disk. I would prefer that you instead load this into an in-memory buffer and load that through the S3 hook. The intention is to prevent files from being written/left there on errors.

See how I am doing it here:

https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/be11f73dd718f690309be0f4763405dc511e5dda/dags/synapse-dataset-to-croissant.py#L618-L633


https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c70c634363f0ea06dbf5ecb9c405dc261d9cc242/synapseclient/core/upload/upload_utils.py#L50-L75

dags/challenge_dag_factory.py
BryanFauble: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/85/files#r2042932851 allows us to side-step this problem

dags/challenge_configs.yaml
thomasyu888, line 1: It would be great if there was documentation around each of the config parameters.  we can then share the documentation to CNB and allow them to give us a round of feedback

dags/challenge_dag_factory.py
jaymedina: Let me know what you think of how I've done it here (the UUID is generated as part of its own task): https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/85/commits/f098885ce4cf0f18d5b7849c2304ddc795c68990

dags/challenge_dag_factory.py
jaymedina: thanks, yeah we've switched to using text and bytes buffers to keep things in-memory: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/85/commits/673d5d91a7d1836ec99e2bd93517ee4137921142

dags/challenge_dag_factory.py
BWMac: Have you tested to make sure it generates different UUIDs for different runs of the same challenge DAG?

jaymedina: I was going to run a stress test once I have the Olfactory challenge profile, but I can set up some test runs now with a dummy profile.

jaymedina: I can confirm that with this logic the UUIDs are different for each run:

image

dags/challenge_configs.yaml
BWMac: Do you think we should have a Confluence page that includes or links to the configurations for each challenge that we can share each time?

jaymedina:
> I'm assuming that all of the DAGs in the configuration are generated when Airflow spins up/when that configuration and/or DAG are updated

Yes, airflow will re-run `challenge_dag_factory.py` as part of the automatic refresh which will loop through all the challenge DAG profiles in `challenge_configs.yaml` and re-create them. I tested this by modifying the name of my challenge from `test` to `testt` and saw that the DAG was renamed accordingly.

> Also will they be paused or unpaused when they are generated?

All DAGs are paused after creation. This seems to be based off [this configuration](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/be11f73dd718f690309be0f4763405dc511e5dda/docker-compose.yaml#L61) and [here too](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/be11f73dd718f690309be0f4763405dc511e5dda/config/airflow.cfg#L42)

dags/challenge_configs.yaml
jaymedina: @BWMac what do you mean by this?:

> ...that includes or links to the configurations for each challenge that we can share each time?

BWMac: Like just a central confluence page that includes all of the configurations on one page, or a "hub" page that links out to pages for individual challenges (or just individual pages per challenge that aren't specifically linked, I guess). Just wondering what the best way to approach the docs would be.

(edited)

BWMac:
@jaymedina Yeah that makes sense, thanks! I ask because what we wouldn't want is for the existing challenge DAGs (that have been unpaused for active submission eval) to become paused by this DAG that dynamically generates each challenge DAG if they are "recreated"

dags/challenge_configs.yaml
jaymedina: Originally my plan was to keep the documentation in the `README` of this repository. The documentation would resemble the README for `nf-synapse-challenge` for the MODEL_TO_DATA/DATA_TO_MODEL workflows, in the way that we describe the parameters, their purpose, and if they're required or not. What do you think?

BWMac: I guess I was thinking about this as documentation on a per challenge basis (I.e. documenting the parameters and also the values we use for each challenge), but now that I am reading this thread again it seems like this was more about the parameters themselves. I think for that the `README` is fine. Do you think we need to document the values we are passing somewhere other than just in the config file for this DAG?

jaymedina: > Do you think we need to document the values we are passing somewhere other than just in the config file for this DAG?

I can't think of a good enough reason for it since we'll have a history of all previous "challenge DAG profiles" via our commit history. Right now it seems like the `README` update would be enough, but LMK if you have anything particular in mind and we can revisit

BWMac: Not sure what in particular you had in mind @thomasyu888 ?

jaymedina:
> I ask because what we wouldn't want is for the existing challenge DAGs (that have been unpaused for active submission eval) to become paused by this DAG that dynamically generates each challenge DAG if they are "recreated"

Great point - I went ahead and tested this (see my last 2 commits where I introduced 2 DAG profiles and modified the second), and both DAGs remained unpaused while `simple2` DAG profile was updated: 

image

Taking a closer look at the [documentation](https://airflow.apache.org/docs/apache-airflow/2.0.2/_api/airflow/models/dag/index.html?utm_source=chatgpt.com), here is the definition for `is_paused_upon_creation`:

> _Specifies if the dag is paused when created for the first time. If the dag exists already, this flag will be ignored. If this optional parameter is not specified, the global config setting will be used._

So reparsing the `challenge_dag_factory.py` file won't cause the recreated DAGs to be paused if they already exist in the metadata DB (were previously created). Thanks for bringing this up!

dags/challenge_dag_factory.py
jaymedina: > we would change this to access the ["raw" github url](https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/2ea5cba6e0dcadb296e3330b626dfa10b541a934/dags/challenge_configs.yaml) and load the HTTP response into memory

I went ahead and [made this change](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/85/commits/580d61cedd083d3ed87207a377dc1cae108b64fc) preemptively so it's more robust (I have it pointed to this branch until this is ready for merge). thanks for the suggestion!

dags/challenge_configs.yaml
thomasyu888: Thanks both! I was thinking of just documenting the parameters, the values will be captured in this config. The README is a good starting place.

Future down the line, if this was a "service" we wanted to provide, we may want to think about how this config is an API contract but for now, a config is more than enough.

README.md
```diff
@@ -72,4 +72,102 @@ to GHCR. In order to accomplish this task:
...
+
+To add a DAG for your challenge, make a pull request introducing a profile to the `challenge_configs.yaml` file with the parameters listed in [Creating Your Challenge DAG Config](#creating-your-challenge-dag-config).
+See below for a template you can follow.
+
+Modify everything in `<>` and remove the `<>` when complete:
+
+```
+<my-challenge>:
+  synapse_conn_id: "SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN"
+  aws_conn_id: "AWS_TOWER_PROD_S3_CONN"
```
linglp: if we are going to hard code `synapse_conn_id` and `aws_conn_id`, should we remove them out of the config?

README.md
```diff
@@ -72,4 +72,102 @@ to GHCR. In order to accomplish this task:
...
+  revision: "<1a2b3c4d>"
```
linglp, line 96: Can we add an explanation of what each parameter means?

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,209 @@
...
+        dag_config.update(config['dag_config'])
+        
+        # Ensure start_date is a datetime object if provided
+        if 'start_date' in dag_config and isinstance(dag_config['start_date'], str):
+            dag_config['start_date'] = datetime.fromisoformat(dag_config['start_date'])
+            
+        # Ensure challenge name is in tags
+        if 'tags' in dag_config:
+            if challenge_name not in dag_config['tags']:
+                dag_config['tags'].append(challenge_name)
```
linglp, line 72: I am curious why is it important for the challenge name to be in "tag"? is it for filtering purpose?

README.md
jaymedina: That's a good point. What do you think @BWMac? Do you see us having different connection IDs for this in the future or can this be a hard-coded global variable in the `challenge_dag_factory.py` script?

dags/challenge_dag_factory.py
jaymedina: Yep! For filtering purposes in the UI. We expect a single challenge to have multiple DAG configuration profiles. Either because it has multiple rounds, or multiple task types, or both. So filtering for a given challenge in the UI so I can easily pause/unpause between rounds and tasks will make things easier for me.

README.md
BWMac: hmmm I don't feel strongly about it but my gut-feeling is to keep this in the config so that we can make that change in the future if we for some reason needed to use different connections.

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,209 @@
...
+            s3_key = f"{key_root}/{run_uuid}/submissions.csv"
+
+            # Upload the file
+            s3_hook.load_file_obj(file_obj=bytes_buffer,
+                                  key=s3_key,
+                                  bucket_name=bucket_name,
+                                  replace=True)
+
+            # Return S3 URI.
+            return f"s3://{bucket_name}/{s3_key}"
```
jaymedina, line 178: Hi, I'm wrapping up on setting things up on the AWS side, and had a question regarding challenge buckets:

We've been setting up a separate bucket for every challenge, but I wonder if we can get away with having a single `challenge-project-scratch` bucket for all the challenges, since the challenge outputs would be divied up based on the S3 key for the given challenge (example: `10days/olfactory_challenge/...`). What do you think @BWMac @thomasyu888 ?

thomasyu888: I think that is a good idea.

However, I think the different challenge buckets are in relation to the S3 buckets that are created in the different workspaces in seqera platform. I'll defer to Brad for confirmation of this.

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,209 @@
...
+from orca.services.synapse import SynapseHook
+
+# Define the path to your challenge configuration file.
+CONFIG_URL = "https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/main/dags/challenge_configs.yaml"
+
+
+def load_challenge_configs(url=CONFIG_URL):
+    """Load challenge configurations from a raw GitHub URL."""
+
+    response = requests.get(url)
```
BryanFauble: Could a timeout be added here:

https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts

dags/challenge_dag_factory.py
BWMac: Sorry I missed this comment initially!

Yes, unless we want to change the way Seqera Platform is configured we should continue with the 1 regular + 1 scratch bucket per challenge (when each challenge has its own workspace). If we had one consolidated challenges workspace, then we could do this, but as it stands I don't think we should make this change.

From a more long-term viewpoint, I think having all challenges run in one workspace makes sense as long as we can appropriately track the billing for each and there are no access control problems introduced by this.

jaymedina: Sounds good, thanks both! ~I'll create separate buckets for Olfactory~ buckets were already made, thanks :)

BTW what is the regular bucket for? In the PEGS dag I don't see us using the regular bucket, only the `scratch`, and I also see that the `scratch` bucket is being used for the dynamic challenge on `tower` prod.

BWMac: The buckets for workspaces are created as part of the `nextflow-infra` CI process, so they were already created when you did [this](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/pull/385).

image


The regular buckets don't have rules to automatically delete contents after X number of days, so we use them for projects where we stage data and want it to persist. Like for running big processing workflows that might take a while we only want to stage the data once.

jaymedina: Good to know, thanks!

README.md
BWMac: Should this be hard-coded in the example if we are allowing for the possibility of using new connections in the future? Should we add a note saying that this is customizable but these are the available connections now?

README.md
```diff
@@ -72,4 +72,102 @@ to GHCR. In order to accomplish this task:
...
+be communicating with.
+
+See below for a list of parameters and their descriptions:
+
+#### Parameters
+
+1. `synapse_conn_id`: The Airflow connection ID used to connect to the Synapse service. This connection is utilized when fetching submission data and updating statuses. **Use `SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN`**.  
+1. `aws_conn_id`: The connection ID for AWS. This enables the DAG to upload CSV files (manifests) to an S3 bucket via the S3 hook. **Use `AWS_TOWER_PROD_S3_CONN`**.  
+1. `revision`: Specifies the version or Git commit revision of the `main` branch in the `nf-synapse-challenge` repository (which houses your workflow). This ensures that the correct version of your workflow is deployed when the DAG triggers a run. **Use `main` or a specific commit SHA which points to the desired version**.  
+1. `challenge_profile`: Identifies the Nextflow Tower challenge profile you contributed in `nextflow.config` of the `nf-synapse-challenge` repository. This parameter customizes the execution environment for the workflow.
```
BWMac: Should there be an example of a profile or a link to an example PR where a profile was added to the Nextflow workflow?

README.md
```diff
@@ -72,4 +72,102 @@ to GHCR. In order to accomplish this task:
...
+
+#### Parameters
+
+1. `synapse_conn_id`: The Airflow connection ID used to connect to the Synapse service. This connection is utilized when fetching submission data and updating statuses. **Use `SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN`**.  
+1. `aws_conn_id`: The connection ID for AWS. This enables the DAG to upload CSV files (manifests) to an S3 bucket via the S3 hook. **Use `AWS_TOWER_PROD_S3_CONN`**.  
+1. `revision`: Specifies the version or Git commit revision of the `main` branch in the `nf-synapse-challenge` repository (which houses your workflow). This ensures that the correct version of your workflow is deployed when the DAG triggers a run. **Use `main` or a specific commit SHA which points to the desired version**.  
+1. `challenge_profile`: Identifies the Nextflow Tower challenge profile you contributed in `nextflow.config` of the `nf-synapse-challenge` repository. This parameter customizes the execution environment for the workflow.  
+1. `tower_conn_id`: This is the connection URI to the Seqera tower workspace for your challenge. This is needed so your challenge DAG can execute the workflow runs in your desired Seqera workspace.
```
BWMac: This is not the URI itself, but an ID like the others above, right?

README.md
```diff
@@ -72,4 +72,102 @@ to GHCR. In order to accomplish this task:
...
+1. `tower_view_id`: The identifier used to query the submission view on Synapse. It tells the DAG, to _tell the workflow_, where to look for submissions to fetch and process.  
+1. `tower_compute_env_type`: Indicates the compute environment (for example, `"spot"`) to be used when launching the workflow. **Use `spot` for challenges that will take less computational time to evaluate the submissions. Use `on-demand` otherwise**.  
+1. `bucket_name`: The S3 bucket where the challenge-related files (such as CSV manifests) will be stored.
```
BWMac: The SP workspaces can only access the S3 buckets assigned to them, so I think we should at least include a note here that points that out. Ideally, I think we would ensure that DAGs are configured that way in a more opinionated fashion.

README.md
linglp: Thanks @BWMac and @jaymedina I brought up this issue because schematic initially included a config file with items that aren’t truly needed to be “configurable.” For example, we allow users to set the required column color of google sheet in the config, but in practice, that option is never needed and we removed it eventually. 

I wonder if simplifying the config overall might be better? That way, we focus only on what users actually need to configure. And if new connection settings or requirements arise in the future, we could always reintroduce relevant options into the config.

thomasyu888: Thats a good point @linglp , Since there aren't that many configuration parameters, what we could probably do is guide users to _not_ change them.  

I'd like to go with @BWMac's gut feeling on this due to his experience with the data pipelines on ITCR.  That said, I could be convinced otherwise, as 

1. we may not need another AWS connection
2. If orca-service is provided access across all challenges, then it's not an issue.  (one caveat around this is governance and one service account having access to _everything_ could be a bit risky, but it is what CNB is doing right now)

README.md
jaymedina: True, technically it's _the ID of the connection URI_. Will update

README.md
jaymedina: Updated!

> Ideally, I think we would ensure that DAGs are configured that way in a more opinionated fashion.

Can you elaborate on this?

BWMac: Basically we at least need to tell people that they can only use the buckets connected to the specific workspace for their challenge, or ideally we add logic to actually enforce that the DAG can only be configured that way. 

If someone tries to use a Bucket that their workspace does not have access to, the workflow will fail in SP.

README.md
jaymedina: > That way, we focus only on what users actually need to configure.

I agree with this line of thinking, but I'm leaning more towards what Brad was saying about this potentially being a configurable option. We could be eliminating future tech debt by adding them as configs straight-away. What if we leave them in the config for now, and revisit if we end up with more config parameters and need to de-clutter?

BWMac: Definitely some good points here. I'm still in favor of leaving them in for now at least, but mostly because I'm really just not that concerned about having "too many parameters" to configure. It will still be super easy to add a new challenge DAG because of @jaymedina 's slick DAG generating code and there being other config examples to copy from.

README.md
```diff
@@ -72,4 +72,111 @@ to GHCR. In order to accomplish this task:
...
+aws_conn_id: "AWS_TOWER_PROD_S3_CONN"
+```
+
+----
+
+### Overview
+
+As part of propping up a new Synapse-hosted challenge, the workflow you choose in [nf-synapse-challenge](https://github.com/Sage-Bionetworks-Workflows/nf-synapse-challenge/tree/main) will need to
+be automated to run for challenge submissions on a rolling basis. For this we leverage Airflow DAGs to perform these
+automated executions of your customized challenge workflow.
```
BryanFauble: Should this portion that explains the purpose of the challenge dag be moved to the top of README under `## Contributing a Challenge DAG`? The purpose is that as one is reading through the README they'll get context for what they're about to read first.

dags/challenge_dag_factory.py
```diff
@@ -0,0 +1,209 @@
...
+        else:
+            dag_config['tags'] = [challenge_name]
+
+    return dag_config
+
+def create_challenge_dag(challenge_name: str, config: dict):
+
+    @task
+    def generate_run_uuid():
+        return str(uuid.uuid4())
```
BryanFauble: Does this need to be decorated with `@task`? This looks like it's executing before it's running in the context of a DAG:
```
    @dag(dag_id=dag_id, **dag_config)
    def challenge_dag():
```

BryanFauble (approved):
Changes look good to me! I'm not seeing any major issues here, great work!

dags/challenge_dag_factory.py
jaymedina: This needs to be one of the tasks in the DAG so it can run in every execution (i.e. a new UUID for every run). I moved it into `challenge_dag()` so it gets scheduled properly

README.md
jaymedina: Got it! I just implemented a new task called `verify_bucket_name` to do this check: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/85/commits/2cb8c91b2d1793ea0d0a381e003d70c38fab9640

I ended up using the `workspace` method from the Seqera hook to determine the expected bucket names and validated against that.

----

Here is an example of passing checks (from the Olfactory profile I'm introducing here):

image

----

And here is an example of the task log for a failing check. I'm using the same `tower_conn_id` for Olfactory, but am using `silly-bucket-scratch` as my `bucket_name`.

<details><summary>Invalid Name Test Profile</summary>

```
silly-challenge:
  synapse_conn_id: "SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN"
  aws_conn_id: "AWS_TOWER_PROD_S3_CONN"
  revision: "652436f5f68a90341150df9c928f390b2469cde1"
  challenge_profile: "olfactory25_challenge_validate"
  tower_conn_id: "OLFACTORY_CHALLENGE_PROJECT_TOWER_CONN"
  tower_view_id: "syn66279193"
  tower_compute_env_type: "spot"
  bucket_name: "silly-bucket-scratch"
  key: "10days/olfactory_challenge"
  dag_config:
    schedule_interval: "*/1 * * * *"
    start_date: "2024-04-09T00:00:00"
    catchup: false
    default_args:
      retries: 2
    tags:
      - "nextflow_tower"

```

</details> 

image

----

A design question for you:

Initially I was going to take the we-fix-it-for-you approach for `verify_bucket_name` where the `bucket_name` gets corrected to `scratch` by default when an invalid `bucket_name` is passed, but I thought it would be harsh since there are 2 valid bucket names and it doesn't make sense to assume everyone will go with the `scratch` counterpart.

Because of this I chose to surface the error with the details. LMK if you would take this approach or the other.

BWMac: Excellent! I think raising the error is the best way because as you said it can be one of the two valid buckets and this will be caught during development when an incorrect config is used. For challenges, `scratch` will probably continue to be the norm, but I think allowing for both is the best approach.

BWMac (approved):
Great work! 🚀

jaymedina:
A successful DAG run all the way through for Olfactory:

image

Thanks all for your feedback! I'll be merging this now.

PR #86, opened 2025-04-24
Update CONFIG_URL to main

jaymedina:
# **Problem:**

CONFIG_URL was pointing in the wrong location & I needed to do some cleanup.

# **Solution:**

- [x] Update CONFIG_URL to point to `main`
- [x] Remove test profile

# **Testing:**

I can confirm this is the URL that works

image


BWMac (approved):

thomasyu888 (approved):
🔥 LGTM!

PR #87, opened 2025-05-02
[SYNPY-1587] SQS Polling POC DAG

BWMac:
# **Problem:**

We needed an example DAG that to demonstrate polling an SQS queue which contains messages about entity creation within a parent entity.

# **Solution:**

As part of [SYNPY-1587](https://sagebionetworks.jira.com/browse/SYNPY-1587), I created this DAG as a demonstration of how one could automatically poll an SQS queue for messages about entities that have been created within a Synapse parent entity (Folder or Project). The SQS queue was created using Terraform modules that can be viewed in this `eks-stack` [PR](https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/62). The Synapse Webhook itself was created using this Python [script](https://gist.github.com/BWMac/fc406be9e4a2aee39c659465e9c9f3e7).

I needed to give Airflow access to SQS in the `dnt-dev` AWS account so I created the `airflow-sqs-access` IAM user with the AmazonSQSFullAccess IAM role attached, and an AWS Secrets connection string `airflow/connections/AWS_DNT_DEV_SQS_CONN` in the `dpe-prod` account. The SQS infrastructure would need to be deployed in `dpe-prod` and similar access would need to be granted if we wanted to use this fully in production.

# **Testing:**

Ensured that there was a message in the SQS queue and ran the DAG. I received this email:
image

# **Notes:**

This DAG should remain paused in production because it is just an example and is not automating a required business process.

[SYNPY-1587]: https://sagebionetworks.jira.com/browse/SYNPY-1587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/dags/synapse-webhook-poc-dag.py
+++ b/dags/synapse-webhook-poc-dag.py
@@ -19,6 +19,11 @@ dag_params = {
     "aws_conn_id": Param(
         "AWS_DNT_DEV_SQS_CONN", type="string", description="AWS connection ID to use"
     ),
+    "region_name": Param(
+        "us-east-1",
+        type="string",
+        description="AWS region where the SQS queue is located",
+    ),
     "sqs_queue_url": Param(
         "https://sqs.us-east-1.amazonaws.com/631692904429/dev-synapse-sqs-create-queue",
         type="string",
@@ -73,7 +78,10 @@ def sqs_polling_synapse_notification_dag():
         Poll the SQS queue for messages.
         This task will return only when messages are found or when it times out.
         """
-        sqs_hook = SqsHook(aws_conn_id=context["params"]["aws_conn_id"])
+        sqs_hook = SqsHook(
+            aws_conn_id=context["params"]["aws_conn_id"],
+            region_name=context["params"]["region_name"],
+        )
         queue_url = context["params"]["sqs_queue_url"]
         max_messages = context["params"]["max_messages"]
         wait_time_seconds = context["params"]["wait_time_seconds"]
--- a/dags/synapse-webhook-poc-dag.py
+++ b/dags/synapse-webhook-poc-dag.py
@@ -109,6 +109,22 @@ def sqs_polling_synapse_notification_dag():
             return response["Messages"]
         return []
 
+    @task.branch()
+    def check_for_messages(messages: List[Dict[str, Any]]) -> str:
+        """
+        Branch task that checks if there are messages to process.
+        """
+        if messages:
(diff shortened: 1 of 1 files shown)


dags/synapse-webhook-poc-dag.py
```diff
@@ -0,0 +1,194 @@
...
+            entries = [
+                {"Id": msg["MessageId"], "ReceiptHandle": msg["ReceiptHandle"]}
+                for msg in response["Messages"]
+            ]
+
+            sqs_hook.get_conn().delete_message_batch(
+                QueueUrl=queue_url, Entries=entries
+            )
+
+            return response["Messages"]
```
BryanFauble: I would consider a change to how this works. My concern is that if `process_messages` or `send_synapse_notification` were to fail in some way, then the messages have already been removed from the queue with (I assume) no way to replay them.

It might be better to grab a message -> Process that message/Do some kind of work -> Delete the message if successful

BWMac: Good point!

BWMac: Thoughts? https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/87/commits/1fe33a5409eaeadae60a74bb54a623dbb5e05831

dags/synapse-webhook-poc-dag.py
```diff
@@ -0,0 +1,202 @@
...
+        ]
+
+        sqs_hook.get_conn().delete_message_batch(QueueUrl=queue_url, Entries=entries)
+
+    # Define task dependencies
+    messages = poll_sqs_queue()
+    branch = check_for_messages(messages)
+    processed_msgs = process_messages(messages)
+    send_notification = send_synapse_notification(processed_msgs)
+    delete_messages = delete_sqs_messages(messages, processed_msgs)
```
BryanFauble: Is it correct that you are passing `messages` and `processed_msgs` here? I see that the function only takes in one arg (and context)

BryanFauble (approved):
Just 1 additional item I flagged, otherwise LGTM!

dags/synapse-webhook-poc-dag.py
BWMac: This was left in by mistake - thanks for that catch

PR #88, opened 2025-05-06
[DPE-1267] Update Olfactory DAG

jaymedina:
# **Problem:**

1. The Olfactory task 1 DAG needs to be updated to point to the latest `nf-synapse-challenge`
2. We now have 2 tasks, so we need a second DAG

# **Solution:**

- [x] Update revision in the DAG config profiles
- [x] Add new DAG profile for task 2

# **Testing:**

- [x] Confirm that both DAGs run successfully to the end, even for invalid submissions

Screenshot for task 1, and I can confirm that task 2 also ran successfully:

image

- [x] Confirm that submission annotations are updated successfully

<details><summary>Annotations Screenshot</summary>

image

</details> 

- [x] Confirm that BEFORE and AFTER e-mails are sent successfully

<details><summary>BEFORE emails</summary>

image

</details> 
<details><summary>AFTER emails, confirmed that the e-mails are not mixed up between submissions</summary>

image

</details>


thomasyu888 (approved):
🔥 LGTM!

PR #89, opened 2025-05-08
[DPE-1288] ALS-KP DAG

BWMac:
# **Problem:**

We need an automated solution for retrieving Dataset metadata from the FAIR C-Path data portal, transforming it to our core data model and loading it into a Dataset Collection in Synapse to power the new federated ALS data portal.

# **Solution:**

This PR implements an Airflow DAG that runs at midnight on the first of each month and performs the following steps:

1. Fetch data from the C-Path API using an authentication token stored in Airflow Variables
2. Transform the raw data using a JSONata mapping expression and validate against a JSON Schema
3. Create or update Synapse Datasets for each item in the transformed data
4. Create or update a Dataset Collection containing all the datasets
5. Create or update annotations for each dataset in the collection
6. Create a new snapshot of the collection if any changes were detected

**Note:** The specific details and data model are subject to change if requested by our stakeholders. This PR will not be merged until these details are confirmed.

# **Testing:**

The DAG was run manually to generate the Dataset Collection that can be viewed on Synapse [here](https://www.synapse.org/Synapse:syn66496326/datasets/).


dags/als-kp-dataset-dag.py
```diff
@@ -0,0 +1,338 @@
...
+        "retries": 2,
+    },
+    "tags": ["als-kp"],
+    "params": dag_params,
+}
+
+
+def load_mapping_from_url(url: str) -> str:
+    """Load the JSONata mapping expression from a URL"""
+    response = requests.get(url)
```
BryanFauble: Make sure a timeout is set when executing this call

dags/als-kp-dataset-dag.py
```diff
@@ -0,0 +1,338 @@
...
+    response.raise_for_status()
+    return response.text
+
+
+def load_schema_from_url(url: str) -> Dict[str, Any]:
+    """Load the JSON Schema from a URL"""
+    response = requests.get(url)
```
BryanFauble: And this call

dags/als-kp-dataset-dag.py
```diff
@@ -0,0 +1,338 @@
...
+    "default_args": {
+        "retries": 2,
+    },
+    "tags": ["als-kp"],
+    "params": dag_params,
+}
+
+
+def load_mapping_from_url(url: str) -> str:
+    """Load the JSONata mapping expression from a URL"""
```
BryanFauble: Nit: Arguments and return in the docstring

dags/als-kp-dataset-dag.py
```diff
@@ -0,0 +1,338 @@
...
+                else item["description"]
+            )
+            dataset = Dataset(
+                name=item["title"],
+                description=dataset_description,
+                parent_id=context["params"]["project_id"],
+            ).store(synapse_client=synapse_client)
+            dataset_collection.add_item(dataset)
+
+        dataset_collection = dataset_collection.store(synapse_client=synapse_client)
```
BryanFauble, line 277: It's awesome to see our changes in SYNPY get used!

dags/als-kp-dataset-dag.py
```diff
@@ -0,0 +1,338 @@
...
+        synapse_client = syn_hook.client
+
+        dataset_ids = [item.id for item in dataset_collection.items]
+
+        # Get current data before update
+        current_data = dataset_collection.query(
+            query=f"SELECT * from {dataset_collection.id}",
+            synapse_client=synapse_client,
+        )
+        current_df = pd.DataFrame(current_data)
```
BryanFauble: Iirc the query function is already returning a DataFrame. Is this extra line needed? Could instead on the line above set the type explicitly like:

`current_data: pd.DataFrame = dataset_collection.query(`

BWMac: Ah yeah that's right, thanks

dags/als-kp-dataset-dag.py
```diff
@@ -0,0 +1,338 @@
...
+                        "publisher",
+                        "species",
+                        "sameAs",
+                    ]
+                },
+            }
+        )
+
+        # Update the rows
+        dataset_collection.update_rows(
```
BryanFauble, line 342: I think you'll want to set this to true

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/70e53e692673930a7c3ce7882aec5faa94632bad/synapseclient/models/dataset.py#L228

BryanFauble (approved):
Great work. I flagged a few things, but overall it looks good to me

dags/als-kp-dataset-dag.py
```diff
@@ -0,0 +1,387 @@
...
+from orca.services.synapse import SynapseHook
+
+from airflow.decorators import task, dag
+from airflow.models import Variable, Param
+
+
+dag_params = {
+    "project_id": Param("syn64892175", type="string"),
+    "mapping_url": Param(
+        "https://raw.githubusercontent.com/amp-als/data-model/refs/heads/main/mapping/cpath.jsonata",
```
thomasyu888: 🔥 Is mapping to main a risk here?

Great work here! Thanks!

BWMac: Good point, I can map to a specific commit hash instead

thomasyu888 (approved):
🔥 Awesome work here!

BWMac:
The output of the DAG has been [approved](https://sagebionetworks.jira.com/browse/DPE-1288?focusedCommentId=249427). Merging now and unpausing in production.

PR #90, opened 2025-05-12
[DPE-1288] Move `jsonata` to requirements-airflow.txt

BWMac:
# **Problem:**

I added a new dependency to `requirements-dev.txt` instead of `requirements-airflow.txt` by mistake. As a result the new DAG implemented in #89 is failing to be imported.

image

# **Solution:**

Move the dependency to `requirements-airflow.txt`.

# **Testing:**

N/A

**Edit:** This issue actually happened because we need to publish a new version of the docker image and update `eks-stack` with the new version number, but this change should still be merged since `jsonata-python` is not a development-specific package

### DIFF
--- a/requirements-airflow.txt
+++ b/requirements-airflow.txt
@@ -6,3 +6,4 @@ py-orca[all] == 1.5.2
 pandas <3.0.0
 slack-sdk >=3.27
 pendulum~=3.0.0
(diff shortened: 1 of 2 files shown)


BryanFauble (approved):
Lgtm

PR #91, opened 2025-05-12
[DPE-1312] Adds Release Container Pipeline

BWMac:
# **Problem:**

Our production Airflow deployment depends on a versioned Docker container being built from this repository, but at present we do not have a release workflow to version the repository and publish a container to GHCR.

# **Solution:**

Implement a workflow that publishes a versioned container. The `latest` tag will be applied when a release is minted, `eks-stack` can be updated to pull the latest release and then it will no longer need to be updated for future releases. New releases only need to be made when a change is implemented to the repository that impacts the Dockerfile, or anything that the Dockerfile references such as dependencies in order for our production deployment to pick up the change.

I also adjusted the Dockerfile and dependencies so that we are only installing from `requirements-airflow.txt` for our production image. `requirements-dev.txt` should only be used for local development or for running any of our local metaflow workflows.

# **Testing:**

Manually ran the pipeline by minting a test release to produce this [container](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pkgs/container/orca-recipes/413830922?tag=0.0.0).

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,39 @@
+name: Release
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  ghcr-publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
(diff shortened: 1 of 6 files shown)


BryanFauble (approved):
This is awesome. Thank for you taking on this additional task

PR #93, opened 2025-05-14
Adds Documentation About Codespaces Secret Rotation

BWMac:
# **Problem:**

In order for our codespaces development environment to have access to the AWS Secrets Manager Backend service, we have codespace-specific secrets configured. The keys for the IAM User must be rotated every 90 days, but this was not well documented.

# **Solution:**

I just rotated the keys, which made me think to add the missing secrets documentation and refresh the secrets docs.


jaymedina (approved):
This is helpful!

PR #95, opened 2025-05-15
ALS-KP DAG Adds Missing Client line

BWMac:
# **Problem:**

The `get` line was missing the client being passed.

# **Solution:**

Pass the client instance.

# **Testing:**

N/A


BryanFauble (approved):
lgtm

PR #96, opened 2025-05-15
[DPE-1267] update schedule_interval for challenge DAGs

jaymedina:
# **Problem:**

Sometimes the challenge DAG won't get to the step that changes the submission status to "EVALUATION_IN_PROGRESS" fast enough, before another cycle is triggered, causing duplicate evaluations of the same submission:

image

# **Solution:**

Update the `schedule_interval` so the DAG runs once every 3 mins instead of every minute.

# **Testing:**

N/A

### DIFF
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -50,7 +50,7 @@ def resolve_dag_config(challenge_name: str, dag_params: dict, config: dict) -> d
     
     # Start with default configuration
     dag_config = {
-        "schedule_interval": "*/1 * * * *",
+        "schedule_interval": "*/3 * * * *",
         "start_date": datetime(2024, 4, 9),
         "catchup": False,
         "default_args": {"retries": 2},


thomasyu888 (approved):
🔥 LGTM!  

Hmm . Is this airflow gathering open submissions?

jaymedina:
> 🔥 LGTM!

Thanks for the quick review 🙏 

> Is this airflow gathering open submissions?

Yeah, it seems like Airflow doesn't get to the `update_submission_status` part before the next job gets triggered sometimes. The reason for me choosing 3 mins is because I usually only see 1 duplicate (meaning it took 2 mins for the first job to finally change the submission status to `RECEIVED` so it doesn't trigger a 2nd duplicate), and then I added 1 minute as a buffer just in case, because I don't really think the wait time is a big deal as long as it's under ~3 mins or so.

image

jaymedina:
@BryanFauble do I need to unpause the DAG in production, or do anything, to get this new interval to register?

jaymedina:
NVM, I see it got updated already!

image

PR #97, opened 2025-05-15
[DPE-1267] update interval in profiles

jaymedina:
# **Problem:**

I updated the default interval, but each profile has a dag config which was silently overwriting the default. This dag config still had 1 minute intervals

# **Solution:**

- [x] Update the profiles

# **Testing:**

N/A

### DIFF
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -50,7 +50,7 @@ def resolve_dag_config(challenge_name: str, dag_params: dict, config: dict) -> d
     
     # Start with default configuration
     dag_config = {
-        "schedule_interval": "*/3 * * * *",
+        "schedule_interval": "*/1 * * * *",
         "start_date": datetime(2024, 4, 9),
         "catchup": False,
         "default_args": {"retries": 2},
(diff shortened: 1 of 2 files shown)


dags/challenge_dag_factory.py
```diff
@@ -50,7 +50,7 @@ def resolve_dag_config(challenge_name: str, dag_params: dict, config: dict) -> d
     
     # Start with default configuration
     dag_config = {
-        "schedule_interval": "*/3 * * * *",
+        "schedule_interval": "*/1 * * * *",
         "start_date": datetime(2024, 4, 9),
```
thomasyu888: Should this be 3 to prevent stuff like this from happening?

thomasyu888 (approved):
🔥 LGTM! Just one minor comment

dags/challenge_dag_factory.py
jaymedina: My idea was to go back to tinkering with the resource management for the Airflow pods on Kubernetes with Bryan, but I can also revert this to 3 to be safe.

jaymedina: I made [a ticket for this](https://sagebionetworks.jira.com/browse/DPE-1318). Not sure where to put it. Modify as you like!

PR #98, opened 2025-05-16
[DPE-1267] Minimize risk of duplicate evaluations for the same Challenge submission

jaymedina:
# **Problem:**

@vpchung mentions she is still experiencing duplicate e-mails sometimes, which means the DAGs are not updating the submission statuses fast enough.

# **Solution:**

Following [conversation](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/98#issuecomment-2891749444) with @BWMac, our workaround solution will involve retrieving submission IDs for the last N submissions to filter them out in submission retrievals for following runs.

- [x] Introduce `get_processed_submission_ids` which gets previous submission IDs already processed
- [x] Create a [Jira ticket](https://sagebionetworks.jira.com/browse/DPE-1351) to figure out the root cause of the repeated evaluations

# **Testing:**

### BEFORE

Previous runs of the task 2 DAG for Olfactory resulted in duplicate evaluations for the same submission

image

### AFTER

Here is the log for a task run that does not find any previous submission IDs from the last 5 runs, which is correct, based on the graph. The log for `get_new_submissions` makes sense. No new submission IDs for the past 5 runs (including the current one).

image

After updating my submission back to RECEIVED, I'm still not able to pick it up in new DAG runs because it's been previously collected, which is the behavior we want:

image

After 4 runs with no submission IDs collected, the DAG forgets about the submission and it gets picked up as a brand new one:

image

I'll bump this up to `n_runs = 6` so it actually checks for the last 5 completed/running submissions.

### DIFF
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -7,8 +7,10 @@ import yaml
 import pandas as pd
 
 from airflow.decorators import dag, task
-from airflow.models import Param
+from airflow.models import Param, DagRun
 from airflow.providers.amazon.aws.hooks.s3 import S3Hook
+from airflow.utils.session import create_session
+from airflow.utils.state import State
 
 from orca.services.nextflowtower import NextflowTowerHook
 from orca.services.nextflowtower.models import LaunchInfo
@@ -75,6 +77,55 @@ def resolve_dag_config(challenge_name: str, dag_params: dict, config: dict) -> d
 
     return dag_config
 
+def get_processed_submission_ids(dag_id: str, n_runs: int = 5) -> set:
+    """
+    Queries the last `n_runs` of successful or running DagRuns for `dag_id`,
+    and pulls their XComs for `get_new_submissions`, returning a set of all submission IDs seen.
+
+    Tip:
+        When the same submission ID needs to be evaluated for testing purposes,
+        set `n_runs` to 0.
+
+    Arguments:
+        dag_id: The ID of the DAG to query.
+        n_runs: The number of DagRuns to query.
+
+    Returns:
+        set: A set of all submission IDs seen.
+
+    """
+
+    processed_submissions = set()
+
+    with create_session() as session:
+
+        # Retrieve the last `n_runs` of successful or running DagRuns
+        previous_runs = (
+            session.query(DagRun).filter(
+                DagRun.dag_id == dag_id,
+                DagRun.state.in_([State.SUCCESS, State.RUNNING])
+            ).order_by(DagRun.start_date.desc()).limit(n_runs).all()
+        )
+
+        # Retrieve the task runs for the task of interest
+        task_id = "get_new_submissions"
+        task_runs = [run.get_task_instance(task_id) for run in previous_runs]
+
+        # Now retrieve the submission IDs for each run of `get_new_submissions` task
+        for task_run in task_runs:
+
+            # Query the metadata DB (``xcom_pull``) to retrieve the submission IDs...
+            # By default, pulling the XComs for `get_new_submissions`
+            # will return the submission IDs, since they are the return values
+            prev_subs = task_run.xcom_pull()
+
+            print(f"Previous submission IDs: {prev_subs}")
+
+            if prev_subs:
+                processed_submissions.update(prev_subs)
+
+    return processed_submissions
+
 def create_challenge_dag(challenge_name: str, config: dict):
 
     # Define parameters for the DAG, including new per-challenge settings.
@@ -120,7 +171,14 @@ def create_challenge_dag(challenge_name: str, config: dict):
             submissions = hook.ops.get_submissions_with_status(
                 context["params"]["tower_view_id"], "RECEIVED"
             )
-            return submissions
+
+            dag_id = context["dag_run"].dag_id
+            processed = get_processed_submission_ids(dag_id=dag_id)
+
+            # Filter out previously-retrieved submissions
+            new_submissions = [s for s in submissions if s not in processed]
+
+            return new_submissions
 
         @task.branch()
         def update_submission_statuses(submissions, **context):
@@ -208,8 +266,8 @@ def create_challenge_dag(challenge_name: str, config: dict):
         tower_run_id = launch_workflow(manifest_path, run_uuid)
         monitor = monitor_workflow(tower_run_id)
 
-        # Fail fast if ``bucket_name`` is invalid, after getting submissions
-        submissions >> submissions_updated >> verify >> [stop, manifest_path]
+        # Fail fast if ``bucket_name`` is invalid
+        verify >> submissions >> submissions_updated >> [stop, manifest_path]
         manifest_path >> tower_run_id >> monitor
 
     return challenge_dag()
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -77,7 +77,7 @@ def resolve_dag_config(challenge_name: str, dag_params: dict, config: dict) -> d
 
     return dag_config
 
-def get_processed_submission_ids(dag_id: str, n_runs: int = 5) -> set:
+def get_processed_submission_ids(dag_id: str, n_runs: int = 6) -> set:
     """
     Queries the last `n_runs` of successful or running DagRuns for `dag_id`,
     and pulls their XComs for `get_new_submissions`, returning a set of all submission IDs seen.
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -208,8 +208,8 @@ def create_challenge_dag(challenge_name: str, config: dict):
         tower_run_id = launch_workflow(manifest_path, run_uuid)
         monitor = monitor_workflow(tower_run_id)
 
-        # Fail fast if ``bucket_name`` is invalid
-        verify >> submissions >> submissions_updated >> [stop, manifest_path]
+        # Fail fast if ``bucket_name`` is invalid, after getting submissions
+        submissions >> submissions_updated >> verify >> [stop, manifest_path]
         manifest_path >> tower_run_id >> monitor
 
     return challenge_dag()
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -116,8 +116,8 @@ def get_processed_submission_ids(dag_id: str, n_runs: int = 5) -> set:
 
             # Query the metadata DB (``xcom_pull``) to retrieve the submission IDs...
             # By default, pulling the XComs for `get_new_submissions`
-            # will return the submission IDs, since they are the return values
-            prev_subs = task_run.xcom_pull()
+            # will return the submission ID list, since that is the return value
+            prev_subs = task_run.xcom_pull(task_ids=task_id)
 
             print(f"Previous submission IDs: {prev_subs}")
 


thomasyu888 (approved):
🔥 LGTM! are there any caveats to the get submission step that it gets submissions that are received?

jaymedina:
After taking a closer look at the execution times of the DAG steps, the "update status" part usually completes within the minute that the DAG starts, so I no longer think it's an issue with that step being too far down in the execution order.

I now think it has to do with the webserver pod freezing randomly due to CPU overload (or something else).

See how the pod has started to freakout for me here once we got to the `monitor_workflow` step.

These challenge DAGs are incredible taxing because of the monitoring step that forces the DAG to keep watching until the workflow completes. I think this is what's causing the hangups. (PS: We have 2 of these DAGs running simultaneously for this challenge, so we need a LOT of resources 🚒 )

@BryanFauble, do you think we can go back into `eks-stack` and reconfigure the resources for the Airflow module?

In terms of urgency, I'm hoping I can get something sorted out by this coming Monday EOD, since the challenge is officially open already, and we don't want the pipeline to run duplicate evaluations for the same submission. But it's nothing to panic over. At worst, the participants will be scratching their heads over why they're getting duplicate e-mails.

image

thomasyu888:
> 
> @BryanFauble, do you think we can go back into `eks-stack` and reconfigure the resources for the Airflow module?
> 
> In terms of urgency, I'm hoping I can get something sorted out by this coming Monday EOD, since the challenge is officially open already, and we don't want the pipeline to run duplicate evaluations for the same submission. But it's nothing to panic over. At worst, the participants will be scratching their heads over why they're getting duplicate e-mails.

@BryanFauble is OOO, so maybe @BWMac can help take a quick peak on Monday.  We can certainly tune the resources in the EKS-stack similar to what you did when you modified it before.

We may want to let Verena know to inform participants that they may be getting duplicate emails for awhile as we iron this out.

jaymedina:
> We may want to let Verena know to inform participants that they may be getting duplicate emails for awhile as we iron this out.

Yes, good call. @vpchung would you mind posting an announcement about this in case submissions trickle in over the weekend?

thomasyu888:
@jaymedina 

Another option is to reduce increase the poke interval to 120 to reduce the burden on the server by a little bit.
```
        @task.sensor(poke_interval=60, timeout=604800, mode="reschedule")
```

BWMac:
@jaymedina With the submission statuses being updated first now and a 3 minute interval, how likely is it that this will still be a problem given the resources airflow currently has allocated? Are there any particular conditions that seem to correlate with the duplicate email issue (I.e. many submissions at once)?

jaymedina:
> how likely is it that this will still be a problem given the resources airflow currently has allocated?

Likely, after looking at the webserver. It seems that the step which updates the submission status executes within the minute, so it doesn't seem like its placement in the task order is an issue.

One thought I had was maybe the issue is with Synapse not reflecting that updated status soon enough, but if that were the case, I would have seen that in PEGS, and I don't remember that happening during stress tests for that challenge, nor did any participants mention duplicate e-mails.

> Are there any particular conditions that seem to correlate with the duplicate email issue (I.e. many submissions at once)?

I had a case where only 1 single submission was processed twice, but I can test it again to be sure.

----

One workaround I can do in this PR is:

Instead of re-ordering the DAG steps, I update the Olfactory challenge DAG profiles with `max_active_runs` to 1. This would ensure that a single run completes before picking up new submissions, so we can be sure that the status for a submission is no longer on `RECEIVED`. It would be in `EVALUATION_IN_PROGRESS` at the least. 

The downside of this is:

1. It would take longer for peoples' submissions to get picked up + processed
2. The longer wait time between workflow runs _could_ lead to a pileup of submissions needing evaluation, so rather than our typical case of 1-3 submissions per workflow run we could have more

With this challenge being a fairly simple evaluation of `csv` files, the workflow runs go pretty fast, and I feel we can get away with setting this parameter to 1, even with the caveats.

LMK what you think @BWMac, and if there are any unforeseen consequences to using this parameter.

PS: This would be a workaround, with the long-term solution being: investigate the source of hangups in the challenge DAGs and improve resource allocation for the AIrflow module if necessary

BWMac:
@jaymedina My understanding after looking over the `eks-stack` [PR](https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/61) you submitted previously is that there is no cap on resource allocations set for airflow and that scaling happens automatically at the node level. So I'm not sure that adjusting resources on Airflow would help any further. Additionally, the webserver doesn't do any of the hard work when a DAG is running.

Edit: Another question, are you seeing tasks get queued or take a long time to kick off when the DAG runs? Or are things running on schedule but just with unintended effects?

> One thought I had was maybe the issue is with Synapse not reflecting that updated status soon enough, but if that were the case, I would have seen that in PEGS, and I don't remember that happening during stress tests for that challenge, nor did any participants mention duplicate e-mails.

I still kinda think this is the most likely culprit. Though it is confusing that this wasn't encountered during stress testing of any previous challenge. Is there any testing you can run by executing the code to confirm if this is the case?

For example, how long does it take for statuses to be updated when you run the code locally vs when it is run by the DAG?

jaymedina:
> For example, how long does it take for statuses to be updated when you run the code locally vs when it is run by the DAG?

It takes < 20 seconds to see the status change reflect on Synapse, both with and without the dag. I'll try the same thing again with multiple submissions and see what happens.

jaymedina:
I'm not able to re-create this dupe e-mail error, unfortunately, but I found the workflow runs for Verena's submissions that caused the dupes **last Friday, May 16**:

(start time: 4:32 PM EST)
https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/olfactory-challenge-project/watch/cja3YWCales7H
and
(start time: 4:33 PM EST)
https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/olfactory-challenge-project/watch/2eDu9S8Vd3VDid

PS: the SP timestamps are all converted to your local timezone, mine being EST.

----

If you go on the production webserver, there are 2 back-to-back jobs that kind-of coincide with the start times for these runs:

image

* The first one started at 20:27 UTC (4:27 PM EST), the second one started at 20:30 UTC (4:30 PM EST).
* The first one retrieved the submissions and updated its status by 4:27:16 PM EST, well before the next run started.

The only thing slightly suspicious is that the first run triggered a workflow that started right at 4:30 PM EST, but the status is updated by then so it shouldnt matter when `launch_workflow` runs. 

----

I'll try re-creating this error again but without watching the DAG on k9s

BWMac:
@jaymedina There are errors in the update status task that would explain the delay in the status being updated before the second run

BWMac:
@jaymedina There are errors in the workflow runs that might explain the delay in the status being updated before the second run.

Edit: But more likely is that the update status task isn't done before the next airflow dag run starts
image
So the airflow dag picks up the same submissions again before the status is updated.

BWMac:
This really doesn't _feel_ like an Airflow issue to me. It seems like the example you provided above behaved as you would expect but that the updated statuses must not have been showing in Synapse when DAG run 2 started. I remember we had an issue with the eventual consistency of these challenge submission views when we first started doing this, but I think I remember you adding functionality to `py-orca` to essentially force a refresh of the view when it was queried. Maybe we need to revisit this to make sure it is working?

jaymedina:
So I closed out of k9s and reset the statuses of my submissions to `RECEIVED`, and got duplicates. Example:

image


Workflow runs:
https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/olfactory-challenge-project/watch/5ya4LlRiyfiHsO
https://tower.sagebionetworks.org/orgs/Sage-Bionetworks/workspaces/olfactory-challenge-project/watch/2QWBVg1IqQlQvi

I don't see anything out of the ordinary in the timestamps for the individual runs on airflow. The status supposedly updates within the minute the run starts, i.e. well before the next run.

> I think I remember you adding functionality to py-orca to essentially force a refresh of the view when it was queried. Maybe we need to revisit this to make sure it is working?

I don't remember doing something like this, but I can do some digging and get back to you.

I definitely think it's suspicious that this stuff happens when the webserver is NOT being monitored.

BWMac:
@jaymedina I'm referring to this https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/7e54c1572c5287c036fbb31774f304a11e9080e3/src/orca/services/synapse/ops.py#L65-L73

jaymedina:
Thanks for the quick reference!

So to make sure that the force refresh works, what I can do is the following:

1. Run the same tests I ran for [this PR ](https://github.com/Sage-Bionetworks-Workflows/py-orca/pull/37) which introduced the change.
2. As another test, I'll change the status of the submissions back to RECEIVED and run it against the DAG while monitoring k9s, and do this 2 more times.

If both these tests pass, we can be pretty sure that it's not an issue with us communicating with an outdated submission view.

I will then try test 2 again but without monitoring on the webserver, to see if that triggers a duplicate again. At that point we'll see a pattern which could give us a clue.

LMK what you think of this plan moving forward @BWMac

BWMac:
I wonder if we should go the "insurance policy" route and include some logic in the DAG that checks for the submission IDs that were processed by previous DAG runs (maybe the last 5 to be safe) and filter out any duplicates as part of `get_new_submissions`. You can get previous task instances using something like:
```
from airflow.models import DagRun
from airflow.utils.session import create_session
from airflow.utils.state import State

with create_session() as session:
    previous_runs = (
        session.query(DagRun)
        .filter(
            DagRun.state.in_([State.SUCCESS, State.RUNNING])
        )
        .order_by(DagRun.start_date.desc())
        .limit(5)  # Check last 5 runs
        .all()
    )

tasks = [run.get_task_instance('get_new_submissions') for run in previous_runs]
```

And from there you should be able to access the XCOM from the tasks (i.e. the list of submission IDs processed previously).
https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/taskinstance/index.html

jaymedina:
>  wonder if we should go the "insurance policy" route

That works - I can update this PR with this solution, and create a separate ticket to figure out the root cause, with [this](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/98#issuecomment-2891520829) as a starting point.

Thanks @BWMac!

jaymedina:
This is ready for a review @BWMac!

I've updated the description with screenshots from my testing.

Edit: I've also made a [Jira ticket ](https://sagebionetworks.jira.com/browse/DPE-1351)for follow-up on this

BWMac (approved):
Great work! 🚀

thomasyu888 (approved):
🔥 nice!

BryanFauble (commented):
@jaymedina  This is great, thanks for the collaborative work to get this across the line.

It's unfortunate that the apache airflow webserver is still having issues every now and then. I'm thinking it's either the health checks we have configured could be patched, or it's just a buggy part of the application. Either way, @BWMac 's comment `the webserver doesn't do any of the hard work when a DAG is running.`, is spot on. The other components are responsible for triggering and running the actual DAGs, so I guess unless the webserver cannot resolve itself I don't see a need for us to spend a bunch of time figuring out the webserver issue at this point

jaymedina:
That makes sense, thanks @BryanFauble for reading through this!

PR #104, opened 2025-06-05
[DPE-1355]: updated data model and added source and url column 

linglp:
# **Problem:**
Currently, when we pull data from C-path, C-path returned the URL suffix of the data in a field called "code". See an example below from the "fetch data" step in the DAG: 
```
{'id': 1815, 'code': 'src_als_1002_2025_03_25', 'xxx': 'xxx'}
```
But this information is not exposed in the final dataset collection. 
After the PR [here]([Update model and mappings by anngvu · Pull Request #14 · amp-als/data-model](https://github.com/amp-als/data-model/pull/14)), we want to update the data model and add "source" and "url" as two additional columns in the dataset collection. 

# **Solution:**
* updated the data model url in the DAG and added two additional columns. 

# **Testing:**
The DAG works: 
Screenshot 2025-06-05 at 1 05 35 PM

And I also could see dataset collection gets updated with "source" and "url" columns: https://www.synapse.org/Synapse:syn66496326/datasets/
Screenshot 2025-06-05 at 1 06 50 PM

### DIFF
--- a/dags/als-kp-dataset-dag.py
+++ b/dags/als-kp-dataset-dag.py
@@ -31,11 +31,11 @@ from airflow.models import Variable, Param
 dag_params = {
     "project_id": Param("syn64892175", type="string"),
     "mapping_url": Param(
-        "https://raw.githubusercontent.com/amp-als/data-model/dd0e476c3659c9b98977d567d0ddce01d5057639/mapping/cpath.jsonata",
+        "https://raw.githubusercontent.com/amp-als/data-model/410a429540a81a63846921ee77d6cf8e33ab6407/mapping/cpath.jsonata",
         type="string",
     ),
     "schema_url": Param(
-        "https://raw.githubusercontent.com/amp-als/data-model/dd0e476c3659c9b98977d567d0ddce01d5057639/json-schemas/Dataset.json",
+        "https://raw.githubusercontent.com/amp-als/data-model/410a429540a81a63846921ee77d6cf8e33ab6407/json-schemas/Dataset.json",
         type="string",
     ),
     "cpath_api_url": Param(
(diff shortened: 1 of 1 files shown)


jaymedina (approved):
LGTM, and thank you for providing screenshots from your testing 🔥 

I had a question just to get a better understanding of this. You mentioned the following:

> _Currently, when we pull data from C-path, C-path returned the URL suffix of the data in a field called "code"_

But when I go to the original [mapping_url ](https://raw.githubusercontent.com/amp-als/data-model/dd0e476c3659c9b98977d567d0ddce01d5057639/mapping/cpath.jsonata) I don't see any field/keyword called `code` here. Was that found somewhere else?

linglp:
> LGTM, and thank you for providing screenshots from your testing 🔥
> 
> I had a question just to get a better understanding of this. You mentioned the following:
> 
> > _Currently, when we pull data from C-path, C-path returned the URL suffix of the data in a field called "code"_
> 
> But when I go to the original [mapping_url ](https://raw.githubusercontent.com/amp-als/data-model/dd0e476c3659c9b98977d567d0ddce01d5057639/mapping/cpath.jsonata) I don't see any field/keyword called `code` here. Was that found somewhere else?

Thank you for taking a look @jaymedina ! When I said “when we pull data from C-Path,” I was referring to [the step](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/dags/als-kp-dataset-dag.py#L166) where we fetch data from the C-Path URL specified [here](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/dags/als-kp-dataset-dag.py#L41). The output of this step is a dictionary containing a "code" field: 

Screenshot 2025-06-06 at 11 49 52 AM

PR #108, opened 2025-07-16
[DPE-1398] Send slack messages when duplicates are found and remove automatic snapshots

linglp:
# **Problem (AC):**
Per discussion with the AMP-ALS team, we want to: 
- Have a human-in-the-loop to validate datasets before creating a snapshot
- Stop changing dataset collection schema 
- Send a Slack message to the channel if duplicates are found
- Provide better error handling when the number of stored datasets does not match the number of annotations to update.

# **Solution:**
- Provide a separate step to detect duplicates and send Slack messages
   - The two datasets are considered duplicates if they have the same title (See Jira comment [here](https://sagebionetworks.jira.com/browse/DPE-1398?focusedCommentId=259869))
- Remove automatic snapshots 
- Improve error handling when the number of stored datasets does not match the number of annotations to update.
- Use a separate JSON file to track validation results from a human. 

# **Testing:**
- Ensure that the DAG is working as expected using test dataset collection:  https://www.synapse.org/Synapse:syn68582286/datasets/
- Ensure that the DAG is working using a test slack channel: # test-amp-als-dataset-message

# Flow chart: 

Untitled diagram _ Mermaid Chart-2025-07-21-190747

### DIFF
--- a/dags/als-kp-dataset-dag.py
+++ b/dags/als-kp-dataset-dag.py
@@ -31,7 +31,7 @@ import time
 
 
 dag_params = {
-    "project_id": Param("syn68155335", type="string"),
+    "project_id": Param("syn64892175", type="string"),
     "mapping_url": Param(
         "https://raw.githubusercontent.com/amp-als/data-model/410a429540a81a63846921ee77d6cf8e33ab6407/mapping/cpath.jsonata",
         type="string",
--- a/dags/als-kp-dataset-dag.py
+++ b/dags/als-kp-dataset-dag.py
@@ -226,9 +226,10 @@ def als_kp_dataset_dag():
         1. Retrieves the current C-PATH datasets.
         2. Identifies and flags new datasets that are duplicates based on their 'title'.
 
-        Args:
+        Arguments:
             transformed_items (List[Dict[str, Any]]):
                 A list of transformed and validated items from the previous task.
+            **context: Airflow task context containing DAG parameters
 
         Raises:
             ValueError: If any 'sameAs' or 'title' values are missing or empty after transformation.
@@ -240,13 +241,13 @@ def als_kp_dataset_dag():
         synapse_client = syn_hook.client
 
         # Get current datasets
-        dataset_collection = DatasetCollection(
-            id=context["params"]["collection_id"]
-        ).get(synapse_client=synapse_client)
-        current_data = dataset_collection.query(
-            query=f"SELECT * from {dataset_collection.id} where publisher='Critical Path Institute'",
-            synapse_client=synapse_client,
+        collection_id = context["params"]["collection_id"]
+        query_str = (
+            f"SELECT * FROM {collection_id} WHERE publisher='Critical Path Institute'"
         )
+
+        current_data = synapse_client.tableQuery(query_str).asDataFrame()
+
         current_datasets = list(current_data["sameAs"])
 
         # Flag duplicates from new data
@@ -274,8 +275,19 @@ def als_kp_dataset_dag():
         return duplicates
 
     @task
-    def generate_slack_message(duplicates: List[Dict[str, Any]], **context) -> str:
-        """Generate the message to be posted to the slack channel."""
+    def generate_slack_message(duplicates: List[Dict[str, Any]]) -> str:
+        """
+        Generate a Slack message summarizing duplicated datasets.
+
+        This task formats a message that includes information about datasets flagged as duplicates,
+        which will be posted to a Slack channel for review.
+
+        Arguments:
+            duplicates (List[Dict[str, Any]]): A list of datasets identified as duplicates.
+
+        Returns:
+            str: A formatted Slack message string describing the duplicate datasets.
+        """
         if not duplicates:
             return ""
         message = "There are datasets that need to be reviewed: \n\n"
@@ -285,8 +297,19 @@ def als_kp_dataset_dag():
         return message
 
     @task
-    def post_slack_messages(message: str):
-        """Post the duplicated datasets to the slack channel."""
+    def post_slack_messages(message: str) -> None:
+        """
+        Post a message to the designated Slack channel.
+
+        This task sends a formatted message (e.g., about duplicated datasets)
+        to a Slack channel using a pre-configured webhook or Slack API.
+
+        Args:
+            message (str): The message string to be posted.
+
+        Returns:
+            None
+        """
         if not message:
             return
         client = WebClient(token=Variable.get("SLACK_DPE_TEAM_BOT_TOKEN"))
@@ -313,7 +336,7 @@ def als_kp_dataset_dag():
                 The list of transformed and validated dataset items to process.
             duplicates (List[Dict[str, Any]]):
                 The list of items identified as duplicates, used to skip adding them.
-
+            **context: Airflow task context containing DAG parameters
         Returns:
             str: The ID of the updated dataset collection.
         """
@@ -338,7 +361,7 @@ def als_kp_dataset_dag():
                 parent_id=context["params"]["project_id"],
             )
             dataset.annotations = {"source": "Critical Path Institute"}
-            dataset.store(synapse_client=synapse_client)
+            dataset.store()
             dataset_collection.add_item(dataset)
         dataset_collection.store()
         return dataset_collection.id
@@ -365,14 +388,10 @@ def als_kp_dataset_dag():
         """
         syn_hook = SynapseHook(context["params"]["synapse_conn_id"])
         synapse_client = syn_hook.client
-
-        dataset_collection = DatasetCollection(id=dataset_collection_id).get(
-            synapse_client=synapse_client
-        )
+        dataset_collection = DatasetCollection(id=dataset_collection_id).get()
 
         current_data = dataset_collection.query(
-            query=f"SELECT * from {dataset_collection.id} where source='Critical Path Institute'",
-            synapse_client=synapse_client,
+            query=f"SELECT * from {dataset_collection.id} where source='Critical Path Institute'"
         )
         dataset_ids = list(current_data["id"])
 
--- a/dags/als-kp-dataset-dag.py
+++ b/dags/als-kp-dataset-dag.py
@@ -454,7 +454,7 @@ def als_kp_dataset_dag():
 
         if len(dataset_ids) != len(annotations["title"]):
             raise ValueError(
-                f" There are {len(dataset_ids)} stored in dataset collection, but there are {len(annotations['title'])} datasets to update. "
+                f"There are {len(dataset_ids)} stored in dataset collection, but there are {len(annotations['title'])} datasets to update."
             )
 
         annotation_data = pd.DataFrame({"id": dataset_ids, **annotations})
--- a/dags/als-kp-dataset-dag.py
+++ b/dags/als-kp-dataset-dag.py
@@ -343,7 +343,7 @@ def als_kp_dataset_dag():
             contents = f.read()
             content_json = json.loads(contents)
             datasets_to_ignore = content_json.get("ignore_cpath_identifier", [])
-        print("dataset identifiers to be ignored after human review: " datasets_to_ignore)
+        print("dataset identifiers to be ignored after human review: " + str(datasets_to_ignore))
         return datasets_to_ignore
 
     @task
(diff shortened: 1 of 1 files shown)


dags/als-kp-dataset-dag.py
```diff
@@ -41,10 +43,7 @@
     "cpath_api_url": Param(
         "https://fair.dap.c-path.org/api/collections/als-kp/datasets", type="string"
     ),
-    "collection_name": Param("Dataset collection (Production)", type="string"),
```
linglp, line 44: collection_name and collection_description were used to create a new dataset collection. But we are retrieving the existing dataset collection now, so these parameters are no longer needed.

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +218,116 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
+            ValueError: If any 'sameAs' or 'title' values are missing or empty after transformation.
 
         Returns:
-            str: The Synapse ID of the created and stored dataset collection
+            List[Dict[str, Any]]: A list of duplicate dataset items if there are any.
         """
         syn_hook = SynapseHook(context["params"]["synapse_conn_id"])
         synapse_client = syn_hook.client
 
-        columns = [
```
linglp, line 242: these are removed because stakeholders do not want the pipeline to change dataset collection schema.

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +218,116 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
+        return duplicates
+
+    @task
+    def generate_slack_message(duplicates: List[Dict[str, Any]], **context) -> str:
+        """Generate the message to be posted to the slack channel."""
+        if not duplicates:
+            return ""
+        message = "There are datasets that need to be reviewed: \n\n"
+        for index, item in enumerate(duplicates):
+            printed_message = f"{index+1}. C-path identifier: {item['sameAs']}, title: {item['title']}, creator: {item['creator']}, keywords: {item['keywords']}, subject: {item['subject']}, collection: {item['collection']}, publisher: {item['publisher']}, species: {item['species']} \n\n"
```
linglp: still waiting for response from Ram and Jess to see if this message is okay or if there's anything else that needs to be added.

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +218,116 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
+            message += printed_message
+        return message
+
+    @task
+    def post_slack_messages(message: str) -> bool:
+        """Post the duplicated datasets to the slack channel."""
+        if not message:
+            return
+        client = WebClient(token=Variable.get("SLACK_DPE_TEAM_BOT_TOKEN"))
+        client.chat_postMessage(channel="test-amp-als-dataset-message", text=message)
```
linglp: TODO: update to production channel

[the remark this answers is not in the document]
linglp: When we use the code in production, we will just use one slack channel so I don't think we should make it as "configurable".

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +218,116 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
+                raise ValueError(f"Missing or empty 'title' in item: {item}")
+
+            if cpath_key not in current_datasets:
+                if title not in seen:
+                    seen[title] = item
+                else:
+                    # found duplicates, add both the first one and this one
+                    duplicates.append(item)
+                    duplicates.append(seen[title])
+        print("Found duplicated datasets:", duplicates)
```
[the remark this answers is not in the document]
linglp: I think using print is fine since the message will just go into Airflow log.

dags/als-kp-dataset-dag.py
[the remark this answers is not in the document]
linglp: The type hinting was wrong. I just corrected it.

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +217,116 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
         dataset_collection = DatasetCollection(
-            name=context["params"]["collection_name"],
-            description=context["params"]["collection_description"],
-            parent_id=context["params"]["project_id"],
-            include_default_columns=True,
-            columns=columns,
+            id=context["params"]["collection_id"]
+        ).get(synapse_client=synapse_client)
+        current_data = dataset_collection.query(
+            query=f"SELECT * from {dataset_collection.id} where publisher='Critical Path Institute'",
```
jaymedina: Is this `Critical Path Institute` always going to be the publisher or should this be programmatically fed?

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +217,116 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
+
+        This task:
+        1. Retrieves the existing dataset collection.
+        2. Iterates over each transformed item, and if a dataset is not flagged as a duplicate:
+            - Creates a dataset in Synapse using the item's title and description.
+            - Adds annotation "source" = "Critical Path Institute" to the dataset
+            - Adds the dataset to the collection.
+            - Updates the dataset collection
+
+        Arguments:
```
jaymedina, line 368: nit: can we update the Arguments to explain what **context is passing?

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +217,116 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
+            if not title:
+                raise ValueError(f"Missing or empty 'title' in item: {item}")
+
+            if cpath_key not in current_datasets:
+                if title not in seen:
+                    seen[title] = item
+                else:
+                    # found duplicates, add both the first one and this one
+                    duplicates.append(item)
+                    duplicates.append(seen[title])
```
BryanFauble: > I chose to compare only the title because the other fields — such as creator, keywords, collection, publisher, and species — all have list values, which makes direct comparison more complicated.

This would be great Acceptance Criteria to define in the ticket. "What makes a dataset a duplicate".

If it is the case that these other items like `creator`, `keywords`.... should be included then we could do a compare like this thread is suggesting: https://stackoverflow.com/questions/7828867/how-to-efficiently-compare-two-unordered-lists-not-sets

dags/als-kp-dataset-dag.py
```diff
@@ -271,32 +336,29 @@ def create_datasets(transformed_items: List[Dict[str, Any]], **context) -> str:
                 name=item["title"],
                 description=dataset_description,
                 parent_id=context["params"]["project_id"],
-            ).store(synapse_client=synapse_client)
+            )
+            dataset.annotations = {"source": "Critical Path Institute"}
+            dataset.store(synapse_client=synapse_client)
```
BryanFauble: nit: You shouldn't need to pass in `synapse_client` anywhere. In theory it should all "Just work" without it. See the call to `dataset_collection.store()`, where (I assume it is working) doesn't have it.

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +217,116 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
-            name=context["params"]["collection_name"],
-            description=context["params"]["collection_description"],
-            parent_id=context["params"]["project_id"],
-            include_default_columns=True,
-            columns=columns,
+            id=context["params"]["collection_id"]
+        ).get(synapse_client=synapse_client)
+        current_data = dataset_collection.query(
+            query=f"SELECT * from {dataset_collection.id} where publisher='Critical Path Institute'",
+            synapse_client=synapse_client,
```
BryanFauble: nit: A minor efficiency gain could be made to skip the `.get()` call on the DatasetCollection. Since you have the ID of the collection you could directly use this within the `SELECT * from {context["params"]["collection_id"]}...` call

BryanFauble (approved):
LGTM!

There are a few areas for improvements, but nothing that would block moving forward once the TODO comments are addressed.

dags/als-kp-dataset-dag.py
linglp: Here, we are trying to retrieve all the C-Path datasets, and yes, all the C-Path datasets have publisher = Critical Path Institute

dags/als-kp-dataset-dag.py
linglp: Got it — thanks, Bryan! I agree this should have been captured in the A/C or even the design doc. I got the requirements directly from the stakeholders during a meeting and didn’t realize at the time that it might become an issue. The stakeholders also haven’t responded yet to my questions about what defines a duplicate dataset, so we will see what they say. Thanks for sending the link of a thread.

dags/als-kp-dataset-dag.py
linglp: Sure. I can update that.

dags/als-kp-dataset-dag.py
linglp: Got it. Thanks for pointing that out. I will fix it.

dags/als-kp-dataset-dag.py
linglp: Thanks for the advice! Just fixed that.

linglp:
@BryanFauble @jaymedina Do you mind re-reviewing the PR? I added a separate task to track the datasets that need to be ignored after human review and validation. I also added a flow chart to explain my thought process.

dags/als-kp-dataset-dag.py
```diff
@@ -325,7 +381,7 @@ def create_datasets(
         ).get(synapse_client=synapse_client)
```
BryanFauble: `synapse_client=synapse_client` can be removed for consistency throughout the script

BryanFauble (approved):
Changes look good to me

dags/als-kp-dataset-dag.py
```diff
@@ -219,49 +220,171 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
+        # Datasets are considered as "duplicates" if they have the same titles
+        duplicates = []
+        seen = {}
+
+        for item in transformed_items:
+            cpath_key = item.get("sameAs")
+            title = item.get("title")
+
+            if not cpath_key:
+                raise ValueError(f"Missing or empty 'sameAs' in item: {item}")
```
thomasyu888, line 266: Does this stop the pipeline?

thomasyu888 (approved):
🔥 LGTM!  Just a minor comment, we may not want the DAG to stop running all together if there is a valueError but maybe if title doesn't exist, it should fail?

jaymedina (approved):
LGTM! My only comment would be to have your flowchart documented somewhere, if not already! 🔥

dags/als-kp-dataset-dag.py
```diff
@@ -5,10 +5,11 @@
 
 1. Fetch data from the C-Path API using an authentication token stored in Airflow Variables
 2. Transform the raw data using a JSONata mapping expression and validate against a JSON Schema
-3. Create or update Synapse Datasets for each item in the transformed data
-4. Create or update a Dataset Collection containing all the datasets
-5. Create or update annotations for each dataset in the collection
-6. Create a new snapshot of the collection if any changes were detected
+3. Find duplicates in the new C-Path data and send a message to AMP-ALS slack channel if duplicates are found. 
+4. Retrieve datasets that need to be ignored after human validation from a JSON file
```
thomasyu888: You may want to add the synapse id of the file here, and state that it is the responsibility of the data managers in AMP-ALS to update the JSON file

dags/als-kp-dataset-dag.py
linglp: Yeah it does stop the pipeline. However, the pipeline should likely be stopped if that key is missing. According to the mapping defined [here](https://github.com/amp-als/data-model/blob/main/mapping/cpath.jsonata), this key is expected to always be present because the C-Path ID should always be present. If a dataset doesn't have an ID, we should probably investigate what's going on.

dags/als-kp-dataset-dag.py
linglp: Sure. Let me add that.

PR #112, opened 2025-07-30
[DPE-1422, DPE-1302, DPE-1407] Add initial maf processing pipeline, permanent fix to pandas float conversion

rxu17:
# **Problem:**
As part of iAtlas, we need to process mutation files through [genome nexus](https://github.com/genome-nexus/genome-nexus-annotation-pipeline) and output them in cBioportal format. This means creating the following files:
- data_mutations.txt
- data_mutations_error_report.txt (genome nexus outputted file of failed annotations and their failure reasons)
- meta_mutations.txt

For more background on these files, you can refer to docs on mutation files here: https://docs.cbioportal.org/file-formats/#mutation-data

We also were experiencing the pandas conversion issues of where pandas reading in integer values in a dataframe that has mixed dtypes converts them into floats with `.0`. This will trigger an issue with cbioportal import as it expects integers not floats.

```
ERROR: data_mutations.txt: lines [4, 5, 8, (197 more)]: column 34: Value in column 't_ref_count' is invalid; values encountered: ['45.0', '174.0', '193.0', '(140 more)']
```

# **Solution:**
This PR consists of multiple JIRA tickets combined together that resolves the above problems:

- Add fix for permanent pandas conversion using `pandas`'s `to_csv` 's `float_format` param: [DPE-1422](https://sagebionetworks.jira.com/browse/DPE-1422)
   - This is added for BOTH clinical and mutation files
- Add initial maf processing workflow: [DPE-1302](https://sagebionetworks.jira.com/browse/DPE-1302)
- Add automation for creating the meta_mutations.txt: [DPE-1407](https://sagebionetworks.jira.com/browse/DPE-1407)

Other additional changes:
There was some shifting around of common functions into a [utils.py file](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/112/files#diff-080af1a58f50e2c589911ab0674e8eb27b69d5473cf5259b4c84ae7666af73bf).

# **Testing:**
- Unit tests pass
- Mutation files get generated
- cBioportal validator errors of integers converted to floats doesn't exist

[DPE-1422]: https://sagebionetworks.jira.com/browse/DPE-1422?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


local/iatlas/cbioportal_export/utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Args:
+        dir_path (str): directory path
+    """
+    for entry in os.listdir(dir_path):
+        entry_path = os.path.join(dir_path, entry)
+        if os.path.isdir(entry_path):
+            shutil.rmtree(entry_path)
+        else:
+            os.remove(entry_path)
```
BryanFauble: nit: This logic works fine - But the other way we could go about it is something like:

```
shutil.rmtree(dir_path)
os.makedirs(dir_path, exist_ok=True)
```

local/iatlas/cbioportal_export/utils.py
```diff
@@ -0,0 +1,60 @@
...
+    Returns:
+        str: data frame in text form with .0 replaced
+    """
+    if header:
+        text = df.to_csv(sep="\t", index=False)
+    else:
+        text = df.to_csv(sep="\t", index=False, header=None)
+
+    text_replaced = text.replace(".0\t", "\t")
+    text_replaced = text_replaced.replace(".0\n", "\n")
```
BryanFauble, line 54: The Synapse python client does this:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/24a0aa0171474860a43ad4e32b50605b23dd7f97/synapseclient/table.py#L2375-L2392

To handle for this case, we may be able to do the same thing here?

local/iatlas/cbioportal_export/maf_to_cbioportal.py
```diff
@@ -0,0 +1,347 @@
...
+    generate_meta_files(
+        dataset_name=args.dataset, datahub_tools_path=args.datahub_tools_path
+    )
+    if not args.dry_run:
+        save_to_synapse(
+            dataset_name=args.dataset,
+            datahub_tools_path=args.datahub_tools_path,
+            output_folder_synid=args.output_folder_synid,
+            version_comment=args.version_comment,
+        )
```
BryanFauble, line 348: If this is a dry_run are we reporting back enough information via log messages to allow us to know that the dry_run is successful?

Or - More specifically, why would I do a dry_run, and what do I look at to verify that a dry_run worked?

local/iatlas/cbioportal_export/maf_to_cbioportal.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+    dataset_dir = os.path.join(datahub_tools_path, "add-clinical-header", dataset_name)
+    # see if dataset_folder exists
+    dataset_folder_exists = False
+    for _, directory_names, _ in synapseutils.walk(syn=syn, synId=output_folder_synid):
+        directories = directory_names  # top level directories
+        break
+
+    for dataset_folder in directories:
+        if dataset_name == dataset_folder[0]:
```
BryanFauble, line 196: Should this be case insensitive?

BryanFauble (approved):
I read over the tests and the logic - They look good to me!

local/iatlas/cbioportal_export/utils.py
rxu17: Awesome thanks! I just pulled my from Genie, will try this one out

local/iatlas/cbioportal_export/maf_to_cbioportal.py
rxu17: Great Q. Let me add more details to the README document, but essentially all the datasets get outputted to your local and just not uploaded to Synapse in the end. Then the cbioportal validator gets run on them with logs outputted locally.

The reason you would want to `dry_run` because you may not want to upload your datasets to synapse if they still have errors from the validator.

local/iatlas/cbioportal_export/maf_to_cbioportal.py
rxu17: Hmm the dataset names appear in a specific format in the datasets and that's how it's outputted / dataset specific folders are created and our collaborators already expect it in that format, so decided to be case sensitive here

local/iatlas/cbioportal_export/maf_to_cbioportal.py
```diff
@@ -0,0 +1,352 @@
...
+        help="Version comment for the files on Synapse. Optional. Defaults to None.",
+    )
+
+    args = parser.parse_args()
+    if args.clear_workspace:
+        utils.clear_workspace(dir_path=f"{args.datahub_tools_path}/add-clinical-header")
+
+    maf_df = pd.read_csv(
+        syn.get(args.input_df_synid).path, sep="\t", comment="#", low_memory=False
+    )
```
thomasyu888, line 320: Future improvement consideration: 

```
df_iter = pd.read_csv('big_data.csv', chunksize=5000)  # Process 5000 rows at a time

for chunk in df_iter:
    print(chunk.head())  # Process each chunk separately
```

thomasyu888 (commented):
🔥 LGTM! Thanks for the work and review

local/iatlas/cbioportal_export/maf_to_cbioportal.py
rxu17: Oh so I meant to iteratively add it in (in another PR) but this part of the code shouldn't exist soon because we'll just be pulling in all the individual mafs and concatenating them, then splitting them up for the genome nexus so this part (reading in the entire maf) shouldn't be there anymore (just a placeholder)

rxu17: PR added here: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/113

PR #113, opened 2025-07-31
[DPE-1302] Add function to read and merge mafs

rxu17:
# **Problem:**
We receive our mafs as individual mafs (depending on the dataset we could have 100 maf files) in a Synapse folder. We should add downloading and merging of those individual mafs to our current maf workflow. 

# **Solution:**
We find it is best to just read in all the mafs and concatenate them all together and then run genome nexus on the bigger mafs rather than on each of the individual mafs (saves more time). 

Our `multiprocessing` (even `asynchio`) seems to be limited to certain number of maf files we can run in parallel mainly because the genome nexus call (which includes API calls) requires us to spin up their docker image each time we call a file.
On top of that, there's some overhead time before genome nexus actually begins annotating the files.

# **Testing:**

- Unit tests run
- Mafs are concatenated and ran through genome nexus


thomasyu888 (approved):
🔥 LGTM!

PR #114, opened 2025-08-02
[DPE-1404, DPE-1403] Add new clinical mapping and run maf processing

rxu17:
# **Problem:**
The Riaz maf still had some outstanding errors before it could be free of cBioportal validation errors:

- Inconsistent clinical `SAMPLE_ID`s vs `Tumor_Sample_Barcode` values in maf file

```
Sample ID not defined in clinical file; values encountered: ["XXX", "XXX", "XXX]
```

- Missing `case_lists_sequenced.txt` which is needed for verifying that the samples in maf file exist (are sequenced) in the clinical samples

# **Solution:**
In order to process the Riaz dataset through to be validated and ingested by cBioportal, there were additional things that needed to happen:

- Apply the new clinical paper ids mapping so that the clinical `SAMPLE_ID`s and maf IDs were consistent (otherwise we run into inconsistent sample id errors when cross-validating between the two files)
- Add code to call cBioportal's [case list generation code](https://github.com/cBioPortal/datahub-study-curation-tools/blob/master/generate-case-lists/generate_case_lists.py) because it will take care of the logic between creating the all clinical samples case list and sequenced samples case lists (needed for part of maf file validation) that our custom case list code doesn't handle

These two fixes had to be coupled together and modifications had to be made to the clinical file as its files are used as the source of the cross-validation with the maf file's files

# **Testing:**
- Unit tests run
- cBioportal validator passed with no `ERROR`s for the maf file, metadata maf file and sequenced case lists file

### DIFF
--- a/local/iatlas/cbioportal_export/clinical_to_cbioportal.py
+++ b/local/iatlas/cbioportal_export/clinical_to_cbioportal.py
@@ -14,8 +14,9 @@ import synapseutils
 
 import utils
 
-my_agent = "iatlas-cbioportal/0.0.0"
-syn = synapseclient.Synapse(user_agent=my_agent).login()
+# my_agent = "iatlas-cbioportal/0.0.0"
+# syn = synapseclient.Synapse(user_agent=my_agent).login()
+syn = synapseclient.login()
 
 EXTRA_COLS = ["study_sample_name", "study_patient_name"]
 
@@ -44,7 +45,11 @@ IATLAS_DATASETS = [
     "Anders_JITC_2022",
     "Zappasodi_Nature_2021",
 ]
-    
+
+IATLAS_DATASETS = [
+    "Riaz_Nivolumab_2017",
+]
+
 ONCOTREE_MERGE_COLS = ["TCGA_Study", "AMADEUS_Study", "Dataset"]
 
 CBIOPORTAL_METADATA_COLS = [
@@ -280,7 +285,7 @@ def add_clinical_header(
     sample_df_subset = input_dfs["sample"][
         input_dfs["sample"]["Dataset"] == dataset_name
     ]
-    
+
     # saves the patient and sample files without pandas float
     sample_df_subset.drop(columns=["Dataset"]).to_csv(
         f"{dataset_dir}/data_clinical_sample.txt",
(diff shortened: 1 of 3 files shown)


local/iatlas/cbioportal_export/clinical_to_cbioportal.py
```diff
@@ -576,50 +621,72 @@ def save_to_synapse(
 
 
 def validate_export_files(
```
thomasyu888, line 623: Does cbioportal validator do these kind of checks?

local/iatlas/cbioportal_export/clinical_to_cbioportal.py
```diff
@@ -64,6 +65,46 @@
...
+        limit in cbioportal.
+
+    Args:
+        input_df (pd.DataFrame): input clinical dataset
+
+    Returns:
+        pd.DataFrame: remapped clinical dataset with new paper ids
+    """
+    cli_remapped = input_df.copy()
+    cli_remapped.loc[~cli_remapped["study_sample_name"].isna(), "sample_name"] = (
```
thomasyu888, line 95: Are the rows that don't have remapped values?

local/iatlas/cbioportal_export/utils.py
```diff
@@ -1,9 +1,44 @@
+import logging
 import os
 import shutil
+import sys
 
 import pandas as pd
 
 
+def create_logger(
```
thomasyu888, line 9: Nice use of logger features here

thomasyu888 (approved):
🔥 LGTM! Just some comments for discussion but code looks great.

local/iatlas/cbioportal_export/clinical_to_cbioportal.py
rxu17: Right now there are as we are still filling in the paper identifiers, but eventually will add to our validation to check that we shouldn't have any NAs in the values to be used for the remap

local/iatlas/cbioportal_export/clinical_to_cbioportal.py
rxu17: Unsure regarding the value of "NA" in patient or sample ids but they shouldn't do checks for any of the other ones.
 - A study can be considered PASS in the cbioportal validator with just having clinical files for example so we want to make sure we have all files present. 
- Also want to check that we didn't lose any PATIENT_IDs and/or SAMPLE_IDs in this pipeline.

PR #115, opened 2025-08-06
[DPE-1429] Add LENS_ID mapping

rxu17:
# **Problem:**
We use paper identifiers for our `SAMPLE_ID` in the iatlas clinical data but those identifiers are very short/abbreviated and don't showcase the full dna, rna sequencing specs in the identifier name.

JIRA Ticket: https://sagebionetworks.jira.com/browse/DPE-1429

# **Solution:**
We will need to create a new clinical iatlas attribute `SAMPLE_DISPLAY_NAME` which will hold the `LENS_ID`s from the generic assay file through the sample_study_name (paper identifier) to lens_id map. This attribute will be the display ID in the visuals when the clinical data gets ingested into cbioportal.

# **Testing:**
- Added unit tests
- No errors thrown with cbioportal validator despite NAs in the `SAMPLE_DISPLAY_NAME`


thomasyu888 (approved):
🔥 LGTM!

PR #116, opened 2025-08-06
[DPE-1401] Add aesthetic updates

rxu17:
# **Problem:**
There are some aesthetic updates/cleaning/formatting required for the iAtlas data to display  in cBioportal. Currently the iAtlas data has things like: 
- string values with the format `[value]_[attribute_name]`
- underscores connecting multi-word values

JIRA Ticket: https://sagebionetworks.jira.com/browse/DPE-1401

# **Solution:**

- remove_suffix_from_column_values - this function removes the [attribute_name] suffix from `[value]_[attribute_name]` for all of the values with that
- update_case_of_column_values - this function takes the `Case` column's casing for each iAtlas attribute in the [clinical mapping](https://www.synapse.org/Synapse:syn66276162) and uses it to change the case for each column to title or all CAPs. This needs to be done after the `remove_suffix_from_column_values` function

# **Testing:**

- Unit tests written
- No additional issues in cbioportal validator results
- Clinical data produces expected results


rxu17:
## DO NOT REVIEW. This may not be needed (will turn into something different entirely)

thomasyu888 (approved):
🔥 LGTM!

PR #117, opened 2025-08-13
[DPE-1437] Remove columns with ALL NAs

rxu17:
# **Problem:**
When the iAtlas data gets displayed on cBioportal, columns with ALL NAs don't get filtered out and gets displayed in the drop down UI like this: 
image

This creates a lot of "extra" display options that are kind of sitting in the dropdown UI without any interaction you can do.

JIRA Ticket: https://sagebionetworks.jira.com/browse/DPE-1437

# **Solution:**
We want to subset out the columns on a dataset by dataset basis that have 100% NAs in all rows. 

A new function `get_all_non_na_columns` is created to get the columns with 100% NAs. This is then used during the generate clinical header step so we're excluding columns with all NAs when exporting the clinical files to be uploaded. Validation is added to check that those columns are removed as expected.

# **Testing:**
- Unit tests
- Added validation to make sure no 100% NA columns
- No cbioportal validation errors

### DIFF
--- a/local/iatlas/cbioportal_export/clinical_to_cbioportal.py
+++ b/local/iatlas/cbioportal_export/clinical_to_cbioportal.py
@@ -6,7 +6,7 @@ from pathlib import Path
 import os
 import subprocess
 import sys
-from typing import Dict
+from typing import Dict, List
 
 import pandas as pd
 import synapseclient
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #118, opened 2025-08-20
[DPE-1441] Add validation rule for allele frequency columns

rxu17:
# **Problem:**
Allele frequency (AF) is a required calculation needed for cbioportal import. If values for the mutation attributes used in the calculation are missing, this value can't be calculated.

JIRA Ticket: https://sagebionetworks.jira.com/browse/DPE-1441

# **Solution:**
AF calculation is as follows:
`AF =  t_alt_count / (t_alt_count + t_ref_count)`

We need to add a validation rule that checks that there are no NAs in either of these two columns: `t_alt_count`, `t_ref_count` in the annotated maf file

We also need to add a validation rule that checks for required maf columns because that's also not available via cbioportal validation.

This PR also adds tests / error logging for the general maf validation checks that were present before

# **Testing:**
- Unit tests
- Run on Riaz mutation data that has missing AFs and check that the error logging is logged

Ran on this [annotated maf](https://www.synapse.org/Synapse:syn68805793.3)
Logging present:
```
2025-08-20 00:19:18,994 - ERROR - There are NAs in the allele frequency columns: ['t_ref_count', 't_alt_count']
```

### DIFF
--- a/local/iatlas/cbioportal_export/maf_to_cbioportal.py
+++ b/local/iatlas/cbioportal_export/maf_to_cbioportal.py
@@ -14,6 +14,123 @@ import utils
 my_agent = "iatlas-cbioportal/0.0.0"
 syn = synapseclient.Synapse(user_agent=my_agent).login()
 
+REQUIRED_MAF_COLS = [
+    "Hugo_Symbol",
+    "Entrez_Gene_Id",
+    "Center",
+    "NCBI_Build",
+    "Chromosome",
+    "Start_Position",
+    "End_Position",
+    "Strand",
+    "Consequence",
+    "Variant_Classification",
+    "Variant_Type",
+    "Reference_Allele",
+    "Tumor_Seq_Allele1",
+    "Tumor_Seq_Allele2",
+    "dbSNP_RS",
+    "dbSNP_Val_Status",
+    "Tumor_Sample_Barcode",
+    "Matched_Norm_Sample_Barcode",
+    "Match_Norm_Seq_Allele1",
+    "Match_Norm_Seq_Allele2",
+    "Tumor_Validation_Allele1",
+    "Tumor_Validation_Allele2",
+    "Match_Norm_Validation_Allele1",
+    "Match_Norm_Validation_Allele2",
+    "Verification_Status",
+    "Validation_Status",
+    "Mutation_Status",
+    "Sequencing_Phase",
+    "Sequence_Source",
+    "Validation_Method",
+    "Score",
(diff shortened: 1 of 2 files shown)


local/iatlas/cbioportal_export/maf_to_cbioportal.py
```diff
@@ -266,21 +266,56 @@ def generate_meta_files(dataset_name: str, datahub_tools_path: str) -> None:
...
+        as these are required to calculate the allele frequency(AF):
+            AF = t_alt_count / (t_alt_count + t_ref_count)
+
+    Args:
+        input_df (pd.DataFrame): input dataframe with allele freq columns
+    """
+    logger = kwargs.get("logger", logging.getLogger(__name__))
+    # check that allele _freq are present
+    allele_freq_cols = ["t_ref_count", "t_alt_count"]
+    if input_df[allele_freq_cols].isna().any().any():
```
thomasyu888: What happens if those columns don't exist?

thomasyu888 (approved):
🔥 LGTM! Just one question

local/iatlas/cbioportal_export/maf_to_cbioportal.py
rxu17: It would break which I am ok with since those should be required columns (I'll add a test for this). I could also make it a "silent" break and add it to logging as well.

I am not 100% if cbioportal validator checks expected maf columns so let me check this real quick.

rxu17: Nevermind, they do not...

rxu17: I'll be adding additional validation to check that req columns are present

local/iatlas/cbioportal_export/maf_to_cbioportal.py
```diff
@@ -266,21 +266,57 @@ def generate_meta_files(dataset_name: str, datahub_tools_path: str) -> None:
...
 
 
-def validate_export_files(input_df: pd.DataFrame, output_df: pd.DataFrame) -> None:
-    """Validates the export files, checking rows
+def validate_export_files(
+    input_df: pd.DataFrame, output_df: pd.DataFrame, **kwargs
+) -> None:
+    """Validates the export files, doing basic checks
+
+        Validation # 1: Check that rows are equal
```
danlu1: nit: number of rows are equal?

danlu1 (approved):
LGTM

PR #120, opened 2025-09-12
[DPE-1278] Push hotdrop public projects to their own table

philerooski:
# **Problem:**

Ticket: [DPE-1278](https://sagebionetworks.jira.com/browse/DPE-1278)

# **Solution:**

I added a new function modeled after an existing function which filters upon the `is_public` property before pushing to a new Synapse table. ~~I haven't yet created the Synapse table.~~ The Synapse table is [here](https://www.synapse.org/Synapse:syn69855119/tables/).

# **Testing:**

I tested in Codespaces by running the DAG via the Airflow web UI. It successfully updated the table I linked to above.

[DPE-1278]: https://sagebionetworks.jira.com/browse/DPE-1278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dags/synapse-datasets-created-7-days.py
```diff
@@ -50,6 +50,7 @@
 }
 
 SYNAPSE_RESULTS_TABLE = "syn64951484"
+SYNAPSE_PUBLIC_RESULTS_TABLE = "syn64951485" #TODO
```
thomasyu888: Can we see the table?

dags/synapse-datasets-created-7-days.py
```diff
@@ -220,16 +221,52 @@ def push_results_to_synapse_table(
...
+                data.append(
+                    [
+                        row.name,
+                        row.id,
+                        row.project_id,
+                        row.node_type,
+                        row.content_type,
+                        row.created_on,
+                        row.created_by,
+                        row.is_public,
```
thomasyu888: Please update: This does not follow the data model set in the acceptance critera.

dags/synapse-datasets-created-7-days.py
```diff
@@ -220,16 +221,52 @@ def push_results_to_synapse_table(
...
+                        today,
+                    ]
+                )
+        syn_hook = SynapseHook(context["params"]["synapse_conn_id"])
+        syn_hook.client.store(
+            synapseclient.Table(schema=SYNAPSE_PUBLIC_RESULTS_TABLE, values=data)
+        )
```
thomasyu888, line 253: not required for this PR: What version of the python client is being used in Airflow?  We may be able to use the new `synapseclient.models -> Table` if it's newer versions of SYNPY.

thomasyu888 (commented):
🔥 Nicely done! I have some comments and I'll defer to someone else on the team for final review.

dags/synapse-datasets-created-7-days.py
BryanFauble: https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/7e54c1572c5287c036fbb31774f304a11e9080e3/setup.cfg#L72

4.8

BryanFauble (approved):
LGTM!

dags/synapse-datasets-created-7-days.py
philerooski: Yes, I just created the table and ran the DAG. You can see the result here: https://www.synapse.org/Synapse:syn69855119/tables/

PR #121, opened 2025-09-15
[DPE-1459] Clarify dev environment set up in README

philerooski:
# **Problem:**

I found instructions on how to set up a dev environment split between README.md and CONTRIBUTING.md. While CONTRIBUTING.md makes a less ambiguous entry point (it links out to README.md after a few steps), the README is the first thing we see and contains the bulk of instructions on setting up the dev environment. Dev containers, docker compose, and Github Codespaces were all new to me, so it wasn’t clear from the README that docker compose should be run from inside the dev container. This caused a lot of confusion, since I could run DAGs just fine from the dev container, but the Airflow environment was completely different from the one I ought to have been using.

# **Solution:**

I expanded upon the dev environment setup instructions and consolidated the instructions into the README. Most of my edits were to the README, and the few changes made to CONTRIBUTING were to either remove redundant dev environment setup information or relocate some information that was once in the README that might be useful to developers but wasn't as salient to setting up a dev environment.

# **Testing:**

I tested both local and remote (Codespaces) methods of setting up a dev environment.


CONTRIBUTING.md
```diff
@@ -2,27 +2,35 @@
...
+Please see [Connecting to AWS EKS](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3389325317/Connecting+to+AWS+EKS+Kubernetes+K8s+cluster) on Confluence if you want to interface with the EKS/Kubernetes cluster. Otherwise, for local development you will likely only be interested in using AWS Secrets Manager as a backend for Airflow Secrets.
+
 There is a helper script in this repository for accessing this Airflow server.
 
 ## Development
 
-To develop on this repository, it's recommended that you use the Dev Containers setup online using GitHub Codespaces. While the entry-level machine type (2-core, 4GB RAM) in Codespaces supports basic editing, you should use a bigger machine type (at least 4-core, 8GB RAM) if you plan on running Airflow using Docker Compose.
+See the [README](./README.md) for instructions on how to set up the development environment.
+
+Any edits to your DAG should get picked up by Airflow automatically. If you're not seeing that happen, you can try restarting the containers:
```
jaymedina: I think this auto-refresh is an Airflow configuration that you can find somewhere in the repo, it's every 2 mins or something like that? Also, you can manually refresh with the refresh button in the Airflow webserver

CONTRIBUTING.md
```diff
@@ -2,27 +2,35 @@
...
-3. Under `Codespaces` click the 3 ellipses and `New with options...`
-4. Choose your branch and 4-core
-5. Once created and you are connected follow the [Quick Start guide](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/README.md#quick-start) to start the Airflow server running in the GitHub codespace.
-6. After your've ran the docker comamnds you will find the Forwarded Address under the `PORTS` tab that you may use to connect to the airflow UI.
+If you edit `Dockerfile`, `docker-compose.yaml`, `requirements-*.txt`, or configuration files, or otherwise want to redo the build process, rebuild the containers:
 
-* Note: The instructions above will create a development environment with all necessary dependencies for Airflow development. The environment setup for the Dev Container is defined in `Dockerfile`.
+```console
+docker compose down
+docker compose up --build --detach
```
jaymedina, line 29: I would opt out of detaching if you want to see the terminal outputs for troubleshooting purposes

thomasyu888 (commented):
🔥 Thanks for putting this together.  Ram from SCIDATA is looking to modify a DAG here and this will help immensely. I'll defer to the team for final review.

CONTRIBUTING.md
philerooski: The scheduler/webserver will pick up new DAGs once every 5 minutes and changes to existing DAGs once every 30 seconds. This is configured in `airflow.cfg`. But (if GPT knows what it's talking about) since the webserver is dependent on the scheduler, and the scheduler is regulated by these refresh intervals, manually refreshing the webserver won't do anything if the scheduler hasn't yet added the DAG to the metadata DB. I edited this section to clarify.

CONTRIBUTING.md
philerooski: We can have our cake and eat it too by attaching to the logs when we need to:

```console
# All logs
docker compose logs -f

# Logs for a specific service(s)
docker compose ps --services
docker compose logs -f airflow-webserver airflow-scheduler
```

By not detaching, the logs hijack our terminal window and we can't interface with Airflow via the CLI without shutting everything down. I added these instructions to CONTRIBUTING.md.

CONTRIBUTING.md
```diff
@@ -2,32 +2,50 @@
 
 ## Infrastructure
 
-We have one Airflow server:
+We have both dev and prod Airflow servers, although the dev server is not always running and there may not be feature parity between dev and prod (e.g., not all prod secrets have analogues in dev):
 
-1. `airflow-prod`: Hosted in the `dpe-prod` AWS account. Deployed using CloudFormation. Only has a private IP address (accessible via SSM port forwarding).
+* `airflow-dev`: Hosted in the `dnt-dev` AWS account.
+* `airflow-prod`: Hosted in the `dpe-prod` AWS account. Deployed using CloudFormation. Only has a private IP address (accessible via SSM port forwarding).
```
BryanFauble:
```suggestion
* `airflow-prod`: Hosted in the `dpe-prod` AWS account. Deployed using OpenTofu. Only accessible via [port forwarding](https://kubernetes.io/docs/tasks/access-application-cluster/port-forward-access-application-cluster/#forward-a-local-port-to-a-port-on-the-pod).
```

BryanFauble (approved):
Great changes, thanks for making these

PR #122, opened 2025-09-18
[DPE-1453] Process ANDERS clinical dataset

rxu17:
# **Problem:**
We need to process the ANDER clinical dataset as part of the iAtlas to cbioportal ingestion work. The dataset has the following main things to address:

- Has all numeric values for the sample_ids used to map in the lens id so will need to convert them
- Has DNA only and RNA samples (sequencing runs) that are not part of the analyses so they will need to be excluded

JIRA: https://sagebionetworks.jira.com/browse/DPE-1453

# **Solution:**
- [Line of code to convert the sample_ids in the lens map to string so it can be merged in](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/122/files#diff-0b8ba0d47465e6c6d3ee2bae70493f2afb814ebc9fda034a36d5680739273172R248)
- [Function to exclude the DNA only and RNA samples (sequencing runs) that are not part of the analyses via pattern in the `sample_name` attribute of `-nr-`, `-nd-`, `-ad-`.](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/122/files#diff-0b8ba0d47465e6c6d3ee2bae70493f2afb814ebc9fda034a36d5680739273172R75) This is to prevent manual removal in the raw clinical data.

# **Testing:**
- Unit tests
- Lens mapping is mapped as expected
- Anders dataset has all unique sample/patient ids after removing samples not part of analyses and passes cbioportal validation

### DIFF
--- a/local/iatlas/cbioportal_export/clinical.py
+++ b/local/iatlas/cbioportal_export/clinical.py
@@ -5,6 +5,7 @@ import logging
 from pathlib import Path
 import os
 import subprocess
+import time
 from typing import Dict, List
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


local/iatlas/cbioportal_export/clinical.py
```diff
@@ -807,29 +833,27 @@ def main():
...
     )
-    lens_id_mapping = get_study_sample_name_to_lens_id_mapping(
-        lens_id_mapping_synid=args.lens_id_mapping_synid, logger=main_logger
-    )
-    cli_dfs["sample"] = add_lens_id_as_sample_display_name(
-        input_df=cli_dfs["sample"],
-        lens_id_mapping=lens_id_mapping,
-        logger=main_logger,
-    )
+    if args.lens_id_mapping_synid is not None:
```
rxu17, line 838: Made optional because sometimes we want to process and validate the data before adding in the optional lens_id

local/iatlas/cbioportal_export/clinical.py
```diff
@@ -209,6 +230,8 @@ def add_lens_id_as_sample_display_name(
     input_df: pd.DataFrame, lens_id_mapping: pd.DataFrame, **kwargs
 ) -> pd.DataFrame:
     """Adds in lens_id as clinical attribute SAMPLE_DISPLAY_NAME
+    Note that lens mapping are dataset specific so sample ids can be all numeric values
+    so some conversion to str is needed
```
danlu1: Can you rephrase here?

local/iatlas/cbioportal_export/clinical.py
danlu1: Can you add this as a comment to the code? Do you mean to validate `cli_dfs` ?

danlu1 (approved):
LGTM just a few comments

local/iatlas/cbioportal_export/clinical.py
rxu17: All of the datasets not just cli_dfs

PR #124, opened 2025-09-23
[DPE-1468] Add neoantigen data into clinical sample data

rxu17:
# **Problem:**
Neoantigen summary counts data will need to be incorporated into the clinical sample data (this is mainly for proper visualization in cbioportal) instead of in its own individual file.

JIRA Ticket: https://sagebionetworks.jira.com/browse/DPE-1468

# **Solution:**
- Create a function [merge_in_neoantigen_study_data](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/124/files#diff-0b8ba0d47465e6c6d3ee2bae70493f2afb814ebc9fda034a36d5680739273172R260-R289) to merge in the neoantigen data into the clinical sample data
- Create [cross-validation validation](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/124/files#diff-9ddf07b9d4e7b74e3d1c67a913fdb118c051cbd5ad6287079605920e366e2314R14-R32) to check that the neoantigen study data's Sample IDs match the maf data's Tumor Sample Barcode values. Now that it's part of the clinical file, we will need to provide our own validation for the IDs instead.

Extras:
- Moved around the cbioportal validator to a validate.py script
- Added a central synapse login function: `synapse_login`

# **Testing:**
- Unit tests
- Tested on iAtlas datasets - the dataset passed general validation and cbioportal validation


local/iatlas/cbioportal_export/validate.py
```diff
@@ -0,0 +1,131 @@
...
+
+    Returns:
+        Dict[str, pd.DataFrame]: dictionary where each key-value pair is
+            file_name : file read in as pandas dataframe
+    """
+    dataset_dir = utils.get_local_dataset_output_folder_path(
+        dataset_name, datahub_tools_path
+    )
+    all_files = {}
+    for file in utils.REQUIRED_OUTPUT_FILES:
```
rxu17, line 52: note that we are not pulling in case lists because cbioportal will validate that

local/iatlas/cbioportal_export/clinical.py
```diff
@@ -258,30 +257,65 @@ def add_lens_id_as_sample_display_name(
...
+    neoantigen_data['SAMPLE_ID'] = neoantigen_data['SAMPLE_ID'].astype(str)
+    df_with_neoantigen = input_df.merge(
+        neoantigen_data,
+        how = "outer",
+        on = "SAMPLE_ID"
+    )
+    if len(df_with_neoantigen) > len(input_df):
+        logger.error(
+            "There are more rows in the clinical data after merging in the neoantigen data."
+        )
```
BryanFauble, line 286: Although this is an error, is it ok that it continues to process after this? Curious if this is a "fatal" error that indicates an upstream process (or this code) caused an issue that needs to be resolved before you can continue.

local/iatlas/cbioportal_export/utils.py
```diff
@@ -2,8 +2,52 @@
...
+        debug=debug, silent=silent, user_agent=f"iatlas-cbioportal/0.0.0"
+    )
+    try:
+        syn.login()
+    except Exception:
+        raise ValueError(
+            "Please view https://help.synapse.org/docs/Client-Configuration.1985446156.html"
+            "to configure authentication to the client.  Configure a ~/.synapseConfig"
+            "or set the SYNAPSE_AUTH_TOKEN environmental variable."
+        )
```
BryanFauble:
```suggestion
    try:
        syn.login()
    except Exception as ex:
        raise ValueError(
            "Please view https://help.synapse.org/docs/Client-Configuration.1985446156.html"
            "to configure authentication to the client.  Configure a ~/.synapseConfig"
            "or set the SYNAPSE_AUTH_TOKEN environmental variable."
        ) from ex
```

The way you currently have this it will hide what the real exception/error was, and it might be useful for debugging

local/iatlas/cbioportal_export/validate.py
```diff
@@ -0,0 +1,131 @@
...
+    validated = f"{datahub_tools_path}/add-clinical-header/{dataset_name}/cbioportal_validator_output.txt"
+    with open(f"{validated}", "w") as outfile:
+        subprocess.run(
+            cmd,
+            shell=True,
+            executable="/bin/bash",
+            stdout=outfile,
+            stderr=subprocess.STDOUT,
+        )
+    print(f"cbioportal validator results saved to: {validated}")
```
BryanFauble: Rather than a print I would expect that you would use the logger you got in main

BryanFauble (approved):
I left a few comments, but nothing too serious. Good work!

local/iatlas/cbioportal_export/validate.py
rxu17: Nice catch!

local/iatlas/cbioportal_export/clinical.py
rxu17: Reasoning I didn't make it into a fatal error because it could be from the data provided itself or from upstream processing, but the data can technically continue to be processed and I wanted all of the errors to be logged into a log file. But, definitely want to log that if there is at least one error, the study validation failed, so will add that.

PR #128, opened 2025-10-21
[DPE-1467] Add `datacite` module and test framework

philerooski:
## **Problem:**

This PR implements a number of high-level changes that support [DPE-1467](https://sagebionetworks.jira.com/browse/DPE-1467). It's not a complete solution to DPE-1467, although it implements some substantial changes to devops so I've broken up the changes.

# **Solution:**
Here's a high-level summary:

### `src/datacite/datacite.py`

This contains the business logic of interacting with the DataCite API, as well as a utility function for writing the data to disk. This will eventually be imported by a DAG.

### `src/datacite/__init__.py`

This allows us to import functions from the previous code like:

```
from datacite import fetch_doi
```

### `tests/test_datacite.py`

This contains unit tests for `src/datacite/datacite.py`

### `tests/fixtures/datacite_fixtures.py`

Logic that is used across tests is in here.

### `tests/README.md`

Since we can expect more than one test suite for the packages we expect to live in `src/`, this README contains an overview of how tests are organized, as well as individual drop-down sections for each test suite. Check the rich diff.

### `tests/__init__.py` and `tests/fixtures/__init__.py`

Boilerplate for importing fixtures and tests.

### `.github/workflows/release.yml`

This adds tests to the release workflow.

### `requirements-airflow.txt`

Add dependencies needed for `src/datacite/datacite.py`

### `requirements-dev.txt`

Add dependencies need to run the tests in `tests/`

### `docker-compose.yaml`

Mount and add `src/` to Airflow's PYTHONPATH.

# **Testing:**
To run tests, check `tests/README.md` or simply run:
```
# Run all tests in the tests directory
pytest tests/ -v
```

If you want to test the _actual_ functionality that will be used in the DAG, you can try:
```
# import iterate_all function from src/datacite/datacite.py
synapse_public_doi = fetch_doi(
  prefixes=["10.7303"],
  state="findable"
)
```

[DPE-1467]: https://sagebionetworks.jira.com/browse/DPE-1467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thomasyu888 (commented):
🔥 This is great, my one initial thought is how the production deployment on K8 is going to

1. Install the new dependencies
2. Read the code from src folder

I'll defer to the rest of the team for first pass on the code

.github/workflows/release.yml
```diff
@@ -6,7 +6,27 @@ on:
...
 jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
```
rxu17, line 17: Is just python 3.10 supported by orca-recipes?

.github/workflows/release.yml
```diff
@@ -6,7 +6,27 @@ on:
   workflow_dispatch:
```
rxu17, line 6: Should tests run for when PRs are open and commits are made as well (separate workflow)?

requirements-airflow.txt
rxu17, line 1: Will we plan to have multiple modules under src/ for the future for `orca-recipes`?

src/datacite/datacite.py
rxu17, line 1: nit: Will there ever be a reason to run this individually outside of the Airflow DAG? If so, should we have a README for this module as well containing this info + more in the boilerplate header here? Might be easier to format and distinguish between code, code parts (like variables) and markdown text

src/datacite/datacite.py
```diff
@@ -0,0 +1,311 @@
...
+
+    Args:
+        data: List of data items returned from the current page.
+        page_size: Expected number of items per full page.
+
+    Returns:
+        True if pagination should continue (full page received),
+        False if this is the last page or no data was returned.
+    """
+    return bool(data) and len(data) >= page_size
```
rxu17, line 184: Do we expect more items than current page size if we are just doing one page at a time? Is this just to account for API edge case scenario?

src/datacite/datacite.py
```diff
@@ -0,0 +1,311 @@
...
+                page_size=page_size,
+                page_number=page_number,
+                detail=detail,
+            )
+            data = payload.get("data", [])
+            if not data:
+                break
+            yield from data
+            # Stop when we returned a short page
+            if not _should_continue_pagination(data, page_size):
```
rxu17, line 339: Should there be a limit for `page_size` that it can't be 0?

src/datacite/datacite.py
```diff
@@ -0,0 +1,311 @@
...
+    page_size: int,
+    page_number: int,
+    detail: bool,
+) -> Dict[str, Any]:
+    """Fetch a single page of DOI objects from DataCite API.
+
+    Args:
+        session: Requests session object with any required headers pre-configured.
+        prefixes: List of DataCite DOI prefixes to query (e.g., ["10.7303"]).
+        state: DOI state to filter by ("findable", "registered", or "draft").
```
rxu17, line 246: Are these the available states? Should we check / validate the input that it's one of these? 

Similarly, are there expected inputs formats for `prefixes`, `page_size`, `page_number`

src/datacite/datacite.py
```diff
@@ -0,0 +1,311 @@
...
+        >>> from datacite import fetch_doi_page
+        >>> 
+        >>> with requests.Session() as session:
+        ...     # Fetch only the first page
+        ...     result = fetch_doi_page(
+        ...         session=session,
+        ...         prefixes=["10.7303"],
+        ...         state="findable",
+        ...         page_size=100,
+        ...         page_number=0,
```
jaymedina: how would a user know what DOIs are in what page? i'm wondering what the use-case would be for only pulling a specific page of DOIs with `fetch_doi_page`

src/datacite/datacite.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    backoff = initial_backoff
+    while True:
+        resp = session.get(url, params=params, timeout=timeout)
+        if resp.status_code == 200:
+            return resp
+        if resp.status_code in (429, 500, 502, 503, 504):
+            time.sleep(backoff)
+            backoff = min(backoff * 2, max_backoff)
+            continue
```
jaymedina, line 142: what if `resp.status_code` never becomes 200? do you want to set an upper-limit for communication retries so this while-loop doesn't repeat forever?

src/datacite/datacite.py
jaymedina, line 184: couldn't there also be cases where the last page is `len(data) = page_size`? in which case this should be `len(data) > page_size`

tests/datacite/conftest.py
```diff
@@ -0,0 +1,262 @@
...
+            title=f"Dataset {i}",
+            author=f"Author {i}"
+        )
+        for i in range(1, 11)
+    ]
+
+
+@pytest.fixture
+def mock_api_response_full_page(sample_doi_objects) -> Dict[str, Any]:
+    """Full page API response (page size matches data length).
```
rxu17, line 172: Should we have a test case where the data length for the full page is > the page size queried?

tests/README.md
rxu17, line 1: Nit: Should the https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/69d1fe6bcda0eeeec0b7d6ac77991b5e5c96a855/README.md#repository-structure / other relevant sections be updated as well?

src/datacite/datacite.py
philerooski: My thinking is that we should give users the option to get data from `https://api.datacite.org/dois` in a more flexible way than fetching every DOI with a specific prefix, but also in a more convenient way than constructing the API query from scratch.

That said, `fetch_doi_page` doesn't do a very good job at this (it only allows you to change the index and number of results). If a flexible function with the ability to pass query parameters as keyword arguments seems useful, I can refactor this. Otherwise, let's just make this function private.

src/datacite/datacite.py
philerooski: Yes, it will save us a potential API call (fetching an empty page) if we make that change.

src/datacite/datacite.py
philerooski: I'll add a `max_retries` parameter.

.github/workflows/release.yml
philerooski: We use the `apache/airflow:2.10.5-python3.10` base image in our Dockerfile.

.github/workflows/release.yml
philerooski: Yes, good point.

.github/workflows/release.yml
rxu17: Nit: Should that info be broadcasted in the docs anywhere?  That we only support `python 3.10` even in our modules

requirements-airflow.txt
philerooski: Yes. What this will look like in most situations is we will have one module per DAG. So for every `my_dag` in the `dags/` folder we get a `my_dag_logic` in the `src/` folder. This helps with maintainability and testing by separating the task logic from the DAG logic. But more future-looking, we will have multiple DAGS which, for example, need to authenticate with Snowflake. So rather than implementing that logic separately for each DAG, each DAG can do something like:
```
from snowflake import connector
```

Because there will be a `connector` method in `snowflake.py`, here:
```
src/
└── snowflake/
    ├── __init__.py
    ├── snowflake.py
 ```

File: src/datacite/datacite.py (hunk 7 above)
Comment (BryanFauble) [reply src/datacite/datacite.py:142]: Since https://github.com/Sage-Bionetworks/synapsePythonClient/blob/4af27a4453001a696f8d1be9f6b973d3fbf62962/synapseclient/core/retry.py#L411 is available due to SYNPY being a dependency already - You should reuse this time based retry here. It will retry up to the configurable time limit.

File: src/datacite/datacite.py
Comment (philerooski) [reply src/datacite/datacite.py:1]: While the primary intention is to run this in a DAG, there's no reason you couldn't borrow it for some other purpose. I don't think that the datacite package as it is now is complex enough to justify a separate README -- probably if you want to use this package you will find it easy enough to read the `datacite.py` module header to get an understanding of how to use this package.

While markdown would be easier on the eyes, I think in this case we can invoke the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). We don't want to maintain a README that contains nearly identical information as the module docstring of the sole module in this package.

File: src/datacite/datacite.py [hunk 9]
```diff
@@ -0,0 +1,311 @@
...
+        Successful HTTP response object with status code 200.
+
+    Raises:
+        requests.HTTPError: For non-retryable HTTP errors (e.g., 400, 401, 404).
+    """
+    backoff = initial_backoff
+    while True:
+        resp = session.get(url, params=params, timeout=timeout)
+        if resp.status_code == 200:
+            return resp
```
BryanFauble: Might also need to handle for a few more status code, like a `204 No Content` response.

src/datacite/datacite.py
philerooski: Although, more generally, if a package in `src/` is complicated enough I think we should include a README that provides an overview of how to use this package. For now I will add a simple README so those browsing on Github don't have to click through to the "full" documentation in the module docstring.

src/datacite/datacite.py
```diff
@@ -0,0 +1,311 @@
...
+        >>> print(f"Wrote {count} objects")
+        Wrote 2 objects
+    """
+    count = 0
+    # Stream to gzip without holding all in memory
+    with gzip.open(out_path, "wb") as gz:
+        for obj in objs:
+            gz.write(_serialize_to_ndjson(obj))
+            count += 1
+    return count
```
BryanFauble, line 380: I am concerned about this and it's something that we need to be very careful with.

On the deployed airflow server we will likely need to create a PersistentVolume and PersistentVolume claim and attach it to the worker pod here: https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/214a5b1b7ed8708e29bb7bd416d2e6852b2154af/modules/apache-airflow/templates/values.yaml#L706-L719

That way it can be mounted as a location where the worker can read/write files to. Additionally - We would probably want to ensure that we have a cleanup process run automatically (Maybe a pre-step to it creating a new gzip file) so that old runs are dropped out.

src/datacite/datacite.py
philerooski: Actually, I'm a little confused. We want to continue pagination if there are more results. It might not be the case that there are more results if `len(data) == page_size` (if the last page includes the last DOI), but we would need to parse the full count out of the API response.

Doing it the way it's done now -- where in exceptional circumstances we might make one extra API call, returning an empty page -- is a little more general and slightly less complicated than doing it that way, I think.

philerooski: No, I think this is just another way of saying `not len(data) < page_size`.

requirements-airflow.txt
rxu17: Hmm, does that mean any code that will be run by an airflow dag should be present under this repo `src/` if it's not too complex/? Or rather how do we determine if something should be present here vs in its own repo?

This made me just think of the [snowflake genie elt scripts](https://github.com/Sage-Bionetworks/snowflake/tree/dev/genie), should those be migrated here because they will technically be called by a DAG in the future?

src/datacite/datacite.py
philerooski: Probably this is handled server-side, but I added a validation check and tests so that `page_size` is between 1 and 1000, inclusive.

tests/datacite/conftest.py
rxu17: Resolved by previous related comment

requirements-airflow.txt
rxu17: Maybe this is more of a dpe topics question, will ask it there

src/datacite/datacite.py
philerooski: I thought that `state` would be enforced server-side, but since "findable" is the only DOI state that you can query if you're unauthenticated, it seems like it ignores whatever I provide for `state` and just returns "findable" DOIs.

I'll add a check.

src/datacite/datacite.py
philerooski: The gzipped data is < 1 MB. Do we need a separate volume for it?

Maybe this is a discussion we can punt to the upcoming DAG/infra changes. A function which writes data to a path is not concerning on its own, I would think.

BryanFauble: image

I SSH'd into one of the worker pods to see what was going on there.

We could try it without creating a separate volume, but typically any pods that want to write data should do so via a volume attached separately. We could even create small volumes like 100 MB.

Even if the file is small, we do need to make sure that they get cleaned up otherwise it would eventually fill up wherever it's written.

src/datacite/datacite.py
philerooski: I'll check for `resp.ok`.

.github/workflows/release.yml
philerooski: I think the development best practices around `src/`, `dags/` and `tests/` should be documented in CONTRIBUTING.md. I'll update.

src/datacite/datacite.py
philerooski: @BryanFauble I'm not sure whether `synapseclient` will remain a dependency in perpetuity. It's only a dependency now because we run all of our DAGs in the same environment.

Actually -- I don't understand how synapseclient is installed. I grep'd for it but didn't see anything that indicated it was made available to every workflow.

thomasyu888: A bit tangent but related: consider something like this for the future: https://airflow.apache.org/docs/apache-airflow-providers-cncf-kubernetes/stable/operators.html.

Right now, the python code is being executed on the airflow executors and the dependencies are being installed as part of those. Eventually, what we actually want is to leverage k8operators for airflow to orchestrate these jobs in k8 pods given an image to run the code.

This will be the final piece for us to fully adopt the orchestration design pattern.

BryanFauble: @philerooski It comes in as a transitive dependency here: https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/f901bbf2223b8dc3996c85fdd7ca2432a3f86548/setup.cfg#L72

But yeah, what @thomasyu888 mentions is good to keep in mind. Swapping over to the k8s operator means we can run any docker container for any of the tasks - Which don't need to include SYNPY. Though, with this I'm fine if you use the minimal amount of code to make this retry and backoff happen, however - it's also really easy to get it wrong.

philerooski:
@thomasyu888 

> Install the new dependencies

That happens on this line:
https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/69d1fe6bcda0eeeec0b7d6ac77991b5e5c96a855/Dockerfile#L9

> Read the code from src folder

We mount the folder (similar to how we already mount `dags/` etc.) and add it to PYTHONPATH. Although I'm not sure if k8s uses the `docker_compose.yml` file or uses a different configuration to set up the environment. @BryanFauble would know.

BryanFauble:
> @thomasyu888
> 
> > Install the new dependencies
> 
> That happens on this line:
> 
> https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/69d1fe6bcda0eeeec0b7d6ac77991b5e5c96a855/Dockerfile#L9
> 
> > Read the code from src folder
> 
> We mount the folder (similar to how we already mount `dags/` etc.) and add it to PYTHONPATH. Although I'm not sure if k8s uses the `docker_compose.yml` file or uses a different configuration to set up the environment. @BryanFauble would know.

@philerooski The deployed Airflow server doesn't work in the same way that the docker compose does.

For example the dags aren't mounted themselves, instead there is a container that runs on a schedule to refetch the dags from github:
https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/214a5b1b7ed8708e29bb7bd416d2e6852b2154af/modules/apache-airflow/templates/values.yaml#L2611-L2698

There also isn't a similar airflow.cfg file that we directly edit, however - If you load up argocd and look at the resources being deployed you'll see that the helm chart will have rendered the value out to a secret or configmap (K8s specific resource). The "Levers" that we have are set through the helm chart values.yaml here:
https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/214a5b1b7ed8708e29bb7bd416d2e6852b2154af/modules/apache-airflow/templates/values.yaml#L2508-L2583

The entire flow is a bit complex because of how `helm` obfuscates stuff, but in their repo they are taking the values from that yaml file and pushing it into an `airflow.cfg` file that is then mounted in the pod that the server uses.

src/datacite/datacite.py
jaymedina: > It might not be the case that there are more results if len(data) == page_size (if the last page includes the last DOI), but we would need to parse the full count out of the API response.

I see what you mean, I agree this is the simpler approach

jaymedina (approved):
excellent documentation ✨ lgtm pending resolved threads

src/datacite/datacite.py
philerooski: urllib3 has methods to do the retry and backoff
```
from requests.adapters import HTTPAdapter
from urllib3.util.retry import Retry
import requests

session = requests.Session()

retry_strategy = Retry(
    total=3,
    backoff_factor=1,
    status_forcelist=[429, 500, 502, 503, 504],
    allowed_methods=["HEAD", "GET", "OPTIONS"]
)

adapter = HTTPAdapter(max_retries=retry_strategy)
session.mount("https://", adapter)
session.mount("http://", adapter)

response = session.get("https://example.com")
```

Although it's also an external dependency, albeit more lightweight than the synapseclient – it has no dependencies of its own. Do you think it's worth implementing with urllib3?

BryanFauble: Read over https://tenacity.readthedocs.io/en/latest/

This gives us a more generic retry mechanism not explicity tied to the HTTP library we are using. The tenacity retry means we can use that for underlying libraries as well to wrap those calls when we have something that is flaky or needs a backoff.

Although, I am fine with either solution here.

rxu17 (approved):
LGTM pending edits!

philerooski:
@BryanFauble Thanks for the explanation. I'm trying to update CONTRIBUTING.md so that developers understand how they can modify the Airflow infrastructure and environment aspects of the repo. Is it the case that changes to files like:

* `Dockerfile`
* `docker-compose.yaml`
* `config/airflow.cfg`

are reflected in the k8s deployment? In other words, are these files used as input (either directly or indirectly) for the k8s deployment? Or am I going to have to update two separate config files – one for local/codespaces, one for k8s – to update a single attribute?

src/datacite/datacite.py
philerooski: I made `_fetch_doi_page` a private function for now. If we ever have a use case where users need more flexibility in how they fetch DOIs we can revisit this.

philerooski:
Summary of changes made since October 23:

* Expanded and refactored README.md and CONTRIBUTING.MD. I'm still not 100% happy with these, but Airflow is a complex stack and these changes add a lot of relevant information for developers.
* Improvements in test organization. Each module now has its own directory for tests.
* Tweaks to function names and which functions are exposed publicly.

PR #129, opened 2025-10-22
[GEN-2263, GEN-2264] Add `genie` module with main genie, BPC and SP elt scripts

rxu17:
# **Problem:**
We want to ingest genie releases for three projects into Snowflake. There currently already exists a `GENIE` database. This will make it easier to have continuously updates of the releases in Snowflake while making it much easier to query for information such as patient-sample tracking (external request) across all three projects vs doing it in synapse.

Based on [previous discussion](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/128#discussion_r2452915834), it made sense to import over the [snowflake genie elt scripts](https://github.com/Sage-Bionetworks/snowflake/tree/dev/genie) from the `snowflake` repo given the relatively low complexity and that most of the external dependencies are already covered by the current Airflow environment.

I did consider migrating these genie scripts over to the genie nextflow repo but the issue would be that we already have aws secrets manager setup for Airflow backend of things and don't want to repeat and duplicate that for Seqera Platform. 

These scripts will eventually be called by an Airflow DAG.

Depends on #128 
Interdependent with: https://github.com/Sage-Bionetworks/snowflake/pull/268

# **Solution:**

Each genie project requires different handling so there's one script for each project:
- Genie BPC will need a config update to ingest releases. This is due to the structure of the BPC projects
- Main genie will be the most automated in ingesting new releases. 
- Genie SP will only ingest certain files as the rest as not as relevant

Creating a common `snowflake_utils` module was needed.
- This allows us to use local snowflake connections (when we are running the scripts locally) as well as snowflake's AirflowHook's connection when we are running it in the DAG. We will be using the GENIE_SERVICE user's specific snowflake connection: `SNOWFLAKE_GENIE_SERVICE_RAW_CONN`
- Some other reusable functions that were repeated in the elt scripts were also added here

Extras:
- Needed to make a new Airflow Snowflake connection and Airflow Synapse connection for snowflake

# **Testing:**
- Unit tests pass

## General Checks for `snowflake_utils` connector functions
- Tested the snowflake_utils connector functions  in a simple Airflow DAG and checked that the connection with airflow works
- Tested the snowflake_utils connector functions locally and make sure the functions worked (just run the genie elt scripts)
   - Got expected logging in airflow DAG triggers `[2025-12-03, 04:13:04 UTC] {python.py:240} INFO - Done. Returned value was: ok_cursor`
   - Got expected logging in airflow DAG triggers `[2025-12-03, 04:13:03 UTC] {python.py:240} INFO - Done. Returned value was: ok_connection`
- Test that the genie elt script outputs in `GENIE_DEV` are as expected

## General Checks for other helper `snowflake_utils` functions
   - Check that there are no duplicates when `overwrite` is true when a table is overwritten: (logging is correct with `Overwriting table 'clinical_patient'.`)
 ```sql
    SELECT
      PATIENT_ID,
      COUNT(*) AS dup_count
    FROM GENIE_DEV.PUBLIC_4_0.CLINICAL_PATIENT
    GROUP BY PATIENT_ID
    HAVING COUNT(*) > 1
    ORDER BY dup_count DESC;
```
    Produces 0 results

    - Check that no table is overwritten when `overwrite` is false:, logging example
       -`2025-12-02 20:48:52,818 [INFO] snowflake_utils: Skipping write — 'gene_matrix' exists and overwrite=False`

## Checks that the uploaded table and the file contents are exactly the same
- Ran some checks against some tables in `GENIE_DEV` and no differences were found

Example checks
```python
import pandas as pd
from pandas.testing import assert_frame_equal
import numpy as np

from snowflake_utils import get_connection, logger, write_to_snowflake
import synapseclient

conn = get_connection()
syn = synapseclient.login()

def normalize_df(df: pd.DataFrame, key_cols):
    df2 = df.copy()

    # consistent column order
    df2 = df2.reindex(sorted(df2.columns), axis=1)

    # sort rows deterministically
    df2 = df2.sort_values(key_cols).reset_index(drop=True)

    # normalize NaN/None to a single representation
    df2 = df2.where(pd.notnull(df2), None)

    return df2

def df_equals_snowflake(conn):
    df_local = pd.read_csv(syn.get("syn11932079", followLink = True).path, sep = "\t", comment = "#")
    key_cols = ["PATIENT_ID"]
    with conn.cursor() as cs:
        cs.execute("SELECT * FROM <database>.<schema>.<table>;")
        df_sf = cs.fetch_pandas_all()

        # same columns?
        if set(df_local.columns) != set(df_sf.columns):
            return False, {"reason": "column mismatch",
                        "local_only": set(df_local.columns) - set(df_sf.columns),
                        "sf_only": set(df_sf.columns) - set(df_local.columns)}

        df_l = normalize_df(df_local, key_cols)
        df_s = normalize_df(df_sf, key_cols)

        # align dtypes if needed (optional, but helps)
        df_s = df_s.astype(df_l.dtypes.to_dict(), errors="ignore")

        try:
            assert_frame_equal(df_l, df_s)
            print("No differences!")
        except Exception as e:
            print(e)
            
        equal = df_l.equals(df_s)
        if equal:
            return True, {}

        # show where they differ
        diff = df_l.compare(df_s, keep_equal=False)
        
        cs.close()
        return False, {"diff": diff}
    
    
df_equals_snowflake(conn)
```


requirements-airflow.txt
```diff
@@ -1,7 +1,7 @@
 apache-airflow-providers-amazon ~=9.2.0
 apache-airflow-providers-celery >=3.10.0
 apache-airflow-providers-snowflake ~=6.0.0
-snowflake-connector-python ~=3.13.2
+snowflake-connector-python ~=3.15.0
```
rxu17, line 4: Needed to upgrade `snowflake-connector-python` to >= 3.14 due to this issue: https://community.snowflake.com/s/article/SnowSQL with local

dags/genie-ingestion-dag.py
rxu17, line 1: This is used for testing the snowflake_utils connections functions. Will be deleted upon merge.

src/snowflake_utils/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+
+## Example Usage
+
+### Connecting to snowflake locally in your code
+
+Setup your environment variables either by executing the following commands directly in your terminal:
+
+```
+export SNOWFLAKE_USER=
```
thomasyu888, line 15: Future potential improvement: consider using .dotenv package

src/snowflake_utils/snowflake_utils.py
```diff
@@ -0,0 +1,207 @@
...
+        "SNOWFLAKE_WAREHOUSE",
+        "SNOWFLAKE_ROLE",
+    ]
+    missing = [v for v in required_vars if not os.getenv(v)]
+    if missing:
+        raise EnvironmentError(f"Missing required Snowflake env vars: {missing}")
+
+    logger.info("Connecting to Snowflake locally...")
+
+    conn = snowflake.connector.connect(
```
thomasyu888, line 57: What happens when there are wrong credentials in this line?

tests/genie/test_genie_bpc_elt.py
thomasyu888, line 1: Nice mocking here

thomasyu888 (commented):
🔥 nice work here. I did a once over, and the high level logic is sound. Will defer to Phil or someone else for final review

Will the functionality within the main() functions of need to be duplicated once it gets to the airflow DAG due to how the connection is done?

src/snowflake_utils/README.md
rxu17: I originally wanted to include it but also wanted to avoid having one more pkg to maintain in the requirements-airflow.txt. Agree that .dotenv is easier to use.

src/snowflake_utils/snowflake_utils.py
rxu17: I believe it would just error out by saying it's missing or wrong/nonexistent.

rxu17:
> Will the functionality within the main() functions of need to be duplicated once it gets to the airflow DAG due to how the connection is done?

It could look like this: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/129/files#diff-2be833d941bbc55a7f647f38a4b91f06b8b6dd065149ad7f0483aa992c1b2949R86-R91 x 3 yes since I don't think we can pass the airflow snowflakehook between tasks so each task could be a genie elt script for example with its own airflow connection if that's what you are asking?

.github/workflows/release.yml
```diff
@@ -16,10 +16,14 @@ jobs:
         with:
           python-version: '3.10'
           
-      - name: Install dependencies
+      - name: Install dev dependencies
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements-dev.txt
```
rxu17, line 22: TODO: For some reason the github actions are not picking up on the `pip install -r requirements-airflow.txt` part, will have to re-check

src/genie/README.md
```diff
@@ -0,0 +1,32 @@
+# Genie ELT scripts
```
rxu17, line 1: TODO: Add links to relevant confluence pages and synapse projects (a lot of genie docs are there)

src/genie/README.md
```diff
@@ -0,0 +1,32 @@
...
+## How to Use
+
+You will need to be a developer on the genie projects and have access to all of the expected accesses in order to run anything in this module. See [GENIE - Getting Started](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/2552037385/Genie) for more information.
+
+- Modify the [genie_bpc_releases yaml](src/genie/genie_bpc_releases.yaml) when a new BPC release has been QC'ed and approved for release for a given cohort. Create a PR to push the changes.
+- Modify the [genie_sp_releases yaml](src/genie/genie_sp_releases.yaml) when a new SP release has been QC'ed and approved for release. Create a PR to push the changes.
+
+Main GENIE releases are automatically ingested.
+
+## Example Usage
```
rxu17, line 14: Nice to have here would be a database schema of all of our genie database tables for a release for a project. I don't think we have one up-to date for the synapse ones and these are supposed to mirror what is on synapse.

src/genie/main_genie_elt.py
```diff
@@ -0,0 +1,229 @@
...
+        if conn is None:
+            conn_obj.close()
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Run the Main GENIE ELT pipeline.")
+    parser.add_argument(
+        "--overwrite", action="store_true", help="Overwrite table data if present."
+    )
+    parser.add_argument(
```
rxu17, line 222: TODO: Have a restriction on which database(s) these scripts can create the necessary schemas and tables in depending on discussion in https://github.com/Sage-Bionetworks/snowflake/pull/268

src/genie/genie_sp_elt.py
```diff
@@ -0,0 +1,126 @@
...
+    schema_name = cohort_config["cohort"]
+    table_name = cohort_config["table_name"]
+    file_synids = cohort_config["file_synid"]
+
+    logger.info(f"Creating/using schema: {schema_name}")
+
+    with conn.cursor() as cs:
+        cs.execute(f"USE DATABASE {database};")
+        cs.execute(
+            f"CREATE SCHEMA IF NOT EXISTS {schema_name} WITH MANAGED ACCESS;"
```
rxu17, line 45: Consider saving each sponsored project's clinical file as a table under a single schema called SPONSORED_PROJECTS or SP since each sponsored project doesn't have versions or multiple files that we want to ingest.

CONTRIBUTING.md
```diff
@@ -167,8 +167,31 @@ Airflow secrets (_e.g._ connections and variables) are stored in Secrets Manager
...
 ### Creating a new secret
 
-New secrets must be created in AWS Secrets Manager in the `dpe-prod` account. For connection URIs, the secret name should have the prefix `airflow/connections/` 
+New secrets must be created in AWS Secrets Manager in the `dpe-prod` account.
+
+1. You will need at least **Developer** access to the account.
+1. Make sure your region is set to **us-east-1**.
+1. For connection URIs, the secret name should have the prefix `airflow/connections/`
 (i.e. `airflow/connections/MY_SECRET_CONNECTION_STRING`). Variables should have the prefix `airflow/variables/` (i.e. `airflow/variables/MY_SECRET_VARIABLE`).
+1. There are two ways to [configure the secrets based on the documentation](https://airflow.apache.org/docs/apache-airflow-providers-snowflake/stable/connections/snowflake.html#configuring-the-connection). If serializing with json, this is a way to turn the key to a single line that could fit into the json:
```
rxu17: Some of these are added from: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/80

.github/workflows/release.yml
thomasyu888: Is there a `--constraints` flag?

src/genie/main_genie_elt.py
thomasyu888: Some TODOs can just be extra tickets if they're not mission critical

src/genie/genie_sp_elt.py
thomasyu888: ^ Can be a separate ticket if this ends up being a lot of work

.github/workflows/release.yml
rxu17: There is a [constraints file](https://github.com/search?q=repo%3ASage-Bionetworks-Workflows%2Forca-recipes%20constraints&type=code) but that's in the overall airflow installation. You mean `--constraints` flag in the validate CI/CD? Not that I can find, why would it ignore a whole step?

rxu17: Just after posting that, I just realized, these changes are all supposed to be in the `validate.yml` as well.... that's why it's not running

rxu17:
After a lot of trying things and experimenting, I think I'll have to bite the bullet and ingest the genie releases into one schema for each project otherwise the downstream effects esp. when querying the tables are much. This new pivot feels like the lesser evil. For the purpose of this PR, I'm just going to scope this just for the clinical tables because that's what we need for the rest of the initiative, and clinical tables are relatively straightforward to handle.

I also will avoid ingesting releases for main genie that are too old (e.g: older than 5 years ago) for this PR if there are too many issues getting them with the much newer releases.

Each project will have a schema MAIN, BPC, and SP. Each schema will contain the specific file format tables for that project.

So a `CLINICAL_SAMPLE` table for main genie would look like:
| SAMPLE_ID | PATIENT_ID | RELEASE | RELEASE_TYPE | MAJOR_VERSION | MINOR_VERSION | OTHER_COLUMNS | 
|-----------|------------|---------|--------------|---------------|---------------|---------------|
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_2_CONSORTIUM | CONSORTIUM | 17 | 2 | ... |
| GENIE-SAGE-1-2 | GENIE-SAGE-1 | 17_2_CONSORTIUM | CONSORTIUM | 17 | 2 | ... |
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_0_PUBLIC | PUBLIC | 17 | 0 | ... |
| GENIE-SAGE-1-2 | GENIE-SAGE-1 | 17_0_PUBLIC | PUBLIC | 17 | 0 | ... |
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_3_CONSORTIUM | CONSORTIUM | 17 | 3 | ... |


A `CANCER_PANEL_TEST` table for BPC will look like the above but with an addition `COHORT` column. 

SP project table will just be one table where there won't be any of the RELEASE, RELEASE_TYPE, MAJOR_VERSION, MINOR_VERSION columns but a column containing the SP Project name like `NAME`. OR we can have one table per SP project since their table schemas vary from project to project (but the query will just be elongated as we have 7 SP projects and this number will grow in the future adding to our query length)

```mermaid
erDiagram
  GENIE_MAIN_CLINICAL_SAMPLE {
    string  SAMPLE_ID
    string  PATIENT_ID
    string  RELEASE
    string  RELEASE_TYPE
    int     MAJOR_VERSION
    int     MINOR_VERSION
    string  OTHER_COLUMNS
  }

  GENIE_BPC_CANCER_PANEL_TEST {
    string  COHORT
    string  SAMPLE_ID
    string  PATIENT_ID
    string  RELEASE
    string  RELEASE_TYPE
    int     MAJOR_VERSION
    int     MINOR_VERSION
    string  OTHER_COLUMNS
  }

  GENIE_SP_CLINICAL {
    string  NAME
    string  SAMPLE_ID
    string  PATIENT_ID
    string  OTHER_COLUMNS
  }

  %% Conceptual "grouping" relationships (not actual foreign keys)
  GENIE_MAIN_CLINICAL_SAMPLE }o--|| GENIE_MAIN_CLINICAL_SAMPLE : "RELEASE has many PATIENT_ID/SAMPLE_ID"
  GENIE_BPC_CANCER_PANEL_TEST }o--|| GENIE_BPC_CANCER_PANEL_TEST : "COHORT has many RELEASEs; each RELEASE has many PATIENT_ID/SAMPLE_ID"
  GENIE_SP_CLINICAL }o--|| GENIE_SP_CLINICAL : "NAME has many PATIENT_ID/SAMPLE_ID"

```

EDIT: For less confusion, I will close this PR and use a PR with the updated ingestion scripts and description. Everything else will more or less stay the same

thomasyu888:
> After a lot of trying things and experimenting, I think I'll have to bite the bullet and ingest the genie releases into one schema for each project otherwise the downstream effects esp. when querying the tables are much. This new pivot feels like the lesser evil. For the purpose of this PR, I'm just going to scope this just for the clinical tables because that's what we need for the rest of the initiative, and clinical tables are relatively straightforward to handle.
> 
> I also will avoid ingesting releases for main genie that are too old (e.g: older than 5 years ago) for this PR if there are too many issues getting them with the much newer releases.
> 
> Each project will have a schema MAIN, BPC, and SP. Each schema will contain the specific file format tables for that project.
> 
> So a `CLINICAL_SAMPLE` table for main genie would look like:
> | SAMPLE_ID | PATIENT_ID | RELEASE | RELEASE_TYPE | MAJOR_VERSION | MINOR_VERSION | OTHER_COLUMNS | 
> |-----------|------------|---------|--------------|---------------|---------------|---------------|
> | GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_2_CONSORTIUM | CONSORTIUM | 17 | 2 | ... |
> | GENIE-SAGE-1-2 | GENIE-SAGE-1 | 17_2_CONSORTIUM | CONSORTIUM | 17 | 2 | ... |
> | GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_0_PUBLIC | PUBLIC | 17 | 0 | ... |
> | GENIE-SAGE-1-2 | GENIE-SAGE-1 | 17_0_PUBLIC | PUBLIC | 17 | 0 | ... |
> | GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_3_CONSORTIUM | CONSORTIUM | 17 | 3 | ... |
> 
> 
> A `CANCER_PANEL_TEST` table for BPC will look like the above but with an addition `COHORT` column. 
> 
> SP project table will just be one table where there won't be any of the RELEASE, RELEASE_TYPE, MAJOR_VERSION, MINOR_VERSION columns but a column containing the SP Project name like `NAME`. OR we can have one table per SP project since their table schemas vary from project to project (but the query will just be elongated as we have 7 SP projects and this number will grow in the future adding to our query length)
> 
> ```mermaid
> erDiagram
>   GENIE_MAIN_CLINICAL_SAMPLE {
>     string  SAMPLE_ID
>     string  PATIENT_ID
>     string  RELEASE
>     string  RELEASE_TYPE
>     int     MAJOR_VERSION
>     int     MINOR_VERSION
>     string  OTHER_COLUMNS
>   }
> 
>   GENIE_BPC_CANCER_PANEL_TEST {
>     string  COHORT
>     string  SAMPLE_ID
>     string  PATIENT_ID
>     string  RELEASE
>     string  RELEASE_TYPE
>     int     MAJOR_VERSION
>     int     MINOR_VERSION
>     string  OTHER_COLUMNS
>   }
> 
>   GENIE_SP_CLINICAL {
>     string  NAME
>     string  SAMPLE_ID
>     string  PATIENT_ID
>     string  OTHER_COLUMNS
>   }
> 
>   %% Conceptual "grouping" relationships (not actual foreign keys)
>   GENIE_MAIN_CLINICAL_SAMPLE }o--|| GENIE_MAIN_CLINICAL_SAMPLE : "RELEASE has many PATIENT_ID/SAMPLE_ID"
>   GENIE_BPC_CANCER_PANEL_TEST }o--|| GENIE_BPC_CANCER_PANEL_TEST : "COHORT has many RELEASEs; each RELEASE has many PATIENT_ID/SAMPLE_ID"
>   GENIE_SP_CLINICAL }o--|| GENIE_SP_CLINICAL : "NAME has many PATIENT_ID/SAMPLE_ID"
> 
> ```
> 
> EDIT: For less confusion, I will close this PR and use a PR with the updated ingestion scripts and description. Everything else will more or less stay the same

Thanks for the research, I agree with this pivot and that it is the lesser evil and for this particular work, ingesting just the sample and patient identifiers is a good plan. A bit unfortunate but maybe we can add in data after the fact. The older releases should have SAMPLE_ID and PATIENT_ID columns though so there shouldn't be schema issues.

PR #130, opened 2025-10-24
Als kp dag refactor

ramayyala:
# **Problem**:
The existing ALS KP Dataset Collection DAG (als-kp-dataset-dag.py) relies on a manual process for handling new and updated datasets from the C-Path API. 
The specific technical problems are:
   - Manual Duplicate Resolution: The DAG identifies potential duplicates based on a title match. It then sends a Slack message to a channel where a human data manager must manually review the datasets and update a separate JSON file in Synapse to specify which ones to ignore. This process requires human intervention for the pipeline to run correctly.
   - No Versioning Concept: The DAG cannot distinguish between different versions of the same underlying dataset (e.g., a raw src dataset vs. a cleaned fv1 dataset). It treats them as potentially separate, duplicate entities rather than versions of the same thing.
  
# **Solution:**
I have replaced the original DAG with a fully automated, version-aware DAG (als-kp-dataset-dag-v2.py). The new solution programmatically handles versioning and updates, eliminating the need for manual intervention.
   1. Automated Version Selection: Instead of flagging duplicates, the new DAG implements a "latest and greatest" selection logic.
      - Dataset Code Parsing: Helper functions (parse_dataset_code, extract_dataset_code_from_url) have been added to parse the dataset's version, unique ID (ALS number), and date from its URL.
      - Versioning Priority: A priority hierarchy is defined (fv3 > fv2 > fv1 > fm2 > fm1 > src), allowing the DAG to understand which version of a dataset is "better".
       - **select_latest_versions** Task: This task groups datasets by their unique ALS number and automatically selects the single best version based on the priority score, using the date as a tie-breaker.

   2. The DAG now distinguishes between creating new datasets and updating existing ones.
       - **identify_dataset_actions** Task: This new task compares the selected best versions from the API with the datasets already in the Synapse Collection. It sorts datasets into two lists: those that need to be created from scratch and those that are updates to existing datasets.
       - **update_existing_datasets** Task: For datasets identified as updates, this task fetches the existing dataset from Synapse and creates a new version of it with the updated metadata. This preserves the
         dataset's history in Synapse.
       - **create_new_datasets** Task: This task ingests only the datasets that are completely new to the collection.

   3. Removal of Manual Steps: The entire Slack notification workflow (generate_slack_message, post_slack_messages) and the reliance on the manually curated ignore_cpath_datasets.json file have been completely removed from the DAG.

   4. Annotation Refresh: A new refresh_collection_annotations task was added to ensure that the collection's main summary table is immediately updated to reflect the new metadata from any newly versioned datasets.

# **Testing:**

  The solution was first validated and manually run by converting it into a notebook solution and testing it on both actual CPath data pulled from the API and on dummy data to really test conditionals. All of these tests would add the datasets to a  dummy Synapse Dataset Collection (syn69962707) to serve as a testing environment, ensuring no impact on the production collection. After confirming manual test runs, results from actual Cpath data pull were added to the production Dataset Collection due to Cpath's request for updated data ingest on our end. Afterwards, was tested in a local airflow install environment.

# **To Do**: 
- Still needs to be tested in the DPE airflow environment
- Need to setup with a long live token from Cpath and store that in the airflow secrets


dags/als-kp-dataset-dag.py
```diff
@@ -28,25 +28,25 @@
...
-        "https://raw.githubusercontent.com/amp-als/data-model/410a429540a81a63846921ee77d6cf8e33ab6407/json-schemas/Dataset.json",
+        "https://raw.githubusercontent.com/amp-als/data-model/refs/heads/main/json-schemas/Dataset.json",
         type="string",
     ),
     "cpath_api_url": Param(
         "https://fair.dap.c-path.org/api/collections/als-kp/datasets", type="string"
     ),
     "ignore_cpath_datasets": Param("syn68737367", type="string"),
-    "collection_id": Param("syn66496326", type="string"),
+    "collection_id": Param("syn69962707", type="string"),
```
linglp, line 46: this will need to be changed back to the production collection id later right?

dags/als-kp-dataset-dag.py
```diff
@@ -161,6 +161,117 @@ def transform_with_jsonata(
...
+
+    Returns:
+        Tuple of (als_number, prefix, date_str)
+
+    Raises:
+        ValueError: If dataset code format is invalid
+    """
+    import re
+
+    # Pattern to match: prefix_als{number}_{date}
```
linglp: what would happen if an example doesn't match?

dags/als-kp-dataset-dag.py
```diff
@@ -161,6 +161,117 @@ def transform_with_jsonata(
...
+        return ""
+    return url.split("/")[-1]
+
+
+def get_version_priority(prefix: str) -> int:
+    """Get priority score for dataset prefix (higher = better).
+
+    Priority: fv3 > fv2 > fv1 > fm2 > fm1 > src
+    """
+    priority_map = {
```
linglp: If there's a new version comes up, it seems like this priority map would need to be manually maintained and updated.

dags/als-kp-dataset-dag.py
```diff
@@ -161,6 +161,117 @@ def transform_with_jsonata(
...
+    from datetime import datetime
+
+    # Group datasets by ALS number
+    als_groups = defaultdict(list)
+
+    for dataset in datasets:
+        url = dataset.get("url", "")
+        dataset_code = extract_dataset_code_from_url(url)
+
+        if not dataset_code:
```
linglp: Datasets without a `url` will be silently skipped. Is that the expected behavior?

dags/als-kp-dataset-dag.py
```diff
@@ -221,103 +332,157 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
-        This task formats a message that includes information about datasets flagged as duplicates,
-        which will be posted to a Slack channel for review.
+        # Apply latest and greatest selection using URL field for ALS codes
+        selected_items = select_latest_versions(new_items)
 
-        Arguments:
-            duplicates (List[Dict[str, Any]]): A list of datasets identified as duplicates.
+        print(f"Selected {len(selected_items)} datasets from {len(new_items)} new items")
 
+        # Return selected items and empty duplicates list (no manual review needed)
```
linglp: I think if `duplicates` just going to be an empty list, then you shouldn't return it at all.

philerooski:
Thanks for the PR, @ramayyala ! I'm not familiar enough with ALS KP to give a proper review of the DAG, but I want to let you know about our new development standards for DAGs. No need to make any changes to this PR, this is just an fyi for future DAGs.

All the details are in CONTRIBUTING.md, but the gist is that we want to separate the DAG task logic from the DAG itself as much is practical in order to make testing of DAGs easier. Check out [this Slack discussion](https://sagebionetworks.slack.com/archives/C02M7QR8D46/p1760656301608079) for the motivation and example of how this looks.

dags/als-kp-dataset-dag.py
```diff
@@ -221,103 +332,157 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
 
     @task
     def find_duplicated_datasets(
-        transformed_items: List[Dict[str, Any]], **context
-    ) -> List[Dict[str, Any]]:
-        """Flag duplicated new C-PATH datasets.
+    transformed_items: List[Dict[str, Any]], **context
+    ) -> Dict[str, List[Dict[str, Any]]]:
+        """Apply latest and greatest version selection for ALS datasets.
```
linglp: Since you have completely changed the goal of this function, this function should be renamed to something else.

dags/als-kp-dataset-dag.py
```diff
@@ -161,6 +161,117 @@ def transform_with_jsonata(
...
+        except ValueError as e:
+            print(f"Skipping dataset with invalid code: {dataset_code}, error: {e}")
+            continue
+
+    selected_datasets = []
+
+    # For each ALS group, select the best version
+    for als_number, versions in als_groups.items():
+        # Sort by priority (desc), then by date (desc)
+        best_version = max(versions, key=lambda x: (
```
linglp: will sort by date alone be simpler and more reliable since then you don't need to manually maintain a priority map?

dags/als-kp-dataset-dag.py
linglp: if there's a `fv4`, the priority will be 0 which can be problematic

dags/als-kp-dataset-dag.py
```diff
@@ -349,140 +514,203 @@ def find_ignored_datasets(**context) -> List[str]:
...
+        ) -> str:
+            """Create brand new datasets for ALS numbers that don't exist yet."""
+
+            syn_hook = SynapseHook(context["params"]["synapse_conn_id"])
+            dataset_collection = DatasetCollection(id=context["params"]["collection_id"]).get()
+
+            created_dataset_ids = []
+
+            for item in datasets_to_create:
+                # Skip if in ignored list
```
linglp: I don't think this is needed anymore in the new design since we are not going to manually maintain an ignored list

dags/als-kp-dataset-dag.py
```diff
@@ -349,140 +514,203 @@ def find_ignored_datasets(**context) -> List[str]:
...
+                }
+                # Create new dataset
+                dataset = Dataset(
+                    parent_id=context["params"]["project_id"],
+                    name=item["title"],
+                    annotations=dataset_annotations,
+                    description=dataset_description).store()
+                dataset_collection.add_item(dataset)
+                created_dataset_ids.append(dataset.id)
+                #Storing annotations using synapseclient rather than models method cause it would not store with models method for some reason.
```
linglp: The comments here are outdated and should be removed?

linglp (commented):
Hi @ramayyala ! Thanks for your hard work. I am thinking: 
* Currently, you determine the latest dataset by sorting based on both the priority map and the date in the URL. I wonder if it might be simpler to rely only on the date. That way, you wouldn’t need to update or maintain the priority map whenever new datasets with new version codes (e.g., fv4) are added.
* Since the select_latest_versions function already determines the latest dataset versions, it might make sense to call it earlier in the pipeline to retrieve only the latest datasets from C-Path. The order will be something like:
```
    data = fetch_cpath_data()
    transformed_items = transform_data(data)
    selected_items = select_latest_versions(transformed_items)
    dataset_actions = identify_dataset_actions(selected_items)
```
That can simplify the logic quite a bit. 
* To test, I am thinking there are three main cases: 
1) synapse has A1, B1, and C-Path data has A1, A2, B1. The expected behavior is that dataset A gets a version update and nothing changes regarding dataset B1. 
2) Synapse has A2, B1 and C-Path data has A1, A2, B1. The expected behavior is nothing gets changed since Synapse already has the latest dataset. 
3) Synapse doesn't have any datasets, but C-Path has A1, A2, B1. The expected behavior is A2 and B1 get added to Synapse. 
I am curious have you tested all three cases?

dags/als-kp-dataset-dag.py
ramayyala: yes, this was just a dummy dataset collection i created so we could test before messing with the prod collection

dags/als-kp-dataset-dag.py
ramayyala: the assumption was that it should always match as i doubt cpath would change this format anytime soon, but we can build in an exception case for this

dags/als-kp-dataset-dag.py
ramayyala: Yeah, that's true, but I also don't think there would be a new version anytime soon like that. This is what they said is there typical protocol, but I think you are right that maybe we should build around it by just focusing on the dates. I think maybe what we do is have the dates be the prioritized way to check for latest versions, and then have a secondary check for fv3, fv2, etc. That way even if the secondary check fails, it still has the date check that should generally always work.

dags/als-kp-dataset-dag.py
ramayyala: Yeah you are right, this was just extraneous code from the original that needs to be removed. I'll take care of this

dags/als-kp-dataset-dag.py
ramayyala: Let me double check that sorting by date would result in the same output and get back to you on this. It should, but good to double check

dags/als-kp-dataset-dag.py
ramayyala: Ok will do

dags/als-kp-dataset-dag.py
ramayyala: Oh yeah this was from notebook testing, i'll clean this up

dags/als-kp-dataset-dag.py
BryanFauble: Something to consider on the date check - Could the date for an "older" version be updated at any point?

What I mean by that, suppose you have:
fv1_als1001_2025_06_26
fv2_als1001_2025_02_26

Is something like this possible? Where `fv2` is the "latest" that we want to use, however, an earlier version has a newer date.

FWIW, I have seen this is other projects (Not this one) where old pieces of data get patched and have a new date attached to it.

ramayyala: yeah this is something I was considering as well. I plan on asking the Cpath team about this to see if it could even occur from there end.

dags/als-kp-dataset-dag.py
```diff
@@ -221,103 +332,157 @@ def transform_data(data: Dict[str, Any], **context) -> List[Dict[str, Any]]:
...
+                    'synapse_id': row['id'],
+                    'dataset_code': dataset_code,
+                    'prefix': prefix,
+                    'priority': get_version_priority(prefix),
+                    'date_str': date_str,
+                    'sameAs': row['sameAs'],
+                    'url': row['url']
+                }
+            except ValueError:
+                continue
```
BryanFauble: Be careful swallowing exceptions without logging what the underlying failure was. In this case this could silently fail

thomasyu888:
Just an update as I spoke to Ram:

> We are currently waiting on CPath to get back to us on some questions we had about their formatting for their date and version system for edge cases.

PR #132, opened 2025-12-06
[GEN-2268, GEN-2269, GEN-2282] Create patient sample tracking DAG

rxu17:
# **Problem:**
There was an external request asking for the ability to track the sample-patients in genie across our three projects - main genie, BPC and sponsored projects (SP) and see if the **latest versions** of the patient-sample pairs in the BPC and SP projects are present in the **latest version** patient-sample pairs of the Genie project. For SP projects, there are no latest versions as each project is uniquely named. 

An example request might be: 

For the latest consortium release of the BRCA cohort in BPC, are all of the sample-patient pairs in that present in the latest consortium release of the main genie project? 

And one would use this table to find that out

JIRA Tickets: https://sagebionetworks.jira.com/browse/GEN-2268, https://sagebionetworks.jira.com/browse/GEN-2269

# **Solution:**

**Full Patient Sample Tracking Workflow**
```mermaid
flowchart TD
  A([New Release Available]) --> B[Update & run Snowflake ingestion scripts]

  B --> C{Any new BPC or SP project releases?}

  C -->|Yes| D[Update & generate new patient/sample tracking Synapse table schema]
  C -->|No| E[Skip schema update]

  D --> F[Update & run patient/sample tracking Airflow DAG]
  E --> F[Update & run patient/sample tracking Airflow DAG]

  F --> G[Query & analyze newly updated patient/sample tracking Synapse table]
```

**DAG structure**

```mermaid
flowchart LR
  A[Query Snowflake ingestion tables] --> B{Validate query result}
  B -- Pass --> C[Upload to Synapse table]
  B -- Fail --> D[Log + alert / fail task]
```

We want to query the genie releases in snowflake and create a patient-sample tracking table. Here are two proposed versions, which are still awaiting a final decision. This code currently reflects a query result that looks like version 2.

Version 1: https://www.synapse.org/Synapse:syn72246564/tables/
 
See here for the data model of the query
image
or you can visit https://www.synapse.org/Synapse:syn72246564 (need genie access) to see the full table and query

Then we would do some light validation.

Validation available:
Checks performed:
      1) DataFrame is non-empty.
      2) Required columns exist.
      3) No NULL/NaN values in the non-release name columns.
      4) No duplicates on key columns (defaults to SAMPLE_ID, PATIENT_ID).

Finally we would clear the results from the current table in Synapse and then upload the results of the query to the table in Synapse

# **Testing:**
- Tested DAG on staging table and DAG successfully updated the table with the result


dags/genie/genie-patient-sample-tracking-dag.py
```diff
@@ -0,0 +1,240 @@
...
+                FROM IDENTIFIER(
+                    'MY_DB.' || lcs.schema_name || '.CANCER_PANEL_TEST_LEVEL_DATASET'
+                )
+            ) t
+        ),
+
+        /* ------------------------------------------------------------ */
+        /* 3) PROJECT-SPECIFIC SCHEMAS (template mapping)                */
+        /* ------------------------------------------------------------ */
+        project_list AS (
```
rxu17: After going through this, I think it might make things easier if the SP project data was just saved under one schema called SP, and then just have each table named as the sponsored project name itself since we really only care about ingesting the single clinical file from each of the sponsored projects unlike BPC and main genie where we ingest all of the datasets. Will modify the SP ELT script

rxu17: Saved the BPC and main genie tables under one schema but each SP project gets its own schema due to too many differences between SP project data schemas

dags/genie/genie-patient-sample-tracking-dag.py
```diff
@@ -0,0 +1,358 @@
...
+    start_date=datetime(2025, 1, 1),
+    catchup=False,
+    default_args={"retries": 1},
+    tags=["snowflake", "genie"],
+    params=dag_params,
+)
+def build_patient_sample_tracking_table():
+
+    @task(task_id="query_validate_and_upload")
+    def query_validate_and_upload(**context) -> List[Dict]:
```
rxu17: For data safety, decided to keep the query, validate and uploading all in one task to avoid passing data between tasks

dags/genie/genie-patient-sample-tracking-dag.py
```diff
@@ -0,0 +1,358 @@
...
+        results = syn.tableQuery(
+            f"SELECT * FROM {PATIENT_SAMPLE_TRACKING_TABLE_SYNID}"
+        ).asDataFrame()
+        syn.delete(synapseclient.Table(PATIENT_SAMPLE_TRACKING_TABLE_SYNID, results))
+        syn.store(
+            synapseclient.Table(schema=PATIENT_SAMPLE_TRACKING_TABLE_SYNID, values=df)
+        )
+
+    @provide_session
+    def cleanup_xcom(session=None):
```
rxu17: May not be needed but unsure if xcom just auto gets used whenever the DAG tasks run?

dags/genie/genie-patient-sample-tracking-dag.py
```diff
@@ -0,0 +1,490 @@
...
+            SELECT * FROM sp_erbb2_pairs
+                UNION ALL
+            SELECT * FROM sp_fgfe_pairs
+                UNION ALL
+            SELECT * FROM sp_kras_pairs
+                UNION ALL
+            SELECT * FROM sp_ntrk_pairs
+        ),
+        /* 2) Wide flags: MAIN latest + one flag per project type */
+        wide AS (
```
rxu17, line 423: Main section of SQL that differs between v1 and v2 of the SQL statement (for v2, this section will be removed)

dags/genie/genie-patient-sample-tracking-dag.py
```diff
@@ -0,0 +1,490 @@
...
+from airflow.decorators import dag, task
+from airflow.models.param import Param
+from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
+from orca.services.synapse import SynapseHook
+from airflow.utils.db import provide_session
+from airflow.models import XCom
+
+logger = logging.getLogger(__name__)
+
+PATIENT_SAMPLE_TRACKING_TABLE_SYNID = "syn72246564"
```
rxu17: TODO: Update to be production synapse table

dags/genie/genie-patient-sample-tracking-dag.py
```diff
@@ -0,0 +1,490 @@
...
+        1) For each BPC cohort, find the most recent (MAJOR_VERSION, MINOR_VERSION)
+            present in the BPC cancer_panel_test_level table.
+            - QUALIFY + ROW_NUMBER keeps only the top (latest) version per cohort.
+        ---------------------------------------------------------------------------*/
+        bpc_latest_per_cohort AS (
+        SELECT
+            COHORT,
+            MAJOR_VERSION,
+            MINOR_VERSION
+        FROM GENIE_DEV.BPC_CLINICAL_FILES.CANCER_PANEL_TEST_LEVEL
```
rxu17: TODO: Update these to pull from `GENIE` once validated

dags/genie/genie-patient-sample-tracking-dag.py
thomasyu888, line 1: Interesting, I didn't realize you can store dags within folders!  What does this look like in the UI?

rxu17: The dag(s) shows up normally in a list with the rest of the dags, I believe

rxu17:
Added a [README](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f78b54b4051f74311de5ff6e2350bf0586ea5aa7/dags/genie/README.md) for the genie dags folder

dags/genie/genie-patient-sample-tracking-dag.py
```diff
@@ -0,0 +1,495 @@
...
+        validate_patient_sample_results(
+            df,
+            required_cols=REQUIRED_COLS,
+            key_cols=["SAMPLE_ID", "PATIENT_ID"],
+        )
+
+        # Delete all rows in current table and upload new results to Synapse table
+        syn = SynapseHook(context["params"]["synapse_conn_id"]).client
+        table_id = context["params"]["patient_sample_tracking_table_synid"]
+        to_delete = syn.tableQuery(
```
danlu1, line 476: Just out of curiosity. Why not upsert rows?

dags/genie/genie-patient-sample-tracking-dag.py
```diff
@@ -0,0 +1,495 @@
...
+        for start in range(0, len(df), CHUNK):
+            logger.info(f"Uploading batch {start} ...")
+            syn.store(
+                synapseclient.Table(
+                    schema=table_id,
+                    values=df.iloc[start:start+CHUNK],
+                )
+            )
+
+    query_validate_and_upload()
```
danlu1, line 492: I'm not familiar with Airflow dag. But this is part of the build_patient_sample_tracking_table dag. Do we have to call it separately?

dags/genie/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+This DAG typically requires updates under the following scenarios:
+
+#### Scenario 1: A New Sponsored Project (SP) Is Added
+
+When a new SP dataset becomes available, you must:
+
+- Add a new `sp_<project>_pairs` Common Table Expression(CTE) block. See existing SP projects (e.g `KRAS SP project`) as an example.
+- Add a new inclusion flag column `IN_<SP_PROJECT_NAME>_RELEASE` (e.g: `IN_AKT1_PROJECT`) in the wide CTE
+- Add the corresponding required column to `REQUIRED_COLS`
```
danlu1: As the table schema, can you specify only `IN_<SP_PROJECT_NAME>_RELEASE` is needed to be added to `REQUIRED_COLS`?

danlu1 (approved):
LGTM. Just a few comments.

dags/genie/genie-patient-sample-tracking-dag.py
rxu17: We have to completely refresh because we are always getting the latest release of the projects - so when a new release comes in, we'd refresh the patient_id, sample_id information and pretty much all of the <project>_release column values (for all rows) would change with a new release and I wouldn't want any old release data in there. Went with this way since it seemed the cleanest.

dags/genie/genie-patient-sample-tracking-dag.py
rxu17: Yes since we only have a function definition if we don't call it here.

dags/genie/README.md
rxu17: Yes there's no other columns that we are adding in SP project section. The 1st bullet is just adding a section to the snowflake query. 

I can do the following to make it clearer
```suggestion
- Add the new column `IN_<SP_PROJECT_NAME>_RELEASE` to `REQUIRED_COLS`
```

dags/genie/genie-patient-sample-tracking-dag.py
danlu1: Gotcha. Makes sense.

PR #136, opened 2025-12-19
Add programmatic grant scrapping with claude API

rxu17:
# **Problem:**
Not all grant websites have APIs or databases with data structures easy to scrape using current web scraping tools. Often times we just have a webpage with a list of grants and information formatted throughout. How can we gather the grant information we need that we can get easily from grant websites with APIs.

# **Solution:**
This adds a grant scrapping tool that takes URLs that directly take you to the website page that contains the list of grants and gathers the required information (using the grant ingestion data model) from them via claude API programmatically.

Claude API cannot access URLs directly so the HTML content of the website is first extracted and then claude is prompted to analyze it and extract the grant data

Limitations - have not tested sites where grant info are hidden behind links / PDF links or have a lot of javascript content that you need to click to get to the grant.

# **Testing:**
Tested on a couple grant websites.
If you have access, see here for an full example: https://www.synapse.org/Synapse:syn72011146

Preview:
image

### DIFF
--- a/src/grants/scrape_non_federal_grants.py
+++ b/src/grants/scrape_non_federal_grants.py
@@ -1,19 +1,3 @@
-"""
-Dependencies: 
-- Access to the data model synapse CSV
-- Developer account on Anthropic
-- API Key token saved as ANTHROPIC_API_KEY in your environment
-- Valid Synapse account with synapse configured in your environment
-
-How to Run:
-
-1. Gather list of 
-2. Set up your environment via:
-pip install anthropic requests pandas synapseclient
-3. Provide a text file urls.txt with list of URLs or update the local urls variable
-4. python3 scrape_grants.py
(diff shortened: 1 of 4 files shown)


src/grants/scrape_non_federal_grants.py
```diff
@@ -0,0 +1,250 @@
...
+</webpage_content>
+
+Please analyze this webpage and extract all grant opportunities listed.
+
+<data_model>
+{data_model_prompt}
+</data_model>
+
+EXTRACTION INSTRUCTIONS:
+1. Extract information for each grant based on the field specifications above
```
rxu17, line 94: Note that there is no filtering of grants currently in this scraping

thomasyu888 (approved):
🔥 really great POC here!

PR #137, opened 2025-12-26
[DPE-1521]: Create DAG to ingest minimal JSONLD for data catalog items

linglp:
# **Problem:**
Google is having difficulty indexing the Croissant JSON being injected in portal pages. To improve discoverability, we need to inject minimal JSON-LD metadata into Synapse pages for datasets listed in the data catalog

Related to: https://sagebionetworks.jira.com/browse/DPE-1521
See design doc: https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4456349721/DPE+-+1521+Create+minimal+JSONLD+for+data+catalog+items

# **Solution:**
This pull request introduces a new Airflow DAG, `synapse_minimal_jsonld_dag.py`, which automates the creation and storage of minimal Croissant JSON-LD metadata files for Synapse datasets. The DAG queries a Synapse data catalog, generates JSON-LD files for each dataset, uploads them to S3, and records their S3 links in a Synapse table. The implementation includes robust telemetry with OpenTelemetry for tracing and logging, and customizes Synapse client behavior for reliability.

# **Test** 
Successful ingestion: syn72041138


linglp:
Also relates to: https://github.com/Sage-Bionetworks-Workflows/eks-stack/pull/75

dags/synapse_minimal_jsonld_dag.py
```diff
@@ -0,0 +1,399 @@
...
+    """
+    Create a Synapse client that can be used to query Synapse.
+
+    Returns:
+        The Synapse client.
+    """
+    syn_client: Synapse = Synapse(skip_checks=True)
+    syn_client._rest_call = MethodType(_rest_call_replacement, syn_client)
+    assert syn_client.credentials is None, "Synapse client is not logged out"
+    return syn_client
```
BryanFauble: This function doesn't seem to be called

dags/synapse_minimal_jsonld_dag.py
```diff
@@ -0,0 +1,399 @@
+
```
BryanFauble, line 1: Is it possible for the name of the script mention `croissant` in some way? The other script I wrote is called `synapse-dataset-to-croissant.py`

So if we model this after the S3 bucket name, this would end up:

`synapse-dataset-to-croissant-minimal.py`

dags/synapse_minimal_jsonld_dag.py
```diff
@@ -0,0 +1,399 @@
...
+    "default_args": {
+        "retries": 1,
+    },
+    "params": dag_params,
+}
+
+SYNAPSE_DATA_CATALOG = "syn61609402"
+# AWS related constants
+REGION_NAME = "us-east-1"
+BUCKET_NAME="synapse-croissant-metadata-minimal"
```
BryanFauble, line 86:
```suggestion
BUCKET_NAME = "synapse-croissant-metadata-minimal"
```

dags/synapse_minimal_jsonld_dag.py
```diff
@@ -0,0 +1,399 @@
...
+            headers=headers,
+            auth=auth,
+            timeout=70,
+            **kwargs,
+        ),
+        verbose=self.debug,
+        **retryPolicy,
+    )
+    self._handle_synapse_http_error(response)
+    return response
```
BryanFauble, line 222: This was the TODO comment I had where you grabbed this code from:

```
# TODO: Remove this on the next > 4.7.0 release of the Synapse Python Client
# This is a temporary hack to include the changes from: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1188
# The hack is used here because the current SYNPY client does not have an HTTP timeout
# for requests to Synapse. As a result and due to the significant number of HTTP calls that
# occur during the DAG, the DAG can stall and never return due to the requests library.
# https://requests.readthedocs.io/en/latest/user/advanced/#timeouts
```


If you also wanted to take on upgrading the SYNPY version (I can help with any deployments) then we can remove this code.

It is a little more involved to update a dependency version:

1. https://github.com/Sage-Bionetworks-Workflows/py-orca/commit/9605595da9da8967d9d6ab669ed2ebe4bc220680 should be upgraded
2. py-orca should be tagged and released
3. https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/b8417b6f98fe625b61ac890c96a8729de931ae7b/requirements-airflow.txt#L5 should be updated
4. https://github.com/Sage-Bionetworks-Workflows/orca-recipes/releases should be tagged and released
5. https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/983d35d8d8674da040ad017ce353acac46722542/modules/apache-airflow/templates/values.yaml#L71 needs to be incremented to the new image version and deployed in spacelift
6. The application needs to be sync'd in ArgoCD

dags/synapse_minimal_jsonld_dag.py
```diff
@@ -0,0 +1,399 @@
...
+                "url": link,
+                "identifier": dataset_id,
+                "creator": {
+                    "@type": "Organization",
+                    "name": "Sage Bionetworks" if pd.isnull(row["contributors"]) else row["contributors"]
+                },
+                "license": "" if pd.isnull(row["license"]) else row["license"]
+            }
+
+            s3_key = f"{dataset_name}_{dataset_id}.minimal_croissant.jsonld"
```
BryanFauble, line 602: The version has to be encoded into the name of this file name. This is because all the files are uploaded into the same location. It should be (I think) valid for multiple versions of the same Dataset to exist in the same Table.

`s3_key = f"{dataset.name}_{dataset_id}.{dataset_version}_datasetCollection_{dataset_collection}_croissant.jsonld"` is the format I used in the other script.

dags/synapse_minimal_jsonld_dag.py
```diff
@@ -0,0 +1,399 @@
...
+            execute_push_to_s3(dataset=dataset,dataset_id=row["id"],s3_key=s3_key, croissant_file=minimal_croissant_file, push_to_s3=push_to_s3, **context)
+            
+            s3_url = f"https://{BUCKET_NAME}.s3.us-east-1.amazonaws.com/{quote_plus(s3_key)}"
+            execute_push_to_synapse(push_to_synapse=push_to_synapse, dataset=dataset, dataset_id=dataset_id, s3_url=s3_url, **context)
```
BryanFauble, line 606: You are missing cleanup logic within this script. Check how I am doing this in the other croissant script.

Cleaning up non-current files in both S3 and Synapse should be a requirement because if a file is no longer public or no longer exists in the Table we want to drop the croissant file as well on the next run to make it no longer available

BryanFauble (changes requested):
There are a few changes I noted. Otherwise the logic is looking great. Please re-request after a few small changes.

dags/synapse_minimal_jsonld_dag.py
linglp: I looked at the data collection in synapse-dataset-to-croissant [here](https://www.synapse.org/Synapse:syn50913342/datasets/) and noticed the `currentVersion` column, which indicates that different versions can coexist in the same table.

However, when I checked the data catalog [table](https://www.synapse.org/Synapse:syn61609402/tables/), I don't see a version column, so I am assuming versioning isn't needed there.

@thomasyu888 Could you clarify whether version tracking is required?

thomasyu888: @linglp version tracking is not required at this time since some of these data catalog datasets point to synapse folders

BryanFauble (approved):
I read over the new round of changes and added a small doc update, but the logic all looks right to me.

PR #140, opened 2026-01-07
[GEN-2263, GEN-2264] Add genie module with main genie, BPC and SP ingestion scripts

rxu17:
# **Problem:**
This is branched off of #129 (but we will be closing #129 and using this instead) except the scope of the ingestion scripts are limited to just the `clinical` data and each table now contains all of the consortium and public releases (for ease of querying and access) under just one schema instead of one schema per release. This is because the scope of the project is to just get the clinical patient and sample pairs. In the future if the need comes up, we can do this for other fileformats. Also see https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/129#issuecomment-3658925617 for more context, but to avoid further downstream extreme complexity in snowflake query + airflow, this PR was created instead.

As well, genie data changes quite a bit throughout the releases so trying. 

# **Solution:**
Each genie project requires different handling so there's one script for each project. Currently we just ingest clinical files for all of the projects:

- Main genie will be the most automated in ingesting new releases. There will be one schema where all of the releases (public and consortium) are stacked into one table called `CLINICAL_SAMPLE`

So a `CLINICAL_SAMPLE` table for main genie would look like:
| SAMPLE_ID | PATIENT_ID | RELEASE | RELEASE_TYPE | MAJOR_VERSION | MINOR_VERSION | OTHER_COLUMNS | 
|-----------|------------|---------|--------------|---------------|---------------|---------------|
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_2_CONSORTIUM | CONSORTIUM | 17 | 2 | ... |
| GENIE-SAGE-1-2 | GENIE-SAGE-1 | 17_2_CONSORTIUM | CONSORTIUM | 17 | 2 | ... |
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_0_PUBLIC | PUBLIC | 17 | 0 | ... |
| GENIE-SAGE-1-2 | GENIE-SAGE-1 | 17_0_PUBLIC | PUBLIC | 17 | 0 | ... |
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | 17_3_CONSORTIUM | CONSORTIUM | 17 | 3 | ... |

- Genie BPC will need a config update to ingest releases. This is due to the structure of the BPC projects
A `CANCER_PANEL_TEST` table for BPC will look like the above for main genie but with an addition `COHORT` column:

| SAMPLE_ID | PATIENT_ID | COHORT | RELEASE | RELEASE_TYPE | MAJOR_VERSION | MINOR_VERSION | OTHER_COLUMNS | 
|-----------|------------|---------|---------|--------------|---------------|---------------|---------------|
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | BRCA | 2_2_CONSORTIUM | CONSORTIUM | 2 | 2 | ... |
| GENIE-SAGE-1-2 | GENIE-SAGE-1 | BRCA | 2_2_CONSORTIUM | CONSORTIUM | 2 | 2 | ... |
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | BRCA | 2_0_PUBLIC | PUBLIC | 2 | 0 | ... |
| GENIE-SAGE-1-2 | GENIE-SAGE-1 | BRCA | 2_0_PUBLIC | PUBLIC | 2 | 0 | ... |
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | BRCA | 2_1_CONSORTIUM | CONSORTIUM | 2 | 1 | ... |
 

- Genie SP will only ingest the clinical files as the rest as not as relevant. There will be one schema-table per SP project because each project's clinical table fields are too vastly differently for them to be stacked into the same table. As a result, downstream our query to put all of the tables together will be more complicated.

| SAMPLE_ID | PATIENT_ID | RELEASE | OTHER_COLUMNS | 
|-----------|------------|---------|---------|
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | SP_PROJECT_NAME | ... |
| GENIE-SAGE-1-2 | GENIE-SAGE-1 | SP_PROJECT_NAME | ... |
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | SP_PROJECT_NAME | ... |
| GENIE-SAGE-1-2 | GENIE-SAGE-1 | SP_PROJECT_NAME | ... |
| GENIE-SAGE-1-1 | GENIE-SAGE-1 | SP_PROJECT_NAME | ... |

Creating a common `snowflake_utils` module was needed.

- This allows us to use local snowflake connections (when we are running the scripts locally) as well as snowflake's AirflowHook's connection when we are running it in the DAG. We will be using the GENIE_SERVICE user's specific snowflake connection: SNOWFLAKE_GENIE_SERVICE_RAW_CONN
- Some other reusable functions that were repeated in the ingestion scripts were also added here

Extras:
- Needed to make a new Airflow Snowflake connection and Airflow Synapse connection for snowflake

# **Testing:**
- Unit tests pass

## General Checks for `snowflake_utils` connector functions
- Tested the snowflake_utils connector functions  in a simple Airflow DAG and checked that the connection with airflow works
- Tested the snowflake_utils connector functions locally and make sure the functions worked (just run the genie elt scripts)
   - Got expected logging in airflow DAG triggers `[2025-12-03, 04:13:04 UTC] {python.py:240} INFO - Done. Returned value was: ok_cursor`
   - Got expected logging in airflow DAG triggers `[2025-12-03, 04:13:03 UTC] {python.py:240} INFO - Done. Returned value was: ok_connection`
- Test that the genie elt script outputs in `GENIE_DEV` are as expected

## General Checks for other helper `snowflake_utils` functions
   - Check that there are no duplicates when `overwrite` is true when a table is overwritten: (logging is correct with `Overwriting table 'clinical_patient'.`)
 ```sql
    SELECT
      PATIENT_ID,
      COUNT(*) AS dup_count
    FROM GENIE_DEV.PUBLIC_4_0.CLINICAL_PATIENT
    GROUP BY PATIENT_ID
    HAVING COUNT(*) > 1
    ORDER BY dup_count DESC;
```
    Produces 0 results

    - Check that no table is overwritten when `overwrite` is false:, logging example
       -`2025-12-02 20:48:52,818 [INFO] snowflake_utils: Skipping write — 'gene_matrix' exists and overwrite=False`

## Checks that the uploaded table and the file contents are exactly the same
- Ran some checks against some tables in `GENIE_DEV` and no differences were found

Example checks
```python
import pandas as pd
from pandas.testing import assert_frame_equal
import numpy as np

from snowflake_utils import get_connection, logger, write_to_snowflake
import synapseclient

conn = get_connection()
syn = synapseclient.login()

def normalize_df(df: pd.DataFrame, key_cols):
    df2 = df.copy()

    # consistent column order
    df2 = df2.reindex(sorted(df2.columns), axis=1)

    # sort rows deterministically
    df2 = df2.sort_values(key_cols).reset_index(drop=True)

    # normalize NaN/None to a single representation
    df2 = df2.where(pd.notnull(df2), None)

    return df2

def df_equals_snowflake(conn):
    df_local = pd.read_csv(syn.get("syn11932079", followLink = True).path, sep = "\t", comment = "#")
    key_cols = ["PATIENT_ID"]
    with conn.cursor() as cs:
        cs.execute("SELECT * FROM <database>.<schema>.<table>;")
        df_sf = cs.fetch_pandas_all()

        # same columns?
        if set(df_local.columns) != set(df_sf.columns):
            return False, {"reason": "column mismatch",
                        "local_only": set(df_local.columns) - set(df_sf.columns),
                        "sf_only": set(df_sf.columns) - set(df_local.columns)}

        df_l = normalize_df(df_local, key_cols)
        df_s = normalize_df(df_sf, key_cols)

        # align dtypes if needed (optional, but helps)
        df_s = df_s.astype(df_l.dtypes.to_dict(), errors="ignore")

        try:
            assert_frame_equal(df_l, df_s)
            print("No differences!")
        except Exception as e:
            print(e)
            
        equal = df_l.equals(df_s)
        if equal:
            return True, {}

        # show where they differ
        diff = df_l.compare(df_s, keep_equal=False)
        
        cs.close()
        return False, {"diff": diff}
    
    
df_equals_snowflake(conn)
```


requirements-airflow.txt
```diff
@@ -1,11 +1,12 @@
 apache-airflow-providers-amazon ~=9.2.0
 apache-airflow-providers-celery >=3.10.0
 apache-airflow-providers-snowflake ~=6.0.0
-snowflake-connector-python ~=3.13.2
+snowflake-connector-python ~=3.15.0
```
rxu17, line 4: Needed to upgrade snowflake-connector-python to >= 3.14 due to this issue: https://community.snowflake.com/s/article/SnowSQL with local

dags/genie-ingestion-dag.py
rxu17, line 1: This will be deleted upon approval of this PR

CONTRIBUTING.md
danlu1, line 1: Very clear instruction makes me feel like I can set the env up without any further questions.

src/genie/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Example Usage
+
+Run ingestion to Snowflake to the genie dev database. for Genie Sponsored Projects (SP). This will also overwrite any pre-existing data in the tables in the GENIE_DEV database and create new tables for new data when applicable.
+
+```bash
+python3 src/genie/genie_sp_ingesion.py --database GENIE_DEV --overwrite
+```
+
+Run ingestion to Snowflake for Genie Biopharma Collaborative (BPC) Project to the genie dev database. This will not overwrite any pre-existing data in the tables in the GENIE_DEV database and will instead skip. This will still create new tables for new data when applicable.
```
danlu1: you mean skip writing to the table?

src/genie/genie_bpc_ingestion.py
```diff
@@ -0,0 +1,472 @@
...
+    minor_version: int
+
+
+def parse_bpc_version(cohort: str, version: str) -> ReleaseInfo:
+    """Parses the ReleaseInfo information from the release folder
+    where the release folder is expected to be of the format:
+    XX.X-consortium or XX.X-public
+
+    E.g:
+     2.1-consortium
```
danlu1: Can you add here to indicate before and after?

src/genie/genie_bpc_ingestion.py
```diff
@@ -0,0 +1,472 @@
...
+        if df.empty:
+            logger.warning(
+                f"[{release_info.cohort} {release_info.release}] Empty clinical file: {raw_name}"
+            )
+            continue
+
+        # remove cohort column to avoid dups if it exists
+        if "cohort" in df.columns:
+            df = df.drop(columns=["cohort"])
+
```
danlu1, line 284: How does this avoid dups? Do you mean it ensures duplicates of simple_patient pair to be removed in a later step, say table writing step?

danlu1 (commented):
Just half way through the reviewing. Leave comments first.

src/genie/genie_bpc_ingestion.py
rxu17: It avoids duplicated columns because the bpc data can be inconsistent and can already have the cohort column already in the data. For standardizing, I decided to remove the pre-existing cohort column instead and re-create it. I can make a note of this in the docstring of the function.

src/genie/README.md
rxu17: Yup. I think I need to revise this whole section

src/genie/genie_bpc_ingestion.py
rxu17: Yup

src/genie/genie_bpc_ingestion.py
danlu1: Oh I misunderstood this simply by looking at the inline comment. I think it's better to rephrase the comments to indicate the cohort column duplicates instead of data itself.

src/genie/genie_bpc_ingestion.py
danlu1, line 1: Add a note in readme that this step is used to ingest all data from a release and data deduplication will happen in table writing step.

src/snowflake_utils/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+    @task()
+    def test_get_connection(**context):
+        snow_hook = SnowflakeHook(context["params"]["snowflake_conn_id"])
+        airflow_conn = snow_hook.get_conn()
+
+        ... # other code
+
+        # call your ELT with the Airflow-owned conn
+        main_genie_elt.main(
```
danlu1: update the module name

src/snowflake_utils/snowflake_utils.py
```diff
@@ -0,0 +1,207 @@
...
+        return conn
+    return create_local_connection()
+
+
+@contextmanager
+def get_cursor(
+    conn: Optional[snowflake.connector.SnowflakeConnection] = None,
+) -> Iterator[snowflake.connector.cursor.SnowflakeCursor]:
+    """Yields a Snowflake cursor.
+        - If conn is passed (Airflow), we do NOT close it.
```
danlu1: Add a comment and indicate this conn will be closed by airflow in the end.

danlu1 (approved):

src/snowflake_utils/README.md
```diff
@@ -0,0 +1,112 @@
+# Snowflake Utils
```
thomasyu888, line 1: Really great readme here. Thanks!

PR #144, opened 2026-02-02
[GEN-2263] Update snowflake env vars and airflow access instructions

rxu17:
# **Problem:**
- Update the snowflake environment variable setup instructions to include piece about using `''` for values with characters like $
- Update the airflow instructions for connecting to Airflow locally

# **Testing:**
Tested through ticket validation

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,9 +106,15 @@ We can see which ports our Airflow services expose under `PORTS`:
 docker compose ps
 ```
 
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM - but i'll defer to someone else on the team for final review.

danlu1 (approved):
Thanks for adding this. We tested in our call and this works perfectly.

PR #145, opened 2026-03-03
improved scraping process

andrewelamb:
# **Problem:**

- Some of the scraped grant data was unable to be read because the presence of commas in the data
- Some of the URLs targeted for scraping weren't valid, didn't have grants, or weren't in a scrpable format.

# **Solution:**

- Added a file of URLs that work
- Changed the Claude instructiosn to use pipes as a delimiter


src/grants/urls.txt
rxu17, line 1: Will these be our main targets for looking for non-federal grants?

src/grants/scrape_non_federal_grants.py
rxu17, line 215: Nit: Should we update the README.md instructions to say to only read from the `urls.txt` since that's the official list (if it is). Then we can probably remove this exception since we want it to fail now (and not use default URLs) or we update the default urls list to be the more formal list we want (since now it's just 1 link)

rxu17 (approved):
LGTM! Just a couple comments

src/grants/scrape_non_federal_grants.py
```diff
@@ -94,16 +99,16 @@ def extract_grants_with_claude(url : str, data_model_df : pd.DataFrame) -> str:
 1. Extract information for each grant based on the field specifications above
 2. Pay careful attention to the description and data type for each field
 3. For example, if a field description mentions "email", extract email addresses for that field
-4. Return the results in CSV format with these exact column headers (in order): {', '.join(field_names)}
+4. Return the results in CSV format using pipe ('|') as the delimiter with these exact column headers (in order): {'|'.join(field_names)}
```
rxu17, line 102: Nit: Could be helpful if not already in the README to add an updated section about what some of the extracted outputs may look like

PR #147, opened 2026-04-02
[DPE-1507] Add DOI -> Synapse Table code framework

thomasyu888:
# **Problem:**

DOIs that are minted on Synapse should make it into the data catalog: https://www.synapse.org/DataCatalog:0.  The data catalog is powered by a Synapse table (see PR)

# **Solution:**

1. Addition of DOI -> Synapse Table Code Framework
1. Filter out datasets that already exist in the data catalog

# **Testing:**
NA


src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,473 @@
...
+    conn_obj = get_connection(conn=conn)
+
+    try:
+        logger.info("Connected to Snowflake.")
+
+        dois_df = extract_public_dois(conn=conn_obj)
+        logger.info("Extract Public DOIs completed successfully.")
+
+        datacite_dois = fetch_doi_prefix(prefixes=["10.7303"], state="findable")
+        logger.info("Fetched all Public datacite data")
```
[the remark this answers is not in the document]
thomasyu888: Great suggestion but not going to implement for now

src/dois/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+anthropic>=0.52.0
+pandas<3.0.0
+synapseclient
+```
+
+Install the module and its dependencies:
+
+```bash
+pip install -e src/dois/
```
linglp: this command here didn't work for me. I will have to add a pyproject.toml to make it work.

src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,652 @@
...
+    syn = synapseclient.login()
+    conn_obj = get_connection(conn=conn)
+
+    try:
+        logger.info("Connected to Snowflake.")
+
+        dois_df = extract_public_dois(conn=conn_obj)
+        logger.info("Extract Public DOIs completed successfully.")
+
+        datacite_dois = fetch_doi_prefix(prefixes=["10.7303"], state="findable")
```
linglp: this hard-coded value should probably be turned to a constant

src/dois/README.md
```diff
@@ -0,0 +1,74 @@
...
+```bash
+pip install -e src/datacite/
+pip install -e src/snowflake_utils/
+```
+
+### Environment variables
+
+| Variable | Required | Description |
+|----------|----------|-------------|
+| `ANTHROPIC_API_KEY` | Yes | Anthropic API key. Used to call Claude for description enrichment. |
```
linglp: I think there need to be instructions created on getting `ANTHROPIC_API_KEY`

src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,652 @@
...
+        dois_df = merge_doi_metadata(synapse_df, datacite_df)
+
+        dois_df = enrich_with_claude(dois_df, syn)
+        logger.info("Claude enrichment completed.")
+
+        existing_ids = get_existing_ids(syn, "syn61609402")
+        dois_df = dois_df[~dois_df["id"].astype(int).isin(existing_ids)]
+        logger.info(f"Filtered to {len(dois_df)} new datasets not yet in syn61609402.")
+
+        dois_df.to_csv("dois_with_metadata.tsv", index=False, sep="\t")
```
linglp: Do you think if you could commit this `dois_with_metadata.tsv` file? And then I could extract the description there and then create a separate JSON file and change `enrich_with_claude` so that we don't have to call claude to re-enrich the data.

linglp: If we want to avoid writing to a local file, we could also write the result to a separate synapse table? And then we could still avoid calling claude for the same data

src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,652 @@
...
+    """Fetch existing IDs from a Synapse table.
+
+    Args:
+        syn: Synapse client instance.
+        table_id: Synapse table ID to query.
+
+    Returns:
+        Set of existing id values.
+    """
+    results = syn.tableQuery(f"SELECT id FROM {table_id}")
```
linglp: avoid using `tableQuery`, and use Table from OOP instead

dags/src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,707 @@
...
+    """
+    with conn.cursor() as cs:
+        cs.execute(query)
+        dois_df = cs.fetch_pandas_all()
+        dois_df.drop_duplicates(inplace=True)
+        del dois_df["TOTAL_SIZE_BYTES_RAW"]
+    return dois_df
+
+
+def transform_synapse_dois(dois_df: pd.DataFrame) -> pd.DataFrame:
```
linglp, line 326: transform_synapse_dois is a good candidate to test since it's pure logic with no external dependencies. Should we create a directory for testing?

dags/doi-to-datacatalog-dag.py
```diff
@@ -0,0 +1,184 @@
...
+    "catchup": False,
+    "default_args": {
+        "retries": 1,
+    },
+    "tags": ["snowflake", "synapse", "data-catalog"],
+    "params": dag_params,
+}
+
+
+def _tmp_path(prefix: str, ts_nodash: str) -> str:
```
linglp, line 60: Nit: add docstring to this function.

dags/doi-to-datacatalog-dag.py
```diff
@@ -0,0 +1,184 @@
...
+        try:
+            dois_df = extract_public_dois(conn=conn)
+        finally:
+            conn.close()
+
+        synapse_df = transform_synapse_dois(dois_df)
+        logger.info("Extracted %d public DOIs from Snowflake.", len(synapse_df))
+
+        path = _tmp_path("synapse", context["ts_nodash"])
+        synapse_df.to_pickle(path)
```
linglp, line 83: I am curious if we could do this in production? I think maybe this is related to how we set up Kubernetes to run airflow tasks? Based on the code [here](https://github.com/Sage-Bionetworks-Workflows/eks-stack/blob/main/modules/apache-airflow/templates/values.yaml#L270), we are using celery executor with 2 replicas. Based on my understanding, it seems like worker A could be running `extract_synapse_dois`, and worker B could be running the `merge` step which is later in the code. If two works are running the tasks separately, worker B can't access the temporary files created by worker A. I am curious if this understanding is correct, @BryanFauble

dags/src/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Retrieving the Anthropic API Key
+
+Follow instructions [here](https://platform.claude.com/docs/en/get-started) to retrieve an Anthropic API key.  `AnthropicHook` resolves the API key in two ways, tried in order:
+
+1. **Airflow connection (production)** — connections are backed by AWS Secrets Manager. Add the API key as a secret under the path Airflow expects for the given `airflow/connections/conn_id`. The `password` field of the connection should be set to the API key value.
+
+2. **Environment variable (local/fallback)** — if the Airflow connection can't be resolved (e.g. running locally outside Airflow), the hook falls back to the `ANTHROPIC_API_KEY` environment variable.
+
+### Setting up locally
```
linglp, line 25: it's easy to set it in one shell session and run the DAG in another, causing a confusing OSError even though echo `$ANTHROPIC_API_KEY` appears to work. Suggest adding a note like:

Verify the key is set in the same shell session you'll run the DAG from:
export ANTHROPIC_API_KEY="sk-ant-..."
python -c "import os; print(os.environ.get('ANTHROPIC_API_KEY'))"  # should print your key

dags/src/README.md
```diff
@@ -0,0 +1,29 @@
...
+
```
linglp, line 26: Imagine that we will be running this in production, who's `ANTHROPIC_API_KEY` that we should be using? And we should probably change the rate limit of the `ANTHROPIC_API_KEY` because I got error: 
```
 File "/Users/lpeng/code/orca-recipes/venv/lib/python3.10/site-packages/anthropic/_base_client.py", line 1373, in post
    return cast(ResponseT, self.request(cast_to, opts, stream=stream, stream_cls=stream_cls))
  File "/Users/lpeng/code/orca-recipes/venv/lib/python3.10/site-packages/anthropic/_base_client.py", line 1146, in request
    raise self._make_status_error_from_response(err.response) from None
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'Your credit balance is too low to access the Anthropic API. Please go to Plans & Billing to upgrade or purchase credits.'}, 'request_id': 'req_011Cahx7pHqq2nWm6Bh5o3CW'}
T
```

dags/doi-to-datacatalog-dag.py
```diff
@@ -0,0 +1,184 @@
...
+    submit_enrichment_batch,
+    submit_test_filter_batch,
+    transform_datacite_dois,
+    transform_synapse_dois,
+)
+from src.synapse_hook import SynapseHook
+
+logger = logging.getLogger(__name__)
+
+dag_params = {
```
linglp, line 36: I think we could add two flags here: 
* submit_batch and submit_test_filter return "" immediately if `skip_enrichment=True` 
* collect_and_load skips the upsert if `dry_run=True`

dags/doi-to-datacatalog-dag.py
```diff
@@ -0,0 +1,184 @@
...
+        logger.info("Loaded %d rows into %s.", n_loaded, context["params"]["data_catalog_table_id"])
+
+    # extract_synapse_dois and fetch_datacite_dois run in parallel
+    synapse_path = extract_synapse_dois()
+    datacite_path = fetch_datacite_dois()
+    merged_path = merge(synapse_path, datacite_path)
+
+    # submit_batch and submit_test_filter run in parallel after merge
+    batch_id = submit_batch(merged_path)
+    test_filter_batch_id = submit_test_filter(merged_path)
```
linglp, line 172: Currently, these two steps are running in parallel. If we have a lot of test projects, will it make sense to first filter out all the test projects first before calling `submit_batch`? 

Also, will it make sense to just do everything in one batch? How about asking Claude to return {"is_test": true/false, "description": "...", "keywords": [...]} in one call?

dags/doi-to-datacatalog-dag.py
```diff
@@ -0,0 +1,184 @@
...
+        merged_df = pd.read_pickle(merged_path)
+
+        anthropic_hook = AnthropicHook(context["params"]["anthropic_conn_id"])
+        enriched_df = collect_enrichment_results(merged_df, batch_id, anthropic_hook.client)
+        logger.info("Enrichment results collected.")
+
+        filtered_df = collect_test_filter_results(enriched_df, test_filter_batch_id, anthropic_hook.client)
+        n_test = filtered_df["anthropic_is_test"].sum()
+        logger.info("Test filter: %d test/demo projects removed, %d meaningful remaining.", n_test, len(filtered_df) - n_test)
+        # meaningful_df = filtered_df[~filtered_df[""]].drop(columns=["anthropic_is_test"])
```
linglp: Nit: this should be removed?

dags/src/datacite/datacite.py
```diff
@@ -0,0 +1,380 @@
...
+    
+    logger.debug(
+        f"Fetching page {page_number} with page_size={page_size}, "
+        f"prefixes={prefixes}, state={state}"
+    )
+    params = _build_query_params(prefixes, state, page_size, page_number, detail)
+    resp = _make_request_with_retry(session, DATACITE_API, params)
+    result = resp.json()
+    data_count = len(result.get("data", []))
+    logger.info(f"Fetched page {page_number}: {data_count} DOI records")
```
linglp: Nit: I saw that we are using a combination of f string and  %s formatting. Maybe just stick with %s formatting in all cases?

dags/src/anthropic_hook.py
```diff
@@ -0,0 +1,33 @@
...
+        if self._client is None:
+            self._client = anthropic.Anthropic(api_key=self._resolve_api_key())
+        return self._client
+
+    def _resolve_api_key(self) -> str:
+        try:
+            from airflow.hooks.base import BaseHook
+
+            return BaseHook.get_connection(self.conn_id).password
+            # return os.environ.get("ANTHROPIC_API_KEY")
```
linglp: Nit: also remove this comment?

dags/src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,707 @@
...
+    merged["includedInDataCatalog"] = False
+    merged["link"] = float("nan")
+    merged["croissant"] = float("nan")
+    merged["appId"] = float("nan")
+    merged["license"] = float("nan")
+    merged.drop(columns=["doi"], inplace=True)
+    return merged
+
+
+def get_existing_ids(syn: synapseclient.Synapse, table_id: str) -> set:
```
linglp: syn is never used in `get_existing_ids`

dags/src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,707 @@
...
+
+    Returns:
+        Number of rows loaded.
+    """
+    existing_ids = get_existing_ids(syn, existing_datasets_table_id)
+    filtered_df = df[~df["id"].astype(int).isin(existing_ids)]
+    logger.info("Filtered to %d new datasets not yet in %s.", len(filtered_df), existing_datasets_table_id)
+
+    table = Table(id=data_catalog_table_id)
+    table.delete_rows(query=f"SELECT * FROM {data_catalog_table_id}")
```
linglp, line 730: Do we have to delete the whole table before inserting? Is it possible to just do "upsert only" without wiping out the whole table?

dags/src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,707 @@
...
+
+
+def transform_datacite_dois(datacite_dois: Iterable[Dict]) -> pd.DataFrame:
+    """Flatten raw DataCite API objects into a DataFrame, deduplicated on 'doi'."""
+    df = pd.DataFrame([i["attributes"] for i in datacite_dois])
+    df.drop_duplicates("doi", inplace=True)
+    return df
+
+
+def merge_doi_metadata(
```
linglp, line 644: I think we should also add a test for this function

dags/src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,707 @@
...
+                "messages": [{"role": "user", "content": prompt}],
+            },
+        ))
+
+    batch = anthropic_client.messages.batches.create(requests=requests)
+    logger.info("Submitted batch %s with %d requests.", batch.id, len(requests))
+    return batch.id
+
+
+def collect_enrichment_results(
```
linglp, line 464: Also add a test for this function?

dags/src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,707 @@
...
+    if isinstance(val, float) and pd.isna(val):
+        return True
+    if isinstance(val, str):
+        return val.strip() == ""
+    if isinstance(val, list):
+        return len(val) == 0
+    return False
+
+
+def _build_prompt(
```
linglp, line 348: Nit: add docstring for this function? and also add test?

linglp (commented):
Thanks for your hard work, Tom!                                                                                                                                                                        

My main concern is that In Kubernetes, each Airflow task runs in its own pod with ephemeral local storage that isn't shared across pods. So when `extract_synapse_dois` writes `/tmp/doi_synapse_<ts>.pkl` and merge tries to read it, the file won't exist if merge is running on a different pod. Should we write the output to a synapse table instead of storing a pickle file? 

Also, I think it will be really helpful if there could be a "dry_run" flag to skip upsert or to skip enrichment. 

Before this can go to production, we need an API key tied to a shared DPE/team Anthropic account rather than a personal one. Do we have one set up, or do we need to create it?

dags/src/dois/doi_to_datacatalog.py
linglp: This is due to: https://sagebionetworks.jira.com/browse/SYNPY-1835. See ticket: https://sagebionetworks.jira.com/browse/DPE-1665

.github/workflows/validate.yml
```diff
@@ -27,6 +27,10 @@ jobs:
       - name: Install airflow deps
         run: |
           pip install -r requirements-airflow.txt
+
+      - name: Install DAG-specific packages
```
jaymedina, line 31: What's the purpose of this package?

cc @linglp @thomasyu888

dags/src/datacite/datacite.py
thomasyu888, line 1: This is migrated code

dags/src/datacite/datacite.py
```diff
@@ -0,0 +1,378 @@
...
+    detail: bool,
+) -> Dict[str, Any]:
+    """Build query parameters for DataCite API request.
+
+    Args:
+        prefixes: List of DataCite DOI prefixes to query (e.g., ["10.7303"]).
+        state: DOI state to filter by ("findable", "registered", or "draft").
+        page_size: Number of results to return per page.
+        page_number: Zero-based page number for pagination.
+        detail: Whether to include detailed metadata in the response.
```
jaymedina, line 70: nit: add "can be true/false"

dags/src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,732 @@
...
+        node_revision.description,
+        node_latest.annotations:annotations                     as annotations
+    from synapse_rds_snapshot.prod_583.doi
+    left join synapse_data_warehouse.synapse.node_latest
+        on doi.object_id = node_latest.id
+    join synapse_rds_snapshot.prod_583.node_revision
+        on node_latest.id = node_revision.owner_node_id
+        and (
+            doi.object_version = -1
+            or node_revision.number = doi.object_version
```
jaymedina, line 48: what happens when there are multiple object versions including -1

dags/src/dois/doi_to_datacatalog.py
```diff
@@ -0,0 +1,732 @@
...
+        descendant_id                                           as file_node_id,
+        file_handle_id
+    from folder_descendants
+    where node_type = 'file'
+),
+
+-- ============================================================
+-- 3. DATASET: explode scope_ids -> look up file_handle_id
+-- ============================================================
+dataset_scope_ids as (
```
jaymedina, line 94: This CTE is not behaving as expected:

Image

Example used: https://www.synapse.org/Synapse:syn70283644.2/datasets/

PR #148, opened 2026-04-09
[POC] Airflow DAG for ingesting RDS snapshots to Snowflake

jaymedina:
# **Problem:**

DPE is working on a SPIKE ticket to establish the best approach for automating the ingestion of Synapse Platform RDS snapshots on S3 -> Snowflake

This is an accompaniment to the design doc [here](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4634968068/DRAFT+SNOW-397+Automated+ingestion+of+RDS+snapshot+parquet+files+from+AWS+S3+--+Snowflake).

DPE is exploring 2 options:
1. Orchestrate data ingestion using Snowflake tasks (a completely Snowflake-native approach)
2. Orchestrate data ingestion using an Apache Airflow DAG

# **Solution:**

This PR is a POC for the Airflow DAG approach: 
image

# **Testing:**
n/a

### DIFF
--- a/dags/poc-snapshots-to-snowflake-dag.py
+++ b/dags/poc-snapshots-to-snowflake-dag.py
@@ -11,14 +11,16 @@ Dependencies:
     pip install apache-airflow-providers-slack
 """
 import os
-import pathlib
+import requests
 import yaml
 
 from datetime import datetime, timedelta
 
+from slack_sdk import WebClient
+
 from airflow import DAG
+from airflow.models import Variable
 from airflow.operators.python import PythonOperator, BranchPythonOperator
-from airflow.providers.slack.operators.slack import SlackAPIPostOperator
 from airflow.providers.snowflake.operators.snowflake import SnowflakeOperator
 from airflow.utils.task_group import TaskGroup
 
@@ -43,14 +45,13 @@ SNOWFLAKE_SCHEMA = "RDS_LANDING"
 SNOWFLAKE_S3_STAGE = f"{SNOWFLAKE_DATABASE}.{SNOWFLAKE_SCHEMA}.RDS_SNAPSHOTS_STAGE"
 SNOWFLAKE_CONN_ID = "SNOWFLAKE_DEVELOPER_SERVICE_RAW_CONN"
 AWS_CONN_ID = "AWS_TOWER_PROD_S3_CONN"
-SLACK_CONN_ID = "SLACK_DATA_ENG_CONN"
-SLACK_CHANNEL = "#data-eng-alerts"
-S3_PREFIX = "rds_snapshots/"
+SLACK_CHANNEL = "#snowflake_ingest_updates"
+SLACK_TOKEN_VAR = "SLACK_DPE_TEAM_BOT_TOKEN"
 
 # ---------------------------------------------------------------------------
 # Load record types from YAML
 #
-# record_types.yaml lives in the same directory as this DAG file.
+# record_types.yaml is fetched from GitHub at DAG parse time.
 # Each entry is a record type name matching the S3 subdirectory convention,
 # e.g. "ACCESS_APPROVAL" maps to s3://.../rds_snapshots/.../ACCESS_APPROVAL/
 #
@@ -64,9 +65,21 @@ S3_PREFIX = "rds_snapshots/"
 #     ... (157 total)
 # ---------------------------------------------------------------------------
 
-_yaml_path = pathlib.Path(__file__).parent / "record_types.yaml"
-with _yaml_path.open() as f:
-    _yaml = yaml.safe_load(f)
+RECORD_TYPES_URL = "https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/main/dags/record_types.yaml"
+
+
+def load_record_types(url=RECORD_TYPES_URL):
+    """Load record types from a raw GitHub URL."""
+    response = requests.get(url, timeout=10)
+    if not response.ok:
+        raise RuntimeError(
+            f"Failed to fetch record types... "
+            f"Status code: {response.status_code}, reason: {response.reason}"
+        )
+    return yaml.safe_load(response.text)
+
+
+_yaml = load_record_types()
 
 RECORD_TYPES: list[str] = _yaml["record_types"]
 
@@ -76,6 +89,34 @@ RECORD_TYPES: list[str] = _yaml["record_types"]
 
 def _fully_qualified(record_type: str) -> str:
     return f"{SNOWFLAKE_DATABASE}.{SNOWFLAKE_SCHEMA}.{record_type}"
+
+
+def _notify_slack_failure(context):
+    client = WebClient(token=Variable.get(SLACK_TOKEN_VAR))
+    client.chat_postMessage(
+        channel=SLACK_CHANNEL,
+        text=(
+            f":red_circle: *Ingest failure* in `{context['dag'].dag_id}`\n"
+            f"Task: `{context['task_instance'].task_id}`\n"
+            f"Run: `{context['run_id']}`\n"
+            f"Log: {context['task_instance'].log_url}"
+        ),
+    )
+
+
+def _notify_new_tables(**context):
+    missing = context["ti"].xcom_pull(
+        task_ids="check_missing_tables", key="missing_tables"
+    )
+    env = context["ti"].xcom_pull(task_ids=None, key="RDS_SNAPSHOTS_STACK") or ENVIRONMENT
+    client = WebClient(token=Variable.get(SLACK_TOKEN_VAR))
+    client.chat_postMessage(
+        channel=SLACK_CHANNEL,
+        text=(
+            f":new: *New rds_landing tables created* (`{env}` env)\n"
+            + ", ".join(missing)
+        ),
+    )
 
 
 def _check_missing_tables(**context) -> list[str]:
@@ -124,6 +165,7 @@ default_args = {
     "retries": 2,
     "retry_delay": timedelta(minutes=5),
     "email_on_failure": True,
+    "on_failure_callback": _notify_slack_failure,
 }
 
 # ---------------------------------------------------------------------------
@@ -169,16 +211,9 @@ with DAG(
             ],
         )
 
-        notify_slack = SlackAPIPostOperator(
+        notify_slack = PythonOperator(
             task_id="notify_slack",
-            slack_conn_id=SLACK_CONN_ID,
-            channel=SLACK_CHANNEL,
-            # The message is rendered at runtime from XCom so it lists the
-            # actual table names that were created in this run.
-            text=(
-                ":new: *New rds_landing tables created* (`{{ var.value.get('RDS_SNAPSHOTS_STACK', 'dev') }}` env)\n"
-                "{{ task_instance.xcom_pull(task_ids='check_missing_tables', key='missing_tables') | join(', ') }}"
-            ),
+            python_callable=_notify_new_tables,
         )
 
         create_tables >> notify_slack
@@ -216,7 +251,7 @@ with DAG(
                         FROM @{SNOWFLAKE_S3_STAGE}
                     )
                     FILE_FORMAT = (TYPE = PARQUET)
-                    PATTERN     = '.*/{record_type}/.*\\\\.parquet'
+                    PATTERN     = '.*\\.{record_type}/.*\\.parquet'
                     ON_ERROR    = 'CONTINUE';
                 """,
             )
--- a/dags/poc-snapshots-to-snowflake-dag.py
+++ b/dags/poc-snapshots-to-snowflake-dag.py
@@ -34,7 +34,7 @@ if ENVIRONMENT.lower() == "prod":
     DAG_SUFFIX = "_prod"
 elif ENVIRONMENT.lower() == "dev":
     S3_BUCKET = "synapse-snowflake-rds-snapshots-dev"
-    SNOWFLAKE_DATABASE = "SYNAPSE_DATA_WAREHOUSE_DEV"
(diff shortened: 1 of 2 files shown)


dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,288 @@
...
+        python_callable=_branch_on_missing,
+    )
+
+    # -------------------------------------------------------------------------
+    # 2. Handle new record types: create table + notify Slack.
+    #    Grouped so the branch target is a single task_id.
+    # -------------------------------------------------------------------------
+    with TaskGroup("handle_new_tables") as handle_new_tables:
+
+        create_tables = SnowflakeOperator(
```
thomasyu888: For the purposes of this POC, I'm not going to be focussing on this and the copy into function below because the upstream tasks are to be solidified at a future date.

dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,288 @@
...
+                snowflake_conn_id=SNOWFLAKE_CONN_ID,
+                sql=f"""
+                    COPY INTO {_fully_qualified(record_type)} (raw)
+                    FROM (
+                        SELECT $1
+                        FROM @{SNOWFLAKE_S3_STAGE}
+                    )
+                    FILE_FORMAT = (TYPE = PARQUET)
+                    PATTERN     = '.*\\.{record_type}/.*\\.parquet'
+                    ON_ERROR    = 'CONTINUE';
```
thomasyu888: This is a risk that we should probably notify slack if there is an error

dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,217 @@
...
+
+ENVIRONMENT = os.getenv("RDS_SNAPSHOTS_STACK", "dev")
+
+if ENVIRONMENT.lower() == "prod":
+    S3_BUCKET = "synapse-snowflake-rds-snapshots-prod"
+    SNOWFLAKE_DATABASE = "SYNAPSE_DATA_WAREHOUSE"
+    DAG_SUFFIX = "_prod"
+elif ENVIRONMENT.lower() == "dev":
+    S3_BUCKET = "synapse-snowflake-rds-snapshots-dev"
+    SNOWFLAKE_DATABASE = "SYNAPSE_DATA_WAREHOUSE_DEV_ADD_CLAUDE_MD"
```
philerooski, line 37: To be updated.

dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,217 @@
...
+            cs.execute(f"ALTER STAGE IF EXISTS {SNOWFLAKE_S3_STAGE} REFRESH;")
+        finally:
+            cs.close()
+
+    @task
+    def create_tables() -> None:
+        # Creates all rds_landing tables if they don't already exist.
+        # Runs once before the COPY INTO fan-out so every target table is
+        # guaranteed to exist when the parallel copy tasks start.
+        hook = SnowflakeHook(snowflake_conn_id=SNOWFLAKE_CONN_ID)
```
philerooski, line 162: How will you create tables using the `developer_service` credentials (assuming it only has access to DATA_ENGINEER)? Might need to update this to be environment dependent.

dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,217 @@
...
+#   record_types:
+#     - ACCESS_APPROVAL
+#     - TEAM
+#     - RESEARCH_PROJECT
+#     - FILE_DOWNLOAD
+#     - NODE_SNAPSHOT
+#     ... (157 total)
+# ---------------------------------------------------------------------------
+
+RECORD_TYPES_URL = "https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/poc-snapshots-to-snowflake/dags/record_types.yaml"
```
philerooski, line 66: To be updated.

dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,217 @@
...
+
+from airflow.decorators import dag, task
+from airflow.models import Variable
+from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
+
+# ---------------------------------------------------------------------------
+# Config
+# ---------------------------------------------------------------------------
+
+ENVIRONMENT = os.getenv("RDS_SNAPSHOTS_STACK", "dev")
```
philerooski, line 29: nit: "stack" is a bit confusing for this variable. You could equally call the Synapse stack number, e.g., prod-576 etc. the `RDS_SNAPSHOTS_STACK`. Perhaps `RDS_SNAPSHOTS_ENVIRONMENT` would be less ambiguous.

dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,217 @@
...
+    #     >> [copy_access_approval, copy_team, ...]   (parallel)
+    #
+    #   COPY INTO is the sole dedup mechanism — Snowflake's load history
+    #   (path + ETag, 64-day window) skips unchanged files and re-loads
+    #   corrected ones automatically. No file sensing needed.
+    # -------------------------------------------------------------------------
+    started = notify_started()
+    stage = refresh_stage()
+    tables = create_tables()
+    copy_tasks = [
```
philerooski, line 208: nit: maybe a variable name like `copy_into_all_record_types` would be more explicit as to what this task list is doing.

also nit: It would be more consistent with other code and documentation to call `ACCESS_APPROVAL`, `RESEARCH_PROJECT`, etc. `data_types`, rather than `record_types`. So amend my previous nit to `copy_into_all_data_types`.

dags/record_types.yaml
```diff
@@ -0,0 +1,3 @@
+record_types:
```
philerooski, line 1: `data_types` is the preferred nomenclature.

dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,217 @@
...
+            text=f":arrow_forward: *s3_to_snowflake_poc{DAG_SUFFIX}* started — refreshing stage and loading parquet snapshots into `{SNOWFLAKE_DATABASE}.{SNOWFLAKE_SCHEMA}`.",
+        )
+
+    @task
+    def notify_finished() -> None:
+        client = WebClient(token=Variable.get(SLACK_TOKEN_VAR))
+        client.chat_postMessage(
+            channel=SLACK_CHANNEL,
+            text=f":white_check_mark: *s3_to_snowflake_poc{DAG_SUFFIX}* finished — all COPY INTO tasks completed successfully.",
+        )
```
philerooski, line 145: Won't messages generated by `_notify_slack_failure` get buried by start/finish messages? Sure, we could read the channel every day, but when 99% of the time all we see is start/finish, we won't. I recommend a different channel for actual errors, or maybe if we @ our usernames and configure the Slack notification settings appropriately we can better differentiate between the signal and the noise, similar to how we can set log levels in Python.

dags/poc-snapshots-to-snowflake-dag.py
```diff
@@ -0,0 +1,217 @@
...
+            """)
+        finally:
+            cs.close()
+
+    # -------------------------------------------------------------------------
+    # Pipeline order:
+    #
+    #   refresh_stage
+    #     >> create_tables
+    #     >> [copy_access_approval, copy_team, ...]   (parallel)
```
philerooski, line 199: We have 150+ data types. I'm dubious whether our Snowflake account could (or should) run 150 COPY INTO queries in parallel. We might even get rate limited on the AWS side.

philerooski (commented):
My impression is that the DAG is trying to do three independent things at once: Slack notifications, Snowflake object configuration, and Snowflake data loading. My concern with having all of these processes in a single DAG is that things are going to get complicated and it's all going to be happening in a single Python script.

*Monitoring and Reporting* (Slack notifications): Monitoring for warnings and failures across 150+ data types has the potential to be messy. We're interested in monitoring at two different levels:

1. The orchestration level: Failures in the DAG itself
2. The platform level: Failures within the external systems we're interfacing with. This mostly means failures in Snowflake, but perhaps failures in Slack is something we might want to consider, too.

This PR implements a Slack reporting mechanism for (1) but there's nothing monitoring (2). It may be its own suite of business logic just to parse the `COPY_HISTORY` table for each data type in Snowflake so that we're aware of data loading failures. With the additional complexity, I think it's justified to have our monitoring and reporting logic at the platform level in its own indendent DAG. Each task would be a different QA check to ensure that the load was successful (and potentially also we can check that the data quality is good, via a dbt test suite).

*Schema Management* (Snowflake object configuration): This entails everything we need to do in Snowflake to ensure that our data loading is successful. This PR simply creates a table from a value in a config file if that table doesn't yet exist, but we'll also need logic for managing and updating table schemas. Because the current behavior simply loads all data into a VARIANT column, we punt the schema management issue downstream, where it may be harder to manage (this is the question we were debating around whether we establish the schema when we ingest the data or when we transform the data into `DBT_LANDING`).

Whether we want to manage this process in Airflow is debatable. Because of the complexities of responding to schema changes, we'll never be able to fully automate the management of our source tables and their schema. We should explore whether we can avoid building Airflow infrastructure to solve this problem if we assume that schema's won't change, or will only contain additive changes. Can we sync the state of our tables with a configuration file (maybe even a DBT source file) as an extension to our existing Github actions deployment process for Snowflake objects?

*Data Loading*: This is the simplest process: it's just a `COPY INTO` statement. Snowflake manages the monitoring by logging to the `COPY_HISTORY` table, so there's nothing for us to do but provide the right parameters. Is this something that we can consolidate into Snowflake tasks, and should we? A benefit would be that it allows us to load Synapse data from RDS using the same infrastructure configuration as we load Synapse data from the Glue pipeline. Another benefit is that, we (and other non-developers) can see the task defintion, task graph, and run history within Snowflake, rather than having all of that managed in a different external system (Airflow) that's only accessible to developers. A downside is that we lose the orchestration flexibility of Airflow, but do we need it in this case? What about tasks downstream from data loading?

philerooski:
@jaymedina 

Just to clarify what I feel like the priority of our next steps are before we bite off more than we can chew:

1. Data loading: This is the lowest bar, and it doesn't matter if we have any of the other stuff if the data isn't actually getting through on a daily basis. We want a solution that's simple, relatively robust, and gives us room to extend the implementation further to complement some of the lower priority action items.
2. Schema management: We're going to need to establish the schemas before we can do any data modeling, and if there's no data modeling, there's no data flowing downstream. Let's start with the assumption that schemas won't change, or will only have non-breaking changes, so that we can arrive at a solution that gets data flowing without having to first solve the hardest parts of the complexities of schema migrations.
3. Logging/Reporting: Let's focus on developer visibility of issues. Slack reporting is nice for non-developers, but I don't want us spending time devising or debugging the optimal front-end solution when what's really important is that data loading issues are visible to developers.

If you want to talk more about what this looks like, let me know.

jaymedina:
Hey @philerooski, thanks for the detailed writeup! These are exactly the right questions to be asking, but I think       
  they're scoped a bit beyond what this PR is meant to explore. This DAG was written before I split the pipeline    
  into separate workflows ([warehouse setup vs. regular ingestion](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4634968068/DRAFT+SNOW-397+Automated+ingestion+of+RDS+snapshot+parquet+files+from+AWS+S3+--+Snowflake#%F0%9F%94%80-Two-Phase-Architecture-Constraint)), so some of the overlap you're seeing reflects    
  that earlier state.                                                                                               
                                                                                                                    
  I'll be introducing the technology tradeoffs more formally in the design doc, where I'm planning to lay out three 
  initial options (which we can iterate on together):                                                                                                          
   
  - Option 1: GitHub Actions for Phase 1 (warehouse setup), Snowflake-native tasks for Phase 2 (ingestion)          
  - Option 2: GitHub Actions for Phase 1, Airflow for Phase 2
  - Option 3: Airflow for both phases (eventually)                                                                  
                                                                                                                    
  That feels like the right venue to hash out which combination makes the most sense, or explore other options.                               
                                                                                                                    
  One thing I want to flag: I don't think we should design around schemas being additive-only. 
  If we bake that assumption in and it breaks, we've introduced tech debt — and it's avoidable tech debt, since the
  Airflow-based approach handles schema evolution without that constraint. Why add the risk when we have a path that
   doesn't require it?

  Happy to continue the conversation in the design doc once I've got the options written up.     

_Note: This comment has been generated with AI assistance and reviewed by me._

jaymedina:
> Just to clarify what I feel like the priority of our next steps are before we bite off more than we can chew:

Excellent, and yes I agree with these priorities. I am thinking in the mental model of:

**Schema Management -> Phase 1
Data Loading & Visibility -> Phase 2**

And when I map your order of priorities to my mental model, I mostly* agree with this organization. Phase 2 is what we will automate first and implement this sprint. Phase 1 we can think more about later. But let's solidify an architecture for both Phase 1 and 2 that we can expand on throughout the year.

*mostly = I disagree with prioritizing Logging/Reporting as last. Developer and Analyst visibility IMO should come with Data Loading. A lot of our user questions revolve around "_When has this table been updated?_", "_How fresh is this data?_" because a lot of analysis requires disclosing this type of information + developers need visibility for obvious reasons, but we already have that with Snowsight.

philerooski:
Thanks for the clarification. I was initially concerned that you we're trying to manage everything in a single Airflow DAG when we already have existing standards around how we do object configuration (Github actions CI/CD) and data loading (Snowflake tasks). That alongside all the logging/reporting logic that we would like to implement is a pretty complex workflow that would exist entirely outside our existing infrastructure.

We can discuss further in the design doc.

PR #150, opened 2026-04-17
[DPE-1789] Support local Airflow development + use airflow slack connector

thomasyu888:
# **Problem:**

- Support local airflow development for some DAGS
- slack uses a separate connection method
- deprecate a dag
- Delete a dag and just wait for the entire file to be merged in via #130 to clean dependencies
- dependency cleanup

# **Solution:**

- Iteratively add to `requirements-local-airflow.txt` with pinned Airflow provider dependencies for local installation. 
- Use airflow maintained slack connector (Added `airflow/connections/DPE_SLACK_BOT_CONN` secret )

# **Testing:**

- Ran `dag.test()` locally against `synapse-by-the-numbers-dag.py` using a local Airflow SQLite database and AWS Secrets Manager connections via the `dpe-prod-admin` profile.
- Verified Snowflake queries execute correctly and results push to the target Synapse table.
- Verified slack hook works locally via one of the dags


README.md
```diff
@@ -116,6 +116,50 @@ The username and password will be "airflow".
...
+```
+
+This creates the local SQLite metadata database (including the `task_instance` table) that Airflow needs to run `dag.test()`.
+
+#### 3. Configure environment variables
+
+Set the following exports (e.g., in your shell profile) so Airflow can resolve connections and variables from AWS Secrets Manager and deserialize custom dataclasses passed between tasks:
+
+```console
+export AIRFLOW__SECRETS__BACKEND=airflow.providers.amazon.aws.secrets.secrets_manager.SecretsManagerBackend
```
thomasyu888, line 188: There is a airflow.cfg, maybe load that into there

dags/synapse-by-the-numbers-dag.py
```diff
@@ -159,4 +159,7 @@ def push_results_to_synapse_table(metrics: List[DownloadMetric], **context) -> N
     top_downloads >> push_to_synapse_table
 
 
-synapse_by_the_numbers_past_month()
+dag = synapse_by_the_numbers_past_month()
+
+if __name__ == "__main__":
```
thomasyu888, line 165: First execution on production airflow cluster, will always run the DAG

requirements-local-airflow.txt
```diff
@@ -0,0 +1,14 @@
+-c https://raw.githubusercontent.com/apache/airflow/constraints-2.10.5/constraints-3.10.txt
+apache-airflow-providers-amazon ~=9.2.0
+apache-airflow-providers-celery >=3.10.0
+apache-airflow-providers-snowflake ~=6.0.0
+snowflake-connector-python ~=3.13.2
+# py-orca[all] == 1.5.2
+# synapseclient[pandas]~=4.0
+# pandas <3.0.0
+slack-sdk >=3.27
```
thomasyu888: There's an opportunity to clean up this requirements.txt file as we've added a lot of requirements to it over the years, but it's unclear to me if these are actually required

requirements-local-airflow.txt
```diff
@@ -0,0 +1,14 @@
+-c https://raw.githubusercontent.com/apache/airflow/constraints-2.10.5/constraints-3.10.txt
```
linglp, line 1: How can we quickly install dependencies for a specific DAG? I am assuming that the idea is to have all the shared dependencies here and manage the dependencies more specific to a given dag in a separate place? Could you clarify that in the ReadMe?

thomasyu888: This is a great question. The Celery executor setup for Airflow currently has all DAGs sharing a common set of requirements, so adding new packages needs to be done carefully.

You’ll notice in the ticket README that synapseclient has to stay separate from this requirements.txt because of a urllib3 version mismatch. That said, the DAG environment is still able to install synapseclient.

Airflow is notoriously nuanced when it comes to dependency management. The constraints file here is part of that complexity, and it is not always obvious which dependency combinations will cause Airflow to fail. In this case, it appears to tolerate the fact that synapseclient depends on a newer version of urllib3.

Separately, for DAG-specific dependencies, I think Airflow may support something like a virtualenv task. I remember reviewing Rixing's PR that had this and it could be worth exploring the tradeoffs around this approach and the impact on processing time.

Ill update the readme

README.md
```diff
@@ -116,6 +116,50 @@ The username and password will be "airflow".
...
+#### 1. Install dependencies
+
+```console
+pip install -r requirements-local-airflow.txt
+```
+
+Then install `synapseclient` separately. This must be done after the above to avoid a `urllib3` version conflict introduced by the Airflow constraints file:
+
+```console
+pip install "synapseclient[pandas]"
```
rxu17: Not sure why I'm running into the [setuptools issue here](https://github.com/pypa/setuptools/issues/5174) when synapseclient imports opentelemetry but might also need a command here to lock down the `setuptools` versions so it avoids that (need to investigate further)

README.md
```diff
@@ -116,6 +116,50 @@ The username and password will be "airflow".
...
+
+Set the following exports (e.g., in your shell profile) so Airflow can resolve connections and variables from AWS Secrets Manager and deserialize custom dataclasses passed between tasks:
+
+```console
+export AIRFLOW__SECRETS__BACKEND=airflow.providers.amazon.aws.secrets.secrets_manager.SecretsManagerBackend
+export AIRFLOW__SECRETS__BACKEND_KWARGS='{"connections_prefix": "airflow/connections", "variables_prefix": "airflow/variables", "profile_name": "<your-aws-profile>"}'
+export AIRFLOW__CORE__ALLOWED_DESERIALIZATION_CLASSES="airflow.* astro.* __main__.*"
+```
+
+You'll need to be authenticated to AWS with a profile (use `aws sso login --profile <<your-aws-profile>`) that has read access to the Secrets Manager secrets under `airflow/connections/` and `airflow/variables/`.
```
rxu17: I would also move this to before the exporting of the airflow variables since it has to be done before

```suggestion
You'll need to configure (use `aws configure sso`) and be authenticated to AWS with a profile (use `aws sso login --profile <<your-aws-profile>>`) that has read access to the Secrets Manager secrets under `airflow/connections/` and `airflow/variables/`. You can use the `dpe-prod` AWS account.
. [See configuration docs for more info.](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html#sso-configure-profile-token-auto-sso). 
```

dags/src/synapse_hook.py
```diff
@@ -0,0 +1,62 @@
+import os
+from typing import Union
+
+import synapseclient
+from synapseclient.models import query, SubmissionStatus
```
linglp: `synapseclient` itself changed its `urllib3` requirement between versions:

`4.10.0`: `urllib3<2,>=1.26.18` (compatible with urllib3<2.0)
`4.11.0+`: `urllib3>=2.6.3` (requires urllib3 2.x)
And py-orca's all extra pins `urllib3<2.0` (see [here](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/main/setup.cfg#L76))

I think we should update the urllib version in Py-orca so that py-orca can also be installed when we do local development. Alternatively, is thte goal to deprecate `py-orca` and move everything to orca-recipes?

thomasyu888: Thanks for reviewing this, I should close or put this in draft mode for a couple of reasons

1. For security, we are going to have to push to airflow 3.0 which is going to ruffle up our dependences
2. Rixing is adding in the testing and src framework here to remove dependency on py-orca in the zenodo PR
3. We can look potentially look to implement the nextflow hook in src eventually to completely remove the py-orca dependency to reduce our maintenance overhead.

README.md
linglp: I experienced similar issues. I think setuptools needs to be pinned down now. See PR: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/167

dags/src/synapse_hook.py
linglp: Thanks for clarifying Tom.

README.md
```diff
@@ -116,6 +116,50 @@ The username and password will be "airflow".
 If you encounter the `nginx bad gateway` errors when navigating to the forwarded port, just wait and refresh a couple of times. Airflow takes a few minutes to become available.
 
 
+### Local Development (without Dev Container)
```
rxu17, line 159: This section may be a better fit in CONTRIBUTING.md given it's development and testing info there: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/CONTRIBUTING.md#testing. There can just be a link to that doc here.

requirements-local-airflow.txt
rxu17: I didn't get to do this for my DAG (we opted out of excel and decided to use csv just to reduce a dependency) but it looks like we could do something like this per dag:

Have a `virtualenv_requirements.py` file or something containing each requirement per DAG:
```python
# virtualenv_requirements.py

VIRTUALENV_REQUIREMENTS = {
    "zenodo_tep_metrics": [
        "requests~=2.31",
        "openpyxl~=3.1",
        "synapseclient[pandas]~=4.0",
    ],
    "nextflow_tower": [
        "requests~=2.31",
        "slack-sdk>=3.27",
    ],
}
```

Then we can do something in our dags like 

```python
from virtualenv_requirements import VIRTUALENV_REQUIREMENTS

@task.virtualenv(
    requirements=VIRTUALENV_REQUIREMENTS["zenodo_tep_metrics"],
)
def export_metrics():
    ...
```

Haven't actually tried this out, but this seems like theoretically it could work

dags/genie/genie-patient-sample-tracking-dag.py
rxu17, line 1: I feel like we can archive this dag?

README.md
rxu17, line 1: I wonder why this is still showing changes? You might have to merge `main` into this branch and deal with some merge conflicts (since previous PRs have modified the README.md)

dags/genie/genie-patient-sample-tracking-dag.py
thomasyu888: I'm happy to archive, and let's align with the genie team first?

requirements-local-airflow.txt
```diff
@@ -1,2 +1,14 @@
 -c https://raw.githubusercontent.com/apache/airflow/constraints-2.10.5/constraints-3.10.txt
 apache-airflow-providers-amazon ~=9.2.0
+# apache-airflow-providers-celery >=3.10.0
+# apache-airflow-providers-snowflake ~=6.0.0
+# snowflake-connector-python ~=3.13.2
+# # py-orca[all] == 1.5.2
```
rxu17: nit: Can we remove these commented out lines since our fix is just to run the shell script under /scripts?

dags/top-public-synapse-projects-from-snowflake.py
```diff
@@ -39,10 +41,11 @@
     #           {"file_view_id": "789012", "group_name": "Group B"}]'
     "fileview_groups": Param(
         '[{"file_view_id": "20446927", "group_name": "HTAN1"},{"file_view_id": "51489960", "group_name": "ELITE"},{"file_view_id": "16858331", "group_name": "NF-OSI"},{"file_view_id": "27210848", "group_name": "MC2"},{"file_view_id": "52794526", "group_name": "GENIE"}]', type="string"),
+    "synapse_results_table": Param("syn53696951", type="string"),
```
rxu17, line 45: nit: Can the comment for this previously global
 variable be added above this line? Same comment for the other DAGs that had this change

rxu17:
I wonder if we can use this PR to test out a POC I did of semi-mimicing the airflow production environment paths (because currently the regular dag import test doesn't catch all import errors due to a pathing mismatch between local airflow dev env and production airflow env) and importing in the DAGs to catch any DAG import errors? Let me quickly add in that PR for everyone to take a look

README.md
```diff
@@ -202,6 +202,7 @@ DAG files with an `if __name__ == "__main__":` block that calls `dag.test()` can
 
 See section [Testing](./CONTRIBUTING.md#testing) in the CONTRIBUTING.md for more info on testing a dag and how you can bypass configuring with AWS.
 
+
```
thomasyu888:
```suggestion
```

dags/top-public-synapse-projects-from-snowflake.py
thomasyu888, line 1: The slack hook was tested locally here

dags/synapse-by-the-numbers-dag.py
```diff
@@ -160,4 +160,7 @@ def push_results_to_synapse_table(metrics: List[DownloadMetric], **context) -> N
...
+    dag.test(run_conf={"synapse_results_table": "syn74496297"})
```
rxu17, line 167: Is this the staging/test table here? Similar question for the other similar dags. If so, nit: to add a comment what table this is or make it into a variable

dags/synapse-datasets-created-7-days.py
```diff
@@ -21,16 +21,16 @@
...
 from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
-from slack_sdk import WebClient
+from airflow.providers.slack.hooks.slack import SlackHook
 
 from src.synapse_hook import SynapseHook
 
 dag_params = {
     "snowflake_developer_service_conn": Param("SNOWFLAKE_DEVELOPER_SERVICE_RAW_CONN", type="string"),
     "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
+    "dpe_slack_bot_conn": Param("DPE_SLACK_BOT_CONN", type="string"),
```
rxu17, line 33: Did we change from using an airflow variable `SLACK_DPE_TEAM_BOT_TOKEN` to airflow connection variable `DPE_SLACK_BOT_CONN` (which i also assume is added to our prod account) for our slack connection?

dags/top-public-synapse-projects-30-days-from-snowflake.py
```diff
@@ -171,4 +171,8 @@ def push_results_to_synapse_table(metrics: List[DownloadMetric], **context) -> N
     top_downloads >> push_to_synapse_table
 
 
-top_public_synapse_projects_30_days_from_snowflake()
+dag = top_public_synapse_projects_30_days_from_snowflake()
+
+if __name__ == "__main__":
+    # Replace with a test Synapse table ID before running locally
+    dag.test(run_conf={"synapse_results_table": "syn74496599"})
```
rxu17, line 178: I am concerned about just having the test/staging table here because in codespaces, if we trigger the dag directly from the UI I don't believe `if __name__ == "__main__":` will run. Should we come up with an alternative safeguard method (like forcing env variables for prod vs dev etc)..?

Does our codespaces use the same aws account (i.e: same airflow variables and connections) as our prod airflow server?

dags/top-public-synapse-projects-from-snowflake.py
```diff
@@ -652,4 +660,16 @@ def push_results_to_synapse_table(
...
 
-top_public_synapse_projects_from_snowflake()
+dag = top_public_synapse_projects_from_snowflake()
+
+if __name__ == "__main__":
+    # backfill=False so the Slack branch runs; slack_channel redirects the post
+    # to #dpe-prs for local testing (the DPE bot must be a member of that channel).
+    # Replace synapse_results_table with a test table ID before running locally.
+    from datetime import date
+    dag.test(run_conf={
```
rxu17, line 670: Similar comment, if we continue to use codespaces as part of development process, this will be an issue I think?

dags/top-public-synapse-projects-30-days-from-snowflake.py
thomasyu888: Good point. I think we will eventually need to come up with dev/prod processing.

I believe your feedback to lingling's agora dag to use env variables is the step in the right direction.

dags/synapse-datasets-created-7-days.py
thomasyu888: I added the slack connection to secrets manager but did not remove anything

dags/synapse-by-the-numbers-dag.py
thomasyu888: Good suggestion + this is a staging table, I can add a comment.

dags/top-public-synapse-projects-30-days-from-snowflake.py
thomasyu888: In addition, im personally not using codespaces so I'm not 100% certain whether the secrets and variables are being pulled from the dev or prod AWS account.  I do think this is something we need to consider as we freshen up the devops.

If you think this type of addition is too soon, I'm happy to comment them out, and rescope this to the slack connector refactor + minor dependency updates

rxu17: I think it may be prod AWS account (as some DAGs I only ever added airflow connections to prod account and they worked in codespaces).

Yeah something to think about like we just want to remove codespaces/local dev container development? But if we don't, then we will likely have to come up with some master config that contains environment (dev vs prod) specifications for each DAG?

dags/synapse-datasets-created-7-days.py
rxu17: Does this make the use of the `SLACK_DPE_TEAM_BOT_TOKEN` void from now on?

I'm seeing [references to it in the CONTRIBUTING.md still](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/fbfb5707e72a4c6bff3269a247a2e837e1c28bf2/CONTRIBUTING.md?plain=1#L315-L322)

thomasyu888: Yes, I'll update contributing docs!

CONTRIBUTING.md
```diff
@@ -293,14 +294,17 @@ AGORA_PROJECT_TOWER_CONN:
...
+`LocalFilesystemBackend`'s `connections_file_path` config above only covers Connections, not Airflow **Variables**. Some DAGs also call `Variable.get("SOME_NAME")` — for example, `SIGNOZ_INGESTION_KEY` or `ZENODO_API_TOKEN`, normally pulled from AWS Secrets Manager. Once you switch away from `SecretsManagerBackend`, there's no source for Variables at all — a plain env var like `SIGNOZ_INGESTION_KEY` in your shell or `.env` is **not** visible to `Variable.get()`. There are two ways to fill that gap without AWS access:
+
+> [!NOTE]
+> Slack credentials are **not** Variables anymore. Every DAG that posts to Slack now uses the Slack provider's `SlackHook` with an Airflow Slack connection (`DPE_SLACK_BOT_CONN`), so the token is supplied through `connections.yaml` like any other connection — see [Configuring a SlackHook connection secret](#configuring-a-slackhook-connection-secret). The former `SLACK_DPE_TEAM_BOT_TOKEN` Variable is no longer read by any DAG.
```
thomasyu888:
```suggestion
```

dags/top-public-synapse-projects-30-days-from-snowflake.py
thomasyu888: Did you want me to comment out the `@dag.test` or keep it as is since the only way these are triggered is by executing it like `python {dag_name}.py`

dags/trending-projects-snapshot-dag.py
rxu17: Will this still work if we parameterize this and use `STACK = 'dev'`?

dags/synapse-datasets-created-7-days.py
```diff
@@ -181,9 +181,9 @@ def generate_slack_message(entity_created: List[EntityCreated], **context) -> st
         return message
 
     @task
-    def post_slack_messages(message: str) -> bool:
+    def post_slack_messages(message: str, **context) -> bool:
         """Post the top downloads to the slack channel."""
-        client = WebClient(token=Variable.get("SLACK_DPE_TEAM_BOT_TOKEN"))
+        client = SlackHook(slack_conn_id=context["params"]["dpe_slack_bot_conn"]).client
         result = client.chat_postMessage(channel="hotdrops", text=message)
```
rxu17: can the channel be a param, and we have a "staging" channel for this? I think we can use [#hotdrop_test](https://sagebionetworks.slack.com/archives/C08F6SKK6Q3)

dags/synapse-datasets-created-7-days.py
rxu17: Should we refactor this to have `dag.test` as well? Also include staging tables for 
```
SYNAPSE_RESULTS_TABLE = "syn64951484"
SYNAPSE_PUBLIC_RESULTS_TABLE = "syn69855119"
```
As well as using `hotdrop_test`

dags/top-public-synapse-projects-30-days-from-snowflake.py
rxu17: Per discussion, let's keep `dag.test`

dags/top-public-synapse-projects-from-snowflake.py
rxu17: Per discussion, we will handle this after we do migration of Airflow

CONTRIBUTING.md
```diff
@@ -589,6 +593,28 @@ New secrets must be created in AWS Secrets Manager in the `dpe-prod` account.
 
 See [Synapse credentials section in the py-orca's env.example](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/main/.env.example) for the expected format of your custom `SYNAPSE_CONNECTION_URI`.
 
+### Configuring a SlackHook connection secret
+
+Slack posting goes through the Slack provider's [`SlackHook`](https://airflow.apache.org/docs/apache-airflow-providers-slack/stable/connections/slack.html), which reads the bot token from an Airflow **connection** (conn type `slack`) rather than from an Airflow Variable. DAGs take the connection ID as a `Param` (default `DPE_SLACK_BOT_CONN`) and do:
```
rxu17, line 598: Is there a link somewhere to where we can add/modify/get a slack bot token?

requirements-local-airflow.txt
```diff
@@ -1,2 +1,16 @@
...
 apache-airflow-providers-amazon ~=9.2.0
+apache-airflow-providers-snowflake ~=6.0.0
+apache-airflow-providers-slack ~=9.0.0
+
+# apache-airflow-providers-celery >=3.10.0
+# snowflake-connector-python ~=3.13.2
+# py-orca[all] == 1.5.2
+# synapseclient[pandas]~=4.0
+# pandas <3.0.0
+# slack-sdk >=3.27
```
rxu17: Nit: We could just remove this since we won't be using it at all? Any reason these are just commented out and not removed?

rxu17 (approved):
Pre-approving! Looks good to me overall, just had some comments

dags/als-kp-dataset-dag.py
rxu17, line 1: Will this be succeeded by https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/130? (reason for removal)

dags/trending-projects-snapshot-dag.py
thomasyu888: Will have to look, actually there is synapse_data_warehouse_dev which houses dev data.

The stack here may not be important, but I should take a look

dags/als-kp-dataset-dag.py
thomasyu888: That's right - we aren't actively using the DAG and there is a refactor so might as well wait for that to be fully merged in.

We may also want to apply more constraints on the ALS dag

CONTRIBUTING.md
thomasyu888: I'll dig up some docs, I actually think this is slack maintained docs

dags/trending-projects-snapshot-dag.py
thomasyu888: The stack value is only prod, i think it would be a larger refactor to parametrize the `synapse_data_warehouse_dev` https://sagebionetworks.jira.com/browse/DPE-1788

CONTRIBUTING.md
thomasyu888: Done here https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/150/commits/c70d5060eeac0dca300a932d2fbdfd8f23d7b5cf

dags/synapse-datasets-created-7-days.py
thomasyu888: Done here https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/150/commits/904d4a0ca30ee378370ebfad01ca3d51c21ec2b8

dags/synapse-datasets-created-7-days.py
thomasyu888: Done here https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/150/commits/904d4a0ca30ee378370ebfad01ca3d51c21ec2b8

PR #155, opened 2026-05-01
[GEN-2607] Modify GENIE Releases Info to update the Patient/Sample tracking table

xindiguo:
# **Problem:**

The list of releases for generating the patient/sample tracking table is currently outdated.

# **Solution:**
Update the config file and release to skip list in the code to retrieve the current list

### DIFF
--- a/src/genie/genie_bpc_releases.yaml
+++ b/src/genie/genie_bpc_releases.yaml
@@ -1,13 +1,28 @@
 # Valid Public Releases 
 - cohort: nsclc
-  version: public_02_2
+  version: public_02_0
   clinical_synid: syn30358089
   cbioportal_synid: syn30358098
 
 - cohort: crc
-  version: public_02_2
+  version: public_02_0
   clinical_synid: syn39802567
   cbioportal_synid: syn39802595
+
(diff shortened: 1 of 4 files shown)


src/genie/genie_bpc_releases.yaml
```diff
@@ -143,7 +143,7 @@
 - cohort: prostate
   version: public_preview_01_0
```
rxu17, line 144: We don't want the public previews right? They are not the official public releases that we want to use

src/genie/genie_bpc_releases.yaml
rxu17, line 1: Are we missing a public release for bladder? Don't remember if they ever had an official one

rxu17 (approved):
LGTM! Just a couple comments

tests/genie/test_main_genie_ingestion.py
rxu17, line 1: Adjusted tests to fit the new genie releases list

thomasyu888:
Discussed with xindi and this can be closed as the relevant tasks on Jira has been closed

PR #159, opened 2026-06-04
[DPE-1645] Add ELITE dataset ID to generate croissant files

thomasyu888:
# **Problem:**
ELITE needed dataset to be added
# **Solution:**
Add dataset

# **Testing:**
NA


dags/synapse-dataset-to-croissant.py
thomasyu888:
```suggestion
    "dataset_collections_for_cleanup": Param(["syn50913342", "syn68939725", "syn71493541", "syn74529385"], type="array"),
```

danlu1 (approved):

PR #160, opened 2026-06-05
[DPE-1645] Add ELITE dataset ID to generate croissant files

thomasyu888:
# **Problem:**
ELITE needed dataset to be added
# **Solution:**
Add dataset

# **Testing:**
NA


thomasyu888:
Approved here: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/159

PR #161, opened 2026-06-09
[DPE-1749] Archive dags

thomasyu888:
# **Problem:**

There are DAGS that are no longer in use that are crowding the airflow space

# **Solution:**

Move to archive_dag folder to save space.

# **Testing:**
NA

To discuss: there are DAGs that are "example" template DAGs and we can potentially use something like .airflowignore to have Airflow ignore them within the UI.


archived_dags/cavatica-launch-poc-v2.py
thomasyu888, line 1: This is a template, but we no longer have this use case

template_dags/nextflow-launch-poc-dag.py
thomasyu888, line 1: This is a template to launch nextflow workflows

thomasyu888: Create a templates folder and rename this `nextflow-launch-poc-dag.py`

archived_dags/cavatica-launch-poc-v2.py
rxu17: Agree with keeping it in the archived_dags

rxu17 (commented):
I would also add the following DAGs to the `templates` folder once it gets created (based on our discussion):
- dags/dcqc-poc-dag.py
- dags/synapse-webhook-poc-dag.py

The DCQC PoC probably needs a boilerplate briefly explaining what it does similar to the synapse webhook dag but that can be later with other documentation updates

After that, this looks good to merge!

template_dags/dcqc-poc-dag.py
rxu17, line 1: Does putting these dags under `dags/templates` make it still discoverable by airflow? Or because of the pathing, they are not adding to the dependencies?

thomasyu888: They are still discoverable by airflow, is your preference to have a top level templates folder so that we aren't polluting the airflow space?

The caveat for these templates is that at one point, they had to be executable

thomasyu888: So i decided to shift these into template_dags outside of the dags folder to reduce adding to depdencies

rxu17 (approved):
LGTM!

PR #165, opened 2026-06-26
[DPE-1732] Add TEP zenodo metrics dag

rxu17:
# **Problem:**
Currently the TREAT-AD Target Enabling Package (TEP) metrics are pulled manually monthly via a script. They are also pulled incorrectly using search terms instead of via community tag (this means we can get fewer records than expected). 

JIRA: https://sagebionetworks.jira.com/browse/DPE-1732

# **Solution:**

## Zenodo Dag
- Add DAG for automating the pulling and exporting of the TEP metrics monthly through the **community tag method** (instead of previously by search terms)
- Add light data validation for the pulled metrics to check that nothing is empty, negative, duplicated, etc
- Add logging and notification system so we get notified if the DAG fails

## Synapse Hook
This is pulled directly from [py-orca's synapse service](https://github.com/Sage-Bionetworks-Workflows/py-orca/tree/main/src/orca/services/synapse) to reduce dependencies (containing the ones we need)

## Synapse Alerts
This PR also adds in synapse alerts that will send an email to the user whenever the Airflow DAG fails for whatever reason. 

Here the DAG is wired to a [standard failures alert through Synapse messaging](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/165/changes#diff-8a6009520168bb59916b2669fdf0585d0dc1a18a12bc2f4c22b2822bbb80422e) (this will send an email to the synapse group/user with an optional custom message) so we are alerted when they fail for any reason. 

You can see this in use [here](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/165/changes#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R281-R362). 

Sample email (when running locally or with codespaces) of the alert you get sent with the custom message: 
image

With codespaces, it uses the orca service account to send the emails.

# **Testing:**

## Unit testing
This PR implements dag specific unit tests and introduces a few standards for unit testing DAGs:
- Having all DAGs have [standard tests that apply to the DAG structure itself (e.g: dag load test, dag structure test) in their test suite](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/d254417ba6365d5ee6d250d743182706fd2b9668/tests/dags/test_zenodo_tep_metrics_dag.py#L336-L363). 
- Migrating out most/all of the python logic in DAGs into module level functions (in the DAG) so unit tests are written directly for those module level functions and greatly reduces complexity of writing unit tests for the DAG task itself (example DAG with [barebones tasks](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/ea212715c291330b44956615d6d042e886bdc02f/dags/zenodo_tep_metrics_dag.py#L405-L476)). This also means every DAG must have unit tests written to be included. 

## Integration testing
This PR pulls the local airflow development from #150 so we can locally test the airflow dag much easier than codespaces (and it was tested on codespaces as well)

Here are the integration test results after I ran just TEP metrics logic code locally and validated it against the consolidated list of expected publications

TEP outputs to synapse folder as 2 separate CSV reports: 
image

TEP output has expected format:
[See JIRA link with attachments](https://sagebionetworks.jira.com/browse/DPE-1732?focusedCommentId=332998)

As a "collaborator", I get sent an email saying new reports are ready: 
image


dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,493 @@
...
+
+            hits = data["hits"]["hits"]
+            if not hits:
+                break
+
+            all_records.extend(hits)
+            logger.info(f"    Retrieved {len(hits)} records")
+
+            total_records = data["hits"]["total"]
+            if len(all_records) >= total_records or len(hits) < params["size"]:
```
rxu17: Here the code uses `all_records` as the determinant for continuously searching for records but we have multiple search terms. But it makes later search terms stop based on records fetched from earlier terms. Need to confirm if this should be the case as we could silently miss pages from lower-priority search terms (if it's priority based) due to this logic

dags/zenodo_tep_metrics_dag.py
rxu17, line 1: All of the logic of code from the original zenodo metrics pulling script is the same (to preserve the same logic and layout, and since results have been validated by collaborators previously) with some adjustments and functionalizing so it's easier to follow.

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,493 @@
...
+            f"https://www.synapse.org/Synapse:{file_info['id']}"
+        )
+        _send_synapse_message(
+            conn_id=context["params"]["synapse_conn_id"],
+            usernames=context["params"]["collaborator_user_list"].split(","),
+            subject=subject,
+            body=body,
+        )
+
+    records = fetch_metrics()
```
rxu17, line 545: Here we have separated out the tasks so that it's independently retryable and visible in the UI, we could merge fetch+validate into one task, but then a validation failure and a fetch metrics failure wouldn't be distinguishable as separate task states. Another downside is we technically "expose" the records through xcom (anyone could view it) through the passing of the data through the different tasks though the records are aggregates so this may not be as big of an issue

dags/zenodo-tep-metrics-dag.py
```diff
@@ -0,0 +1,493 @@
...
+            body=body,
+        )
+
+    records = fetch_metrics()
+    validated = validate_metrics(records)
+    file_info = export_to_synapse(validated)
+    notify_collaborators(file_info)
+
+
+zenodo_tep_metrics_dag()
```
thomasyu888: There is a way to support local airflow dev without codespaces that might make our lives easier that we may want to make a standard: See this PR: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/150.

requirements-airflow.txt
```diff
@@ -10,3 +10,4 @@ jsonata-python ~=0.5.3
 boto3 >=1.7.0,<2.0
 requests ~=2.31
 PyYAML ~=6.0.3
+openpyxl ~=3.1
```
thomasyu888: Could this be a `@virtualenv` addition to avoid adding more dependencies to the airflow executor environemtn?

dags/zenodo-tep-metrics-dag.py
```diff
@@ -0,0 +1,493 @@
...
+
+    if errors:
+        raise ValueError(
+            "Zenodo TEP metrics validation failed:\n- " + "\n- ".join(errors)
+        )
+
+    logger.info(f"Validation passed for {len(records)} records.")
+
+
+def build_workbook(records: List[Dict[str, Any]], filepath: str) -> Tuple[int, int]:
```
thomasyu888: Could you link an example of a excel workbook? I'm curious to understand what excel features we are using that couldn't be accomplished with just a CSV

dags/zenodo-tep-metrics-dag.py
rxu17: Yup, thanks I saw that. Trying to get it to work on my end

dags/zenodo-tep-metrics-dag.py
rxu17: I think it's mainly to have two sheets of data - TEP Reports and TEP Components in one workbook?

thomasyu888: Thanks - this is helpful

requirements-local-airflow.txt
```diff
@@ -0,0 +1,14 @@
+-c https://raw.githubusercontent.com/apache/airflow/constraints-2.10.5/constraints-3.10.txt
```
thomasyu888, line 1: Thoughts of trimming this down and starting fresh per DAG? This way we can rebuild our dependency layer as we add the ability to develop locally for the 9 production DAGs we have.

requirements-airflow.txt
rxu17: Opted to remove this instead and output CSVs instead

dags/zenodo_tep_metrics_dag.py
rxu17: We decided to use the community tag method to search for records as that was the original planned method. This eliminates the issues we face here with the multiple search terms method

dags/zenodo_tep_metrics_dag.py
rxu17: Zenodo API allows for open access as it's an open repo so this is okay

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,479 @@
...
+    "Total Downloads",
+    "Unique Downloads",
+    "Link",
+]
+
+dag_params = {
+    "synapse_conn_id": Param("SYNAPSE_ORCA_SERVICE_ACCOUNT_CONN", type="string"),
+    # Synapse folder/project the Excel report is uploaded to. Update to the
+    # shared TREAT-AD reporting location.
+    "synapse_export_folder": Param("syn75951837", type="string"),
```
rxu17: Will modify to be production synapse folder once approved

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,479 @@
...
+    # Comma-separated Synapse usernames of collaborator(s) to notify when the
+    # monthly report is available. (No support for python lists in Params.)
+    "collaborator_user_list": Param("3460442", type="string"),
+    # Comma-separated Synapse usernames or IDs of DPE/developers to 
+    # alert on failure.
+    "dev_user_list": Param("3460442", type="string"),
```
rxu17: Will modify to be DPE service team for `dev_user_list` and list of actual collaborators to receive the reports in `collaborator_user_list`

dags/zenodo_tep_metrics_dag.py
rxu17: Not relevant anymore as we've changed the internal record pulling logic and output formatting logic

CONTRIBUTING.md
```diff
@@ -97,6 +97,46 @@ tests/
...
+from dags import zenodo_tep_metrics_dag as dag_module   # the DAG under test
+```
+
+A few requirements make this import work:
+
+- **The DAG filename must be a valid Python module name** — use underscores, not
+  hyphens so it can be imported with `from dags import <dag_module>` (e.g.
+  `dags/zenodo_tep_metrics_dag.py`, not `dags/zenodo-tep-metrics-dag.py`)
+- **Keep `dags/`, and the top-level `src/` as namespace packages**, 
+  do not add an `__init__.py` to any of them. Because neither `src/` nor
```
thomasyu888, line 160: Is the init file needed for production airflow to read the files in dags/src?

thomasyu888 (commented):
🔥 LGTM!

1. I like the alert on failure - this is a good example of a codespaces and/or team meeting topic so that the team is made aware
2. Is there an airflow best practice on whether to or not to include helper functions within the dag script itself

Pre-approving, but will defer to lingling for final review.

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,510 @@
...
+    # monthly report is available. (No support for python lists in Params.)
+    "collaborator_user_list": Param("3460442", type="string"),
+    # Comma-separated Synapse usernames or IDs of DPE/developers to
+    # alert on failure.
+    "dev_user_list": Param("3460442", type="string"),
+}
+
+dag_config = {
+    # Run on the first day of the month at midnight
+    "schedule_interval": "0 0 1 * *",
```
linglp: I think `schedule_interval` will be deprecated: https://airflow.apache.org/docs/apache-airflow/2.4.0/release_notes.html. I think going forward, we should use `schedule` to reduce tech debt.

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,510 @@
...
+import csv
+import logging
+import os
+from datetime import datetime
+from typing import Any, Dict, List, NamedTuple
+
+from airflow.decorators import dag, task
+from airflow.models import Param, Variable
+import requests
+from synapseclient.models import File, Folder
```
linglp: Remove `Folder` since `Folder` is not used

dags/src/synapse_hook.py
```diff
@@ -0,0 +1,62 @@
...
+            synapse_client=self.client,
+        )
+        return df["id"].tolist()
+
+    def update_submission_status(
+        self,
+        submission_id: Union[int, str],
+        submission_status: str,
+    ) -> None:
+        if type(submission_id) not in [str, int]:
```
linglp: should we use `isinstance()` instead of `type` here?

dags/src/synapse_hook.py
```diff
@@ -0,0 +1,62 @@
+import os
+from typing import Union
+
+import synapseclient
+from synapseclient.models import query
+
+
+class SynapseHook:
```
linglp, line 8: Should we also add test for SynapseHook? Or should that be a different ticket?

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,510 @@
...
+    """
+    errors: List[str] = []
+
+    missing = [f for f in REQUIRED_RECORD_FIELDS if f not in record]
+    if missing:
+        errors.append(f"Validation failed: record {index} missing fields: {missing}")
+
+    for field in METRIC_FIELDS:
+        value = record.get(field)
+        if not isinstance(value, int) or isinstance(value, bool) or value < 0:
```
linglp, line 227: Do we know if something like `10.0` or float would get returned?

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,510 @@
...
+            records (List[Dict[str, Any]]): Records from fetch_metrics
+
+        Returns:
+            List[Dict[str, Any]]: The validated records
+        """
+        validate_records(records)
+        return records
+
+    @task
+    def export_to_synapse(
```
linglp, line 501: should we add a unit test for `export_to_synapse`?

linglp (approved):
LGTM @rxu17 ! Thanks for working on this. I have some small comments below, but overall it looks good. I am going to pre-approve now.

dags/zenodo_tep_metrics_dag.py
rxu17: As far as I know, no floats get returned. I wonder if that would be expected behavior we would want to let through (we get a report with floats which may be odd for the viewer) or we want that flagged so we can take a look.

dags/zenodo_tep_metrics_dag.py
rxu17: I can move some of the logic out to a module level function so we can unit test it (otherwise I think the tests for @task function logic get complex). I think preferred practice is the above

CONTRIBUTING.md
rxu17: I think right now it has no problem reading the files in `dags` right? But for `dags/src`, that's what I don't know (without running it in production) and what we will find out

thomasyu888: Let's try it!

rxu17:
> Is there an airflow best practice on whether to or not to include helper functions within the dag script itself

If by helper functions you just mean the non-dag task functions and such, [here is a documentation page that discusses that](https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#top-level-python-code). Main concern also seems to be [dag loading time](https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#dag-loader-test) and so we'll see if we need to do something like the following:
```python
@task()
def do_stuff_with_pandas_and_torch():
    import pandas
    import torch

```
instead of 

```python
import pandas
import torch 

@task()
def do_stuff_with_pandas_and_torch():
```
This exploration could definitely be this ticket or another ticket (to avoid delays on this)

tests/dags/test_zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,431 @@
...
+            [VALID_RECORD],
+            run_date="20250101",
+            synapse_conn_id="conn",
+            folder_id="syn123",
+        )
+    # Even when the upload raises, the local CSVs are removed.
+    assert not list(tmp_path.glob("*.csv"))
+
+
+def test_dag_loads_without_import_errors():
```
rxu17: This can be a dynamic dag test (so each dag doesn't need this in the individual test suites) in the future but for now let's add it iteratively as some dags already have import errors and we don't want this test to be enforced. Although if we have a PR that updates a common shared function across multiple DAGs we may want to write these tests in for all of the DAGs then

dags/src/synapse_hook.py
rxu17: Yes I can definitely add tests (just copied this over from #150), since I think the plan is to have this PR be merged first before #150 ? @thomasyu888 If so, I'll also add in the local airflow development instructions from there as well

thomasyu888: That'd be wonderful, thank you so much.

dags/src/synapse_hook.py
```diff
@@ -0,0 +1,62 @@
...
+        return df["id"].tolist()
+
+    def update_submission_status(
+        self,
+        submission_id: Union[int, str],
+        submission_status: str,
+    ) -> None:
+        if not isinstance(submission_id, (str, int)):
+            raise TypeError("submission_id must be a string or int.")
+        sub_status = self.client.getSubmissionStatus(submission_id)
```
rxu17, line 40: Had to use `getSubmissionStatus` through the client instead of 

```python
from synapseclient.models import SubmissionStatus
...
sub_status = SubmissionStatus(id=submission_id).get()
```
because of the `synapseclient` version provided by `py-orca` is a bit older (4.8) so we'd need to update it to >=4.11 to use the new submission status model. I'll create a ticket for that?

Or do we set plans in motion to migrate away from `py-orca`?

thomasyu888: I would create a ticket first. we are going to have to re-align here: https://sagebionetworks.jira.com/browse/DPE-1652

For security maintainance, I simply don't think we have the bandwidth on the team to maintain the K8 stack if we are _just_ using it for Airflow, so something like MWAA would be helpful.

README.md
```diff
@@ -118,6 +137,54 @@ The username and password will be "airflow".
 If you encounter the `nginx bad gateway` errors when navigating to the forwarded port, just wait and refresh a couple of times. Airflow takes a few minutes to become available.
 
 
+### Local Development (without Dev Container)
```
rxu17, line 139: Pulled from Tom's local airflow dev PR #150 with slight modifications

README.md
```diff
@@ -118,6 +137,54 @@ The username and password will be "airflow".
...
+
+If you want to test a DAG locally without Docker or Codespaces, you can run it directly against a local Airflow SQLite database.
+
+#### 1. Install dependencies
+
+```console
+pip install -r requirements-local-airflow.txt
+```
+
+Then install `synapseclient` separately. This must be done after the above to avoid a `urllib3` version conflict introduced by the Airflow constraints file:
```
thomasyu888: Could we flag this up on dpe-team? Perhaps the SYNPY devs could assist with this

thomasyu888: Or it may not be worth it if the client works with airflow 3

README.md
```diff
@@ -118,6 +137,54 @@ The username and password will be "airflow".
...
+
+```console
+pip install "synapseclient[pandas]"
```
rxu17: Might need to just add a step to `pip install setuptools<82`, our current production airflow deployment seems to be fine / haven't run into the `setuptools` issue?

thomasyu888 (commented):
🔥 LGTM, thanks for the great additions, and I'd like to defer to @linglp for final review.

I see that there are now merge conflicts, so ill defer to you on how best to address those

tests/dags/test_zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,363 @@
...
+    reports_path = str(tmp_path / "metrics_TEP_Reports.csv")
+    components_path = str(tmp_path / "metrics_TEP_Components.csv")
+
+    assert result["report_records"] == 1
+    assert result["component_records"] == 1
+    assert result["paths"] == [reports_path, components_path]
+
+    # CSV cells are read back as strings.
+    reports = _read_csv(reports_path)
+    assert reports[0] == dag_module.EXCEL_HEADERS  # header row
```
linglp: the headers are CSV headers? and not Excel headers?

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,489 @@
...
+    "unique_downloads",
+    "link",
+    "category",
+]
+
+# Numeric metric fields that must be non-negative integers
+METRIC_FIELDS = ["views", "unique_views", "downloads", "unique_downloads"]
+
+# Headers for the Excel report (in the order they appear in the workbook)
+EXCEL_HEADERS = [
```
linglp: The output is in CSV format, so the headers here are CSV headers? And the comment needs to be updated?

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,489 @@
...
+) -> List[Dict[str, str]]:
+    """Build the CSV reports and upload them to a Synapse folder.
+
+    Local CSV files are always removed afterwards, even if an upload fails.
+
+    Arguments:
+        records (List[Dict[str, Any]]): Validated TEP records
+        run_date (str): YYYYMMDD date string used in the report filenames
+        synapse_conn_id (str): Synapse connection id
+        folder_id (str): Synapse folder/project to upload the reports to
```
linglp: Nit: this is called `folder_id`, so this should just be `folder`?

README.md
linglp: I would rather put it in `requirements-local-airflow.txt` than as a manual step in the README, so that we could keep the install instructions to just `pip install -r requirements-local-airflow.txt`

README.md
linglp: Should these be moved to contribution guide?

linglp (approved):
LGTM! Just some minor comments

dags/zenodo_tep_metrics_dag.py
rxu17: Thanks for the catch!

dags/zenodo_tep_metrics_dag.py
rxu17: Corrected to specify that it's the Synapse ID of the Synapse folder and changed `synapse_export_folder` to be `synapse_export_folder_id`

dags/src/synapse_alerts.py
```diff
@@ -0,0 +1,136 @@
...
+
+See CONTRIBUTION.md section DAG Failure Alerts for more info and usage examples.
+"""
+import logging
+from typing import Any, Callable, Dict, List, Optional
+from urllib.parse import urlparse
+
+from src.synapse_hook import SynapseHook
+
+logger = logging.getLogger(__name__)
```
BryanFauble: nit: This could also be an opportunity to create a reusable logging configuration every DAG can import/use.

dags/src/synapse_alerts.py
```diff
@@ -0,0 +1,136 @@
...
+        subject (str): Message subject
+        body (str): Message body
+    """
+    usernames = [u.strip() for u in usernames if u.strip()]
+    if not usernames:
+        logger.warning(f"No Synapse users provided; skipping message: {subject}")
+        return
+    client = SynapseHook(conn_id).client
+    owner_ids = [client.getUserProfile(u).get("ownerId") for u in usernames]
+    client.sendMessage(owner_ids, subject, body)
```
BryanFauble: nit: https://rest-docs.synapse.org/rest/POST/message.html is the endpoint that this calls and returns data for - There could be a log message here indicating something from the response call to Synapse (https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/message/MessageToUser.html )

dags/zenodo_tep_metrics_dag.py
```diff
@@ -0,0 +1,491 @@
...
+                    sum(r["downloads"] for r in output.records),
+                    sum(r["unique_downloads"] for r in output.records),
+                    "",
+                ]
+            )
+
+    logger.info(
+        f"Wrote CSV reports to {outputs[0].path} and {outputs[1].path} "
+        f"({len(main_teps)} reports, {len(supporting)} components)"
+    )
```
BryanFauble, line 389: This should be wrapped in a try/final block. In the final block the CSV file should be removed. There is a small edge case since this is called in a loop where a second or later iteration of the loop raises an exception and `export_reports_to_synapse` doesn't run it's try/final block to remove the CSV files.

BryanFauble (approved):
A few small changes to address, and nothing beyond what has already been flagged by others.

Thank you for the updates @rxu17 !

README.md
rxu17: @linglp @thomasyu888 

I found that having `setuptools` in the `requirements-local-airflow.txt` config, and doing that before installing `synapseclient` will just have it uninstall the `setuptools` later to be later version like 83. 

Airflow's suggestion here: https://github.com/apache/airflow/blob/main/constraints/README.md is to just modify their constraints file altogether and have our own local version. 

So we could just modify it all together (e.g: change the conflicting `urllib3` constraint for example so we can install `synapseclient` within the requirements.txt) in our local version of the 2.10.5 constraints.

Other way is to just have multiple shell commands to run (and create a shell script for it) so we don't have to maintain a full local copy but developers will just have to run the shell script then.

e.g:
```console
#!/usr/bin/env bash
set -euo pipefail

AIRFLOW_VERSION="2.10.5"
PYTHON_VERSION="3.11"

echo "Installing dependencies..."
python -m pip install \
  -c "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-${PYTHON_VERSION}.txt" \
  "apache-airflow-providers-amazon~=9.2.0"

python -m pip install \
  "synapseclient[pandas]~=4.8" \
  "setuptools<82"

echo "Verifying environment..."
python -m pip check
```

And we can maintain a `requirements-local-airflow.txt` still or just add them here in the shell script. Either way, we end up with one extra file whether it's the local copy of the constraints or the shell script

linglp: Thanks for pointing this out, @rxu17 ! I tried it on my end, and I could replicate what you saw. I think: 
* `pip install -r requirements-local-airflow.txt`: the `setuptools <82` pin lives inside that requirements file, so pip's resolver for that command respects the cap.
* But when we did: `pip install "synapseclient[pandas]"`, this is a separate command, and there's no cap for `setuptools` 

In your script, Python version should probably be 3.10 instead of 3.11, to stay consistent with the workflow file and `local/dev_setup.sh`?

I think we could have something like: 
```
echo "Installing dependencies..."
python -m pip install -r requirements-local-airflow.txt

echo "Installing synapseclient (kept separate to avoid a urllib3 conflict)..."
python -m pip install "synapseclient[pandas]" "setuptools<82"
```
in the shell script. If we have both `requirements-local-airflow.txt` and something like: 
```
python -m pip install \
  -c "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-${PYTHON_VERSION}.txt" \
  "apache-airflow-providers-amazon~=9.2.0"
```
I think we then risk maintaining the same dependencies in both places.

rxu17: Yup I just had python 3.11 as an example. Do you mean just having those manual instructions in the markdown then? 

Yeah I'm fine with having the manual instructions as well (I think we should constrain to synapseclient 4.8 to match `py-orca` as that's what the production airflow uses I believe)

So actually if we do the above shell script example that I put, I'm don't think we would be maintaining the same dependencies in both places? The `requirements-local-airflow.txt` could look like 

```
-c https://raw.githubusercontent.com/apache/airflow/constraints-2.10.5/constraints-3.10.txt
... [any other libraries]
```

And the shell script would be something like:
```shell
#!/usr/bin/env bash
set -euo pipefail

echo "Installing dependencies..."
pip install -r requirements-local-airflow.txt

python -m pip install \
  "synapseclient[pandas]~=4.8" \
  "setuptools<82"

echo "Verifying environment..."
python -m pip check
```

Then the docs would update to just run shell script. I'm fine with just putting it manually in the docs as well (one less file), either method, we'll just have to update it when the time comes for migration/other things

linglp: I think our understanding is aligned! I am slightly in favor of the shell script approach because it lets developers run a single command instead of two separate steps.

README.md
rxu17: I was debating it but since we have the codespaces airflow development setup part in README, for consistency, I decided to keep the airflow development (just the base setup) in same section in README but link out to the more detailed development steps and options in CONTRIBUTING

README.md
rxu17: It looks like it's hard to bypass (even airflow themselves have a workaround here: https://github.com/apache/airflow/blob/main/constraints/README.md) so we opted for local airflow setup bash script instead, see thread below

PR #167, opened 2026-07-03
[AG-2123] Added instructions to use connection YAML file when connecting with nextflow tower locally

linglp:
# Problem:
Testing a DAG locally that talks to real services (e.g., Nextflow Tower) currently requires authenticating to AWS Secrets Manager just to resolve Airflow Connections/Variables. This is unnecessary friction for quick local iteration and blocks contributors without AWS SSO access.

# Solution: 
1. Document `LocalFilesystemBackend` (backed by a local `connections.yaml`) as an alternative to
  `SecretsManagerBackend`, so contributors can test against real services like Nextflow Tower
  without AWS SSO.
2. Add a "Testing DAGs Locally" section clarifying unit tests vs. local integration runs

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,12 @@ Then install `synapseclient` separately. This must be done after the above to av
 pip install "synapseclient[pandas]"
 ```
 
+Then install `py-orca`, which depends on `synapseclient` and pulls in the same `urllib3` conflict, so it must also be installed after the above:
+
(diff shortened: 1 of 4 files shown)


linglp:
Close this pull request and open: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/168

PR #168, opened 2026-07-06
[AG-2123] Added instructions to use connection YAML file for local airflow development 

linglp:
# **Problem:**
Testing a DAG locally that talks to real services (e.g., Nextflow Tower) currently requires authenticating to AWS Secrets Manager just to resolve Airflow Connections/Variables. This is unnecessary friction for quick local iteration and blocks contributors without AWS SSO access.

# **Solution:**
Document `LocalFilesystemBackend` (backed by a local connections.yaml) as an alternative to SecretsManagerBackend, so contributors can test against real services like Nextflow Tower without `AWS SSO`.

# **Note:**
I skipped the part related to install dependencies for local development since they are part of https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/150


.gitignore
```diff
@@ -133,6 +133,7 @@ logs/
 
 # Secrets environment
 .env
+connections.yaml
```
linglp, line 136: added this because in the future, `connections.yaml` will have personal access token, and we don't want to commit this.

README.md
```diff
@@ -59,6 +78,66 @@ docker compose up --build --detach
...
+  host: tower.sagebionetworks.org
+  schema: api
+  password: "<your-tower-personal-access-token>"
+  extra:
+    workspace: sage-bionetworks/<your-workspace>
+```
+
+`host`/`schema` become the Tower API endpoint (`https://<host>/<schema>`), `password` is your Tower personal access token, and `extra.workspace` is the fully-qualified `<org>/<workspace>` name.
+
+`LocalFilesystemBackend`'s `connections_file_path` config above only covers Connections, not Airflow **Variables**. Some DAGs also call `Variable.get("SOME_NAME")` — for example, `SLACK_DPE_TEAM_BOT_TOKEN`, a Slack bot token normally pulled from AWS Secrets Manager. Once you switch away from `SecretsManagerBackend`, there's no source for Variables at all — a plain env var like `SLACK_DPE_TEAM_BOT_TOKEN` in your shell or `.env` is **not** visible to `Variable.get()`.
```
linglp: this is to explain why you can't just list the environment variables in `.env`

README.md
```diff
@@ -59,6 +78,66 @@ docker compose up --build --detach
 
 Congrats! You have completed set up of the dev environment.
 
+### Testing DAGs Locally
+
+There are two distinct ways to test a DAG's task logic without deploying to Airflow:
+
+- **Unit tests** — mock external dependencies (hooks like `NextflowTowerHook`/`SynapseHook`, `WebClient`, `Variable`, etc.) and call each task's `python_callable` directly. These are fast, require no credentials or network access, and run automatically in CI on every push. They validate your Python logic (parameter wiring, branching, string formatting), not whether the real external services behave as you assume. See `tests/` for examples.
+
+- **Local integration runs** via `dag.test()` — hit real external services with real credentials pulled from AWS Secrets Manager. These validate the real integration (e.g., actually launching a Nextflow Tower workflow or posting to Slack), but are slower, require AWS SSO access, and can have real side effects. Use this as a manual sanity check before/after changing integration behavior, not as an automated substitute for unit tests.
```
rxu17:
```suggestion
- **Local integration runs** via `dag.test()` — hit real external services using real credentials. Credentials can come from AWS Secrets Manager (matching production) or from a local `connections.yaml/AIRFLOW_VAR_*` configuration via `LocalFilesystemBackend`. These validate the real integration (e.g., actually launching a Nextflow Tower workflow or posting to Slack), but are slower, require AWS SSO access, and can have real side effects. Use this as a manual sanity check before/after changing integration behavior, not as an automated substitute for unit tests.
```

README.md
```diff
@@ -59,6 +78,66 @@ docker compose up --build --detach
...
+  host: tower.sagebionetworks.org
+  schema: api
+  password: "<your-tower-personal-access-token>"
+  extra:
+    workspace: sage-bionetworks/<your-workspace>
+```
+
+`host`/`schema` become the Tower API endpoint (`https://<host>/<schema>`), `password` is your Tower personal access token, and `extra.workspace` is the fully-qualified `<org>/<workspace>` name.
+
+`LocalFilesystemBackend`'s `connections_file_path` config above only covers Connections, not Airflow **Variables**. Some DAGs also call `Variable.get("SOME_NAME")` — for example, `SLACK_DPE_TEAM_BOT_TOKEN`, a Slack bot token normally pulled from AWS Secrets Manager. Once you switch away from `SecretsManagerBackend`, there's no source for Variables at all — a plain env var like `SLACK_DPE_TEAM_BOT_TOKEN` in your shell or `.env` is **not** visible to `Variable.get()`.
```
rxu17: Would try to split up these two sections - using connections.yaml and exporting airflow vars as separate so that it doesn't appear the export airflow vars is required for the above section? Since there are DAGs that just need the exporting airflow vars (instead of using aws secrets manager)?

README.md
```diff
@@ -59,6 +78,66 @@ docker compose up --build --detach
...
+
+```console
+cp connections.yaml.example connections.yaml
+```
+
+Fill in the real credentials for whichever connections your DAG needs. For example, to launch workflows on the real Nextflow Tower:
+
+```yaml
+AGORA_PROJECT_TOWER_CONN:
+  conn_type: tower
```
rxu17: Would these keys be the same for every nextflow tower launch?

README.md
```diff
@@ -59,6 +78,66 @@ docker compose up --build --detach
 
 Congrats! You have completed set up of the dev environment.
 
+### Testing DAGs Locally
```
rxu17: It feels like this may be a better fit in the CONTRIBUTING.md? There is a testing section: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/CONTRIBUTING.md#testing

README.md
linglp: I think so, but you have to use your own `tower_conn_id`

README.md
linglp: Yes, I think this idea.

rxu17 (approved):
LGTM! Just a few comments. Do you want to merge this first before any other PRs after addressing the comments so we can start to consolidate our airflow doc changes across all of our PRs?

PR #170, opened 2026-07-14
fix: fix import error in agora_nf

linglp:
# **Problem:**
```
traceback (most recent call last):
  File "/Users/lpeng/code/orca-recipes/dags/agora_nf.py", line 45, in <module>
    from dags.src.utils import validate_required_secrets
ModuleNotFoundError: No module named 'dags'
```
# **Solution:**
Fix import error

### DIFF
--- a/dags/agora_nf.py
+++ b/dags/agora_nf.py
@@ -219,4 +219,4 @@ if __name__ == "__main__":
         ],
         variable_names=["SLACK_DPE_TEAM_BOT_TOKEN"],
     )
-    # dag.test(run_conf={"dataset": "model_details"})
+    dag.test(run_conf={"dataset": "model_details"})
--- a/dags/agora_nf.py
+++ b/dags/agora_nf.py
@@ -34,14 +34,16 @@ DAG Parameters:
 import os
 from datetime import datetime
 from typing import Any
+
 from airflow.decorators import dag, task
+from airflow.models import Variable
 from airflow.models.dag import DAG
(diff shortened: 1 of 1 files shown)


dags/agora_nf.py
```diff
@@ -42,7 +42,7 @@
 from orca.services.nextflowtower.models import LaunchInfo
 from slack_sdk import WebClient
 from orca.services.synapse import SynapseHook
-from dags.src.utils import validate_required_secrets
+from src.utils import validate_required_secrets
```
thomasyu888, line 47: NIt: could we follow pep8 standards for import: https://peps.python.org/pep-0008/#imports?

rxu17 (approved):
LGTM!

dags/agora_nf.py
linglp: import sorting should in the future be fixed by `ruff` in pre-commit. We already have this ticket: https://sagebionetworks.jira.com/browse/DPE-1762

dags/agora_nf.py
```diff
@@ -217,4 +219,4 @@ def post_email_messages(message: str, **context: Any) -> bool:
         ],
         variable_names=["SLACK_DPE_TEAM_BOT_TOKEN"],
     )
-    dag.test(run_conf={"dataset": "model_details"})
+    # dag.test(run_conf={"dataset": "model_details"})
```
thomasyu888: I don't think we need to comment this out?

PR #171, opened 2026-07-14
[IT-4153] Use developer AWS SSO credentials instead of shared IAM key

BryanFauble:
# **Problem:**

Local development authenticates to AWS Secrets Manager (`dpe-prod`) using a **shared** `airflow-secrets-backend` IAM access key, fanned out as GitHub Codespace secrets and injected into the docker-compose stack as `AWS_ACCESS_KEY_ID/SECRET/SESSION_TOKEN`. That key must be manually rotated every 90 days for CIS 1.14 ([IT-4153](https://sagebionetworks.jira.com/browse/IT-4153)), and a shared long-lived key handed to every developer is itself a security smell.

# **Solution:**

Developers authenticate with their **own** AWS Identity Center (SSO) credentials — the existing `dpe-prod` `Developer` permission set — so there is no shared key to rotate. The secrets backend uses boto3's default credential chain, so SSO-derived credentials resolve transparently.

- **`docker-compose.yaml`**: add `AWS_PROFILE` / `AWS_REGION` passthrough and an optional read-only `~/.aws` mount (`${HOST_AWS_DIR:-${HOME}/.aws}` → `/home/airflow/.aws`), while keeping the existing exported-credential passthrough. Two supported paths:
  - **Local Dev Container / VS Code** — set `AWS_PROFILE` (+ `HOST_AWS_DIR` if needed); the mounted profile auto-refreshes SSO tokens.
  - **Codespaces** — `aws sso login`, then `bash scripts/aws-sso-to-env.sh <profile>` writes short-lived creds into `.env`.
- **`scripts/aws-sso-to-env.sh`** (new): logs in via SSO and exports short-lived credentials into `.env` (strips any stale ones first).
- **`.env.example`**: two labeled, mutually-exclusive credential blocks.
- **`README.md` / `CONTRIBUTING.md`**: new "AWS credentials" section documenting the one-time `aws configure sso` setup and both delivery paths; Secrets + Troubleshooting sections rewritten for SSO; removed the shared-IAM-user / manual-90-day-rotation guidance and the Confluence rotation-runbook link.

Codespaces remains supported — just SSO-authenticated per developer rather than relying on injected shared-key secrets.

# **Testing:**

- `docker compose config` renders successfully: `AWS_PROFILE` interpolates, `AWS_REGION` defaults to `us-east-1`, and the `~/.aws` mount resolves to an absolute host path mounted read-only at `/home/airflow/.aws` (correct `HOME` for the uid-50000 airflow user).
- `bash -n` syntax check on the helper script; confirmed `aws configure export-credentials --format env-no-export` emits exactly `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN` (the vars compose reads).
- Verified README/CONTRIBUTING anchor links resolve; added the new section to the TOC.

Reviewer verification:
1. `aws sso login --profile <p>`; set `AWS_PROFILE` in `.env`; `docker compose up --build --detach`; `./airflow.sh python` → run the boto3 snippet in the README → lists `airflow/connections/*` + `airflow/variables/*` with no `UnrecognizedClientException`.
2. Codespaces path: unset `AWS_PROFILE`; `bash scripts/aws-sso-to-env.sh <p>`; `docker compose up`; same snippet succeeds.

## Note / dependency

Deployed Airflow's keyless access is handled by the companion PR (`eks-stack` #88, `airflow-croissant` IRSA role). The shared `airflow-secrets-backend` key and its Codespace secrets should be deleted only **after both PRs are live and verified** — this PR removes the local dependency on that key; #88 removes the prod dependency.

[IT-4153]: https://sagebionetworks.jira.com/browse/IT-4153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -573,13 +573,13 @@ Airflow secrets (_e.g._ connections and variables) are stored in Secrets Manager
 
 These are all located in AWS Secrets Manager in the `dpe-prod` account.
 
-Developers access these secrets with their own AWS Identity Center (SSO) credentials — there is no shared IAM user or long-lived access key to manage or rotate. Configure an SSO profile once (`aws configure sso`, start URL `https://d-906769aa66.awsapps.com/start`, region `us-east-1`, account `766808016710`, `Developer` role) and provide it to the containers per the [AWS credentials](./README.md#aws-credentials-required-for-all-dev-container-options) section of the README. In deployed Airflow, the pods assume an IRSA role (`airflow-croissant`) and no credentials are stored in the connection secrets.
+Developers access these secrets with their own AWS Identity Center (SSO) credentials — there is no shared IAM user or long-lived access key to manage or rotate. Set up an SSO profile once and log in per the [AWS credentials](./README.md#aws-credentials-required-for-all-options) section of the README (the source of truth for the `aws configure sso` / `aws sso login` steps). In deployed Airflow, the pods assume an IRSA role (`airflow-croissant`) and no credentials are stored in the connection secrets.
 
 ### Creating a new secret
 
 New secrets must be created in AWS Secrets Manager in the `dpe-prod` account.
 
-1. You will need at least **Developer** access to the account.
+1. You will need at least **Developer** (or **Administrator**) access to the account.
 1. Make sure your region is set to **us-east-1**.
 1. For connection URIs, the secret name should have the prefix `airflow/connections/`
 (i.e. `airflow/connections/MY_SECRET_CONNECTION_STRING`). Variables should have the prefix `airflow/variables/` (i.e. `airflow/variables/MY_SECRET_VARIABLE`).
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -752,31 +752,30 @@ By following these guidelines, you will successfully contribute a deployable cha
...
-**To fix it:**
+* **Codespaces** (using exported credentials in `.env`): re-run the helper to write fresh short-lived credentials, then restart the stack so the containers pick them up:
+  ```console
+  aws sso login --profile <your-sso-profile>
+  bash scripts/aws-sso-to-env.sh <your-sso-profile>
+  docker compose up --build --detach
+  ```
 
-Follow the [Rotating `airflow-secrets-backend` IAM User Credentials](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/4530602005/Rotating+airflow-secrets-backend+IAM+User+Credentials#Phase-3:-Update-Use-Case-1-(GitHub-Codespaces)) runbook on Confluence, which covers how to rotate the keys, update the Codespace secrets, and validate the update.
+If it still fails, confirm your profile is configured and you are assigned the `Developer` role in `dpe-prod` (`766808016710`):
```
rxu17: nit: Could also have the Administrator role as well?

CONTRIBUTING.md
```diff
@@ -752,31 +752,30 @@ By following these guidelines, you will successfully contribute a deployable cha
...
-   ```console
-   env | grep AWS
-   ```
-2. Locate the value of `AWS_ACCESS_KEY_ID` in the output.
-3. Compare it against the active access key for the `airflow-secrets-backend` IAM user in the AWS Console (`org-sagebase-dpe-prod` account).
-   Alternatively, you can list active keys with the AWS CLI:
-   ```console
-   aws iam list-access-keys --user-name airflow-secrets-backend --profile <your-profile-name>
-   ```
+**To fix it, refresh your SSO session:**
```
rxu17: These directions feels a bit redundant with the directions in:
- AWS credentials (required for all dev container options)
- Codespaces
- Docker Compose sections
- [aws-sso-to-env.sh](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/171/changes#diff-779e61577f3fd4b80b99c36c499293f71fee449318da90eb8d3ed433b49201a0) header comments 
- [‎.env.example](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/171/changes#diff-a3046da0d15a27e89f2afe639b25748a7ad4d9290af3e7b1b6c1a5533c8f0a8c) comments. 
- Some parts of docker-compose.yaml

Is there a way to just link to one section that has the source of truth instructions (e.g: `aws sso login --profile`, `bash scripts/aws-sso-to-env.sh <your-sso-profile>`) in all of the other instructions? Or just consolidate?

rxu17: Also given that now [local development with Airflow (no dev containers)](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/README.md#local-development-without-dev-container) is an option, and now `aws sso login` is now the single common prerequisite across all three paths (local without dev container, local with dev container and codespaces) could that be added as a higher level pre-requisite?

rxu17 (approved):
LGTM! Had a few comments

CONTRIBUTING.md
BryanFauble: Yes!

CONTRIBUTING.md
BryanFauble: Great points - Could you take a look at the latest commit and let me know your thoughts on that format @rxu17 ?

rxu17: Looks good to me!

PR #172, opened 2026-07-16
[GEN-1732] Make zenodo metrics output filename date agnostic

rxu17:
# **Problem:**
This is a small PR to fix the date formatting in the Zenodo metrics output. Airflow's `logical_date` variable can vary from the actual trigger date and is actually the `execution_date` or start of the date date interval of a dag run. See these docs:
- https://airflow.apache.org/docs/apache-airflow/stable/installation/upgrading_to_airflow3.html#manual-dag-runs-and-logical-date
- https://airflow.apache.org/docs/apache-airflow/stable/faq.html#what-does-execution-date-mean

So even if you do a manual run, it will use the start of the date interval. 

`ds_nodash` is the logical_date, which equals the start of the data interval, not when the run fires. Airflow runs a scheduled job at the end of its interval but labels it by the start. The dag's schedule is `0 8 * * MON#3` (3rd Monday each month). The cron occurrences around now are May 18 → June 15 → July 20, 2026. So:
Data interval | Fires (actual run) | logical_date = ds_nodash
-- | -- | --
May 18 → Jun 15 | Jun 15 | May 18
Jun 15 → Jul 20 | Jul 20 | Jun 15

For the actual trigger/run date regardless of manual or not, would want to use `context["dag_run"].start_date` instead (with some additional formatting and also to get it into the right timezone).

Source | Manual trigger on 07/15 | Meaning
-- | -- | --
ds_nodash (logical_date) | 20260518 | start of data interval
data_interval_end | 20260615 | last schedule boundary
dag_run.start_date | 20260715 | when the run actually ran

# **Solution:**
To keep things cleaner and simplier, decided to move from using a filename formatted with a date and just have two files with various versions

# **Testing:**
- Unit tests pass
- Outputs are just two files
image


dags/zenodo_tep_metrics_dag.py
```diff
@@ -434,6 +438,7 @@ def export_reports_to_synapse(
             uploaded = File(
                 path=output.path,
                 parent_id=folder_id,
+                version_comment=SYNAPSE_VERSION_COMMENT,
```
thomasyu888, line 483: Nit: should we add provenance?

thomasyu888 (commented):
🔥 From a data product perspective, I like this. The upload date per version can be used to know when the pipeline was executed. The version comment is a nice touch as well, in case someone uploads to the file in a non DAG way.

I'll defer to @linglp for final review.  Thanks!

dags/zenodo_tep_metrics_dag.py
rxu17: I like this, this can be actually another thing we add to best practices if the dag outputs to synapse (so we know the sources of the dag outputs in synapse)

linglp (approved):
LGTM!! Thanks for the fix, @rxu17 !

PR #174, opened 2026-07-21
Refactor all DAGs to use `src.synapse_hook`

thomasyu888:
# **Problem:**
We are looking to deprecate `py-orca` to reduce maintenance

# **Solution:**
1. Implement dags with `src.synapse_hook`
2. Bonus: prep for airflow 3 and convert `schedule_interval` to `schedule`

# **Testing:**
NA

depends on #175

### DIFF
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -95,7 +95,7 @@ def resolve_dag_config(challenge_name: str, dag_params: dict, config: dict) -> d
     
     # Start with default configuration
     dag_config = {
-        "schedule_interval": "*/3 * * * *",
+        "schedule": "*/3 * * * *",
         "start_date": datetime(2024, 4, 9, tzinfo=timezone.utc),
         "catchup": False,
         "default_args": {"retries": 2},
@@ -213,7 +213,7 @@ def create_challenge_dag(challenge_name: str, config: dict):
         @task
         def get_new_submissions(**context):
             hook = SynapseHook(context["params"]["synapse_conn_id"])
-            submissions = hook.ops.get_submissions_with_status(
+            submissions = hook.get_submissions_with_status(
                 context["params"]["tower_view_id"], "RECEIVED"
             )
 
@@ -230,7 +230,7 @@ def create_challenge_dag(challenge_name: str, config: dict):
             if submissions:
                 hook = SynapseHook(context["params"]["synapse_conn_id"])
                 for submission in submissions:
-                    hook.ops.update_submission_status(
+                    hook.update_submission_status(
                         submission_id=submission, submission_status="EVALUATION_IN_PROGRESS"
                     )
                 return "stage_submissions_manifest"
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -18,8 +18,14 @@ from src.synapse_hook import SynapseHook
 
 
 # Define the path to your challenge configuration file.
-CONFIG_URL = "https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/main/dags/challenge_configs.yaml"
-
+CONFIG_URL = os.environ.get(
+    "CONFIG_URL",
+    (
+        "https://raw.githubusercontent.com/"
+        "Sage-Bionetworks-Workflows/orca-recipes/main/"
+        "dags/challenge_configs.yaml"
+    ),
+)
 
 def load_challenge_configs(url=CONFIG_URL):
     """Load challenge configurations from a raw GitHub URL."""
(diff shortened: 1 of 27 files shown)


dags/challenge_dag_factory.py
```diff
@@ -11,10 +11,11 @@
...
-
 from orca.services.nextflowtower import NextflowTowerHook
 from orca.services.nextflowtower.models import LaunchInfo
-from orca.services.synapse import SynapseHook
+
+from src.synapse_hook import SynapseHook
+
 
 # Define the path to your challenge configuration file.
 CONFIG_URL = "https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/main/dags/challenge_configs.yaml"
```
rxu17: This actually runs into an error when I tried [my tests for all dags on it](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/175) (otherwise all other tests pass):
```
AssertionError: Import error in challenge_dag_factory.py:
E         Traceback (most recent call last):
E           File "/Users/rxu/orca-recipes/orca-recipes-env/lib/python3.10/site-packages/airflow/models/dag.py", line 4255, in factory
E             with DAG(
E           File "/Users/rxu/orca-recipes/orca-recipes-env/lib/python3.10/site-packages/airflow/models/dag.py", line 653, in __init__
E             raise ValueError("At most one allowed for args 'schedule_interval', 'timetable', and 'schedule'.")
E         ValueError: At most one allowed for args 'schedule_interval', 'timetable', and 'schedule'.
E         
E       assert 'Traceback (most recent call last):\n  File "/Users/rxu/orca-recipes/orca-recipes-env/lib/python3.10/site-packages/air...nd \'schedule\'.")\nValueError: At most one allowed for args \'schedule_interval\', \'timetable\', and \'schedule\'.\n' is None
```
Because the `CONFIG_URL` points to the config on the main branch which still uses `schedule_interval`, I wonder if we should make this into an environment variable so users can configure it to point to their branch's CONFIG 

```python
CONFIG_URL = os.environ.get(
    "CONFIG_URL",
    (
        "https://raw.githubusercontent.com/"
        "Sage-Bionetworks-Workflows/orca-recipes/main/"
        "dags/challenge_configs.yaml"
    ),
)
```
And then add something like the following either under this challenge dag section in the CONTRIBUTING.md ([maybe here](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/main/CONTRIBUTING.md#overview))

> ### Local development
> 
> By default, `challenge_dag_factory.py` loads `challenge_configs.yaml` from the production (`main`) branch of this repository. To test changes to `challenge_configs.yaml` before they are merged, set the `CONFIG_URL` environment variable to either a local copy of the YAML file or the raw GitHub URL for your feature branch.

> 
> For example, to test a feature branch:
> 
> ```console
> export CONFIG_URL="https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/<YOUR_BRANCH>/dags/challenge_configs.yaml"
> ```

rxu17:
There are still mentions of `schedule_interval` in the docs. Here are the areas they will need to be updated to `schedule` (minus the dag ones which you already updated): https://github.com/search?q=repo%3ASage-Bionetworks-Workflows%2Forca-recipes%20schedule_interval&type=code

dags/challenge_dag_factory.py
rxu17: In the long run, we may want to update the challenge dag factory to first accept a relative filepath (non http) of the [challenge_configs.yaml](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/174/changes#diff-1ed283822fae1cba37e1349bf5fb66b408983bb2e3695dd469c3b2dd94a40cc5) in your repo's feature branch so it will always have those updates, and then have a try except to default to using the `main` production branch one if the challenge_configs.yaml doesn't exist locally (since maybe scheduler can't "see" anything other than the dags? Although this is something we have to confim)

thomasyu888: lets create a ticket for when we come back to the challenge DAG which will most likely be the end of the year when there is challenge infrastructure to set up

For now, this is good iteration that you created in your branch, but I think we'll eventually want to be able to test the config from the local yaml

rxu17: Yup, created the ticket. I decided to add this small doc snippet [to the all dags test suite PR instead](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/175/changes#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055). And we can revisit when we tackle challenge infrastructure.

rxu17 (approved):
LGTM! Just a few comments

BryanFauble (approved):

PR #175, opened 2026-07-21
[DPE-1764, DPE-1755] Add test suite to test all dags

rxu17:
# **Problem:**
We don't have tests that test dag structure/conventions (e.g: dags can be imported, etc)

JIRA Tickets:
- https://sagebionetworks.jira.com/browse/DPE-1755
- https://sagebionetworks.jira.com/browse/DPE-1764

# **Solution:**
Add tests that test all dags on the following:
- Each dag has at least one task
- Each dag can be imported with no errors (mimics a runtime environment that can match the expected load paths of the dags). This way all commits have these unit tests run against them to catch any import errors (instead of having to go to codespaces or production dag to see them)

# **Testing:**
- Tests run!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,12 +10,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      
+
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: '3.10'
-          
+
       - name: Install dev dependencies
         run: |
           python -m pip install --upgrade pip
@@ -24,11 +24,15 @@ jobs:
       - name: Install airflow deps
         run: |
           pip install -r requirements-airflow.txt
-          
+
       - name: Run tests
+        env:
+          # Setting the configuration URL for the challenge configs
+          CONFIG_URL: >-
+            https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/${{ github.sha }}/dags/challenge_configs.yaml
         run: |
           python -m pytest tests/ -v --tb=short
-  
+
   ghcr-publish:
     needs: test
     runs-on: ubuntu-latest
--- a/dags/challenge_dag_factory.py
+++ b/dags/challenge_dag_factory.py
@@ -18,8 +18,14 @@ from src.synapse_hook import SynapseHook
 
 
 # Define the path to your challenge configuration file.
-CONFIG_URL = "https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/main/dags/challenge_configs.yaml"
-
+CONFIG_URL = os.environ.get(
+    "CONFIG_URL",
+    (
+        "https://raw.githubusercontent.com/"
+        "Sage-Bionetworks-Workflows/orca-recipes/main/"
+        "dags/challenge_configs.yaml"
+    ),
+)
 
 def load_challenge_configs(url=CONFIG_URL):
     """Load challenge configurations from a raw GitHub URL."""
@@ -74,8 +80,8 @@ def enforce_utc_timezone(dag_config: dict):
 
             # Enforce UTC timezone
             dag_config[param] = dt.astimezone(timezone.utc)
-        
-    
+
+
 def resolve_dag_config(challenge_name: str, dag_params: dict, config: dict) -> dict:
     """
     Return the DAG configuration for a challenge.
@@ -92,7 +98,7 @@ def resolve_dag_config(challenge_name: str, dag_params: dict, config: dict) -> d
     Returns:
         dict: The resolved DAG configuration.
     """
-    
+
     # Start with default configuration
     dag_config = {
         "schedule": "*/3 * * * *",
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -24,11 +24,15 @@ jobs:
       - name: Install airflow deps
         run: |
           pip install -r requirements-airflow.txt
-          
+
       - name: Run tests
+        env:
```
rxu17, line 29: Needed to point to the feature branch's version of this file instead of `main` branch when doing the tests. This can probably be removed if we decide to refactor the challenge dag to accept local files as well for the config. These changes here, in `validate.yml` and `challenge_dag_factory.py` are to address this issue: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/174/changes#r3620356050

tests/dags/test_all_dags.py
```diff
@@ -0,0 +1,132 @@
...
+        Dict[str, dict]: A dictionary containing import errors and DAG information
+    """
+    # Marker prefixed onto the subprocess's JSON result line so we can pick it
+    # out of stdout even if Airflow logs other noise.
+    result_marker = "__DAGBAG_RESULT__"
+
+    # Script run in the subprocess: build a DagBag over the whole folder with an
+    # Airflow-style sys.path and emit import errors + discovered dag ids as one
+    # JSON line
+    dagbag_parse_script = textwrap.dedent(
```
thomasyu888, line 55: Could you speak to why having a subprocess call here is required over just python calls? This looks a little awkward and I'm probably missing something

thomasyu888 (commented):
🔥 LGTM - I have come comment and I appreciate the overall testing framework which will start making things more robust.

I'm ok with the challenge config url for now until we have a ticket to retest the challenge framework, I'm sure there was a reason why we used URL in the beginning rather than a local file path.

Will defer to @linglp for a final review

tests/dags/test_all_dags.py
rxu17: Yeah unfortunately. 

From my understanding, it's because the `subprocess` is for mimicing airflow's import environment as closely as possible (fresh from scratch) since airflow doesn't see the repository root vs running `DagBag` directly in the pytest process inherits pytest's `sys.path` (so repo root already added to `sys.path`), which can mask import issues. 

For example, when I ran `python -m pytest`, it didn't catch a DAG doing the following import error `from dags.src.utils ...` when it should have done `from src.utils ...` because the repository root was on `sys.path` so running in `subprocess` lets us control `PYTHONPATH` and `cwd` so imports behave the way they do when Airflow parses the DAGs.

dags/challenge_dag_factory.py
```diff
@@ -18,8 +18,14 @@
 
 
 # Define the path to your challenge configuration file.
-CONFIG_URL = "https://raw.githubusercontent.com/Sage-Bionetworks-Workflows/orca-recipes/main/dags/challenge_configs.yaml"
-
+CONFIG_URL = os.environ.get(
+    "CONFIG_URL",
```
rxu17, line 22: Added this so we can specify our feature branch's https: version of the url either running locally or via github actions

tests/dags/test_all_dags.py
thomasyu888: Could this be added as a comment somewhere? This is valuable information.

tests/dags/test_all_dags.py
```diff
@@ -0,0 +1,150 @@
...
+REPO_ROOT = Path(__file__).resolve().parents[2]
+DAG_FOLDER = REPO_ROOT / "dags"
+DAG_FILES = sorted(DAG_FOLDER.rglob("*.py"))
+
+# DAGs that read Airflow Variables/Connections at parse time and therefore
+# cannot be import validated without those secrets present in the environment.
+KNOWN_PARSE_TIME_DEPENDENCIES: set[str] = {
+    "synapse-dataset-to-croissant.py",
+    "synapse_dataset_to_croissant_minimal.py",
+}
```
BryanFauble, line 40: I think that this is only true for OTEL logging/tracing:

https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f41fefd174e62284dd7ce7f808cb07a2f7d4690a/dags/synapse_dataset_to_croissant_minimal.py#L175-L176
https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f41fefd174e62284dd7ce7f808cb07a2f7d4690a/dags/synapse-dataset-to-croissant.py#L207-L208


Assuming that is the only spot - I know there are ways to set up "no-op" loggers/tracers that we can gate behind a "No secrets present" check. Happy to help explore that if there is use in getting these 2 DAGs tested as well.

BryanFauble (approved):

tests/dags/test_all_dags.py
rxu17: Thanks for looking into that - I created a ticket here: https://sagebionetworks.jira.com/browse/DPE-1773 to look into that. I wouldn't say it's super urgent since those dags work (but definitely when we make updates/add new ones) but just adding it for tracking

PR #176, opened 2026-07-21
[AG-2175] changed the scheduled time from 9pm ET to 11pm ET 

linglp:
# **Problem:**
See ticket: https://sagebionetworks.jira.com/browse/AG-2175

# **Solution:**
Change scheduled time to 11pm ET

### DIFF
--- a/dags/agora_nf.py
+++ b/dags/agora_nf.py
@@ -74,7 +74,7 @@ dag_params = {
 
 dag_config = {
     "schedule": "0 3 * * *",
-    # 11:00 PM ET (EDT, UTC-4) on July 9, 2026 == 03:00 UTC on July 10, 2026
+    # July 10, 2026 at 03:00 UTC (11 PM ET)
     "start_date": datetime(2026, 7, 10, 3, 0),
     "catchup": False,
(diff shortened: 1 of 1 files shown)


thomasyu888 (commented):
🔥 LGTM - and I'll defer to @rxu17 for final review

rxu17 (approved):
LGTM!

dags/agora_nf.py
```diff
@@ -73,9 +73,9 @@
 }
 
 dag_config = {
-    "schedule": "0 1 * * *",
-    # July 10, 2026 at 01:00 UTC
-    "start_date": datetime(2026, 7, 10, 1, 0),
+    "schedule": "0 3 * * *",
+    # July 10, 2026 at 03:00 UTC (11 PM ET)
```
rxu17: nit: add in PT as well since some collaborators also work in PT time?

BryanFauble (approved):

dags/agora_nf.py
linglp: Thanks. I added that.

PR #177, opened 2026-07-22
[DPE-1319] Croissant update for Duo codes (Expanded)

BryanFauble:
https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/164 describes the problem, solution and testing very well.

What is captured below is a summary of differences from the original pull request:

# **Problem:**

  1. The generated JSON-LD still tripped a mlcroissant context warning. #164 noted "one benign informational warning remains: the `@context` differs from mlcroissant's bundled reference context (extra/missing convenience keys)." The `@context` used https:// namespace URIs and had dropped several convenience keys, so it didn't match the reference context the validator ships with.
  2. The Croissant construction was untestable outside Airflow. In #164 the entire croissant_file dict literal and its helpers lived inline in the DAG, interleaved with Snowflake/Synapse/S3/OTEL I/O. Validating the output required an AST-extraction harness (per #164's Testing section) rather than just running the real code.

# **Solution:**

JSON-LD field differences from #164 (these are the only differences in the emitted document — everything else, creator/citeAs/usageInfo/license/checksum handling/dropped fields, is byte-for-byte identical to #164):

- `@context` namespace URIs switched https:// → http:// for cr, rai, and dct:
- cr: https://mlcommons.org/croissant/ → http://mlcommons.org/croissant/
- rai: https://mlcommons.org/croissant/RAI/ → http://mlcommons.org/croissant/RAI/
- dct: https://purl.org/dc/terms/ → http://purl.org/dc/terms/
- dct:conformsTo value switched https:// → http://: https://mlcommons.org/croissant/1.1 → http://mlcommons.org/croissant/1.1.
- Restored/added `@context` convenience keys that #164 had dropped: path, repeated, replace, and added samplingRate (cr:samplingRate).

Together these align the `@context` exactly with mlcroissant's bundled reference context, resolving the context-mismatch warning. The http:// form is what the reference context and validator use.

Refactor (no output change):

- Extracted all pure Croissant-construction logic into a new module dags/src/croissant_utils.py — build_croissant_metadata() plus construct_distribution_section_for_files, construct_record_set_section_for_files, construct_bibtex_citation, construct_usage_info, construct_creators, the DUO_CODE_TO_NAME map, and a shared CROISSANT_CONTEXT constant. These functions do no I/O: given a dataset entity, its file instances, and a DataFrame of md5/content-types, they return the finished JSON-LD dict.
- Split the Snowflake lookup out of construct_distribution_section_for_files into query_snowflake_for_file_md5_and_types(), so all I/O (Snowflake, Synapse, S3, Airflow, OTEL) stays in the DAG and the transformation stays pure.
- The DAG now calls query_snowflake_for_file_md5_and_types(...) then build_croissant_metadata(...) in place of the inline dict literal.

Design decision: keeping the construction pure and separate from I/O is what makes local validation possible without the AST-extraction workaround #164 needed. No new dependencies; the DUO static-map trade-off from #164 carries over unchanged.

# **Testing:**

I used this python script (https://gist.github.com/BryanFauble/c6df2f3d2d22b2598b1f999b6e2378c0 ) to generate and create the JSONLD files and run it through the mlcroissant validation check.


thomasyu888 (approved):
🔥 Thanks for doing this, on a high level this looks great.  I think it would be helpful if you sent an example JSON-LD document that is built from this for @cconrad8 to review.

It seems like we will also have to "force re-run" everything so we re-create all the existing JSON-LD files.

I'll defer to @rxu17 for a final review on this

dags/src/croissant_utils.py
rxu17, line 1: Could a test suite be added for this utils script?

rxu17, line 1: Wondering will multiple dags end up using this utils file in the future?

rxu17:
> I used this python script (https://gist.github.com/BryanFauble/c6df2f3d2d22b2598b1f999b6e2378c0 ) to generate and create the JSONLD files and run it through the mlcroissant validation check.

Would this be the way to test this DAG locally for anyone? I am wondering, would this dag run if you passed it the custom config like: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f6ab134dbfe8787567bff4d145faf7bcab39ce35/CONTRIBUTING.md?plain=1#L256-L338 using `dag.test()` and following instructions for [local running](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f6ab134dbfe8787567bff4d145faf7bcab39ce35/README.md#local-development-without-dev-container)

cconrad8:
@rxu17 @thomasyu888 agreed if there's an example file I can spot check that would be great.

dags/src/croissant_utils.py
BryanFauble: I would image that we move the minimal croissant DAG over if we also need to make change there - Though that was not in scope for this first pass.

BryanFauble:
> It seems like we will also have to "force re-run" everything so we re-create all the existing JSON-LD files.

@thomasyu888 

This is not a problem - On the next run it would update all of the existing files. This process replaces files on each run.

BryanFauble:
@cconrad8 I added a ZIP with 4 JSON-LD files to https://sagebionetworks.jira.com/browse/DPE-1319

PR #178, opened 2026-07-23
IT-4153 Use IRSA role for S3 instead of stored AWS connection

BryanFauble:
Both croissant DAGs built the `S3Hook` from the `AWS_SYNAPSE_CROISSANT_METADATA_S3_CONN` Airflow connection, which holds a static access key. That connection overrides the IRSA credentials the pod already has.

Setting `aws_conn_id=None` makes `AwsBaseHook` fall back to boto3's default credential chain, which picks up the `AWS_ROLE_ARN` + web-identity token injected via the service-account annotation and assumes the role instead.

- `aws_conn_id=None` at all four `S3Hook` call sites (two files)
- Removed the now-unused `aws_conn_id` DAG param

The stored Airflow connection can be deleted separately once these run green.


danlu1 (approved):
Do you plan to replace param and Variable to corresponding ones in airflow.sdk?

PR #180, opened 2026-07-30
[DPE-1766] Add NextflowTower hook

thomasyu888:
# **Problem:**

Attempting to simplify permissions by deprecating py-orca

# **Solution:**

shifting seqera hook into this repo

# **Testing:**
was able to test the agora_nf workfow locally, and it triggered the workflow & is sensing the workflow that is running in seqera platform

### DIFF
--- a/dags/agora_nf.py
+++ b/dags/agora_nf.py
@@ -41,7 +41,7 @@ from airflow.models.dag import DAG
 from airflow.models.param import Param
 from slack_sdk import WebClient
 
-from src.seqera_hook import LaunchInfo, SeqeraHook
+from src.nextflow_tower_hook import LaunchInfo, NextflowTowerHook
 from src.synapse_hook import SynapseHook
 from src.utils import validate_required_secrets
 
@@ -94,7 +94,7 @@ def agora_nf_run_dag() -> DAG:
         Args:
             workspace_id (str): Workspace ID for tower run
         """
-        hook = SeqeraHook(context["params"]["tower_conn_id"])
+        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
         dataset = context["params"].get("dataset") or ""
         info = LaunchInfo(
             run_name=context["params"]["tower_run_name"],
@@ -127,7 +127,7 @@ def agora_nf_run_dag() -> DAG:
             True once the workflow has reached a terminal state (success,
             failure, or cancellation), signaling the sensor to stop poking.
         """
-        hook = SeqeraHook(context["params"]["tower_conn_id"])
+        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
         workflow = hook.get_workflow(run_id)
         print(f"Current workflow state: {workflow.status.state.value}")
         return workflow.status.is_done
@@ -144,7 +144,7 @@ def agora_nf_run_dag() -> DAG:
             A message reporting the workflow's dataset, terminal state,
             duration, and a link to the run in Tower's UI.
         """
-        hook = SeqeraHook(context["params"]["tower_conn_id"])
+        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
         workflow = hook.get_workflow(run_id)
 
         emoji = "🎉" if workflow.status.is_successful else "❌"
--- a/dags/agora_nf.py
+++ b/dags/agora_nf.py
@@ -148,7 +148,7 @@ def agora_nf_run_dag() -> DAG:
         workflow = hook.get_workflow(run_id)
 
         emoji = "🎉" if workflow.status.is_successful else "❌"
-        dataset = (workflow.params or {}).get("dataset") or "all datasets"
+        dataset = workflow.params.get("dataset") or "all datasets"
 
         duration = "unknown"
         if workflow.submit and workflow.complete:
--- a/dags/agora_nf.py
+++ b/dags/agora_nf.py
@@ -39,10 +39,9 @@ from airflow.decorators import dag, task
 from airflow.models import Variable
 from airflow.models.dag import DAG
 from airflow.models.param import Param
-from orca.services.nextflowtower import NextflowTowerHook
-from orca.services.nextflowtower.models import LaunchInfo
 from slack_sdk import WebClient
 
+from src.seqera_hook import LaunchInfo, SeqeraHook
 from src.synapse_hook import SynapseHook
 from src.utils import validate_required_secrets
 
@@ -95,7 +94,7 @@ def agora_nf_run_dag() -> DAG:
         Args:
             workspace_id (str): Workspace ID for tower run
         """
-        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
+        hook = SeqeraHook(context["params"]["tower_conn_id"])
         dataset = context["params"].get("dataset") or ""
         info = LaunchInfo(
             run_name=context["params"]["tower_run_name"],
@@ -111,7 +110,7 @@ def agora_nf_run_dag() -> DAG:
                 "dataset": dataset,
             },
         )
-        run_id = hook.ops.launch_workflow(
+        run_id = hook.launch_workflow(
             info, context["params"]["tower_compute_env_type"], ignore_previous_runs=True
         )
         return run_id
@@ -128,8 +127,8 @@ def agora_nf_run_dag() -> DAG:
             True once the workflow has reached a terminal state (success,
             failure, or cancellation), signaling the sensor to stop poking.
         """
-        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
-        workflow = hook.ops.get_workflow(run_id)
+        hook = SeqeraHook(context["params"]["tower_conn_id"])
+        workflow = hook.get_workflow(run_id)
         print(f"Current workflow state: {workflow.status.state.value}")
         return workflow.status.is_done
 
@@ -145,11 +144,11 @@ def agora_nf_run_dag() -> DAG:
             A message reporting the workflow's dataset, terminal state,
             duration, and a link to the run in Tower's UI.
         """
-        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
-        workflow = hook.ops.get_workflow(run_id)
+        hook = SeqeraHook(context["params"]["tower_conn_id"])
+        workflow = hook.get_workflow(run_id)
 
         emoji = "🎉" if workflow.status.is_successful else "❌"
-        dataset = workflow.params.get("dataset") or "all datasets"
+        dataset = (workflow.params or {}).get("dataset") or "all datasets"
 
         duration = "unknown"
         if workflow.submit and workflow.complete:
--- a/dags/src/nextflow_tower_hook.py
+++ b/dags/src/nextflow_tower_hook.py
@@ -192,7 +192,8 @@ def api_kwargs(
 
 @dataclass
 class WorkflowStatus:
-    """A workflow run's state, and what that state implies."""
+    """A workflow run's state, and what that state implies.
+    https://docs.seqera.io/platform-cloud/monitoring/overview"""
 
     state: WorkflowState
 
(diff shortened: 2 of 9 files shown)


dags/agora_nf.py
```diff
@@ -145,11 +144,11 @@ def generate_message(run_id: str, **context: Any) -> str:
...
             duration, and a link to the run in Tower's UI.
         """
-        hook = NextflowTowerHook(context["params"]["tower_conn_id"])
-        workflow = hook.ops.get_workflow(run_id)
+        hook = SeqeraHook(context["params"]["tower_conn_id"])
+        workflow = hook.get_workflow(run_id)
 
         emoji = "🎉" if workflow.status.is_successful else "❌"
-        dataset = workflow.params.get("dataset") or "all datasets"
+        dataset = (workflow.params or {}).get("dataset") or "all datasets"
```
rxu17: Does `workflow.params` errors out when it's None? Hence the {}?

thomasyu888: Yes, although - I think this was the existing behavior, so I'm unsure if we really need to add this addition. I can remove

thomasyu888:
```suggestion
        dataset = workflow.params.get("dataset") or "all datasets"
```

dags/src/nextflow_tower_hook.py
```diff
@@ -0,0 +1,1124 @@
...
+Local fallbacks (used only when the connection cannot be resolved, e.g. under
+``dag.test()`` without a metadata database) follow the names Seqera's own CLI
+uses: ``TOWER_ACCESS_TOKEN``, ``TOWER_API_ENDPOINT``, and ``TOWER_WORKSPACE``.
+
+Typical DAG usage:
+
+    hook = SeqeraHook(context["params"]["tower_conn_id"])
+    run_id = hook.launch_workflow(LaunchInfo(...), "ondemand")
+    workflow = hook.get_workflow(run_id)
+    workflow.status.is_done
```
rxu17, line 28: Will we leave it up the DAGs to do internal polling, leaving out the old `monitor_workflow_status` function: https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/b353b15618b28989c46abf287bf44e36a7ab3231/src/orca/services/nextflowtower/ops.py#L259-L279?

dags/src/seqera_hook.py
```diff
@@ -0,0 +1,1124 @@
...
+
+        Returns:
+            The Airflow connection, or None when Airflow or the connection
+            isn't available (e.g. a local run without a metadata database).
+        """
+        try:
+            from airflow.hooks.base import BaseHook
+
+            return BaseHook.get_connection(self.conn_id)
+        except Exception:
```
rxu17: Any reason for us to use a generic `exception` now over `AirflowNotFoundException`? See https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/b353b15618b28989c46abf287bf44e36a7ab3231/src/orca/services/base/hook.py#L61-L75. I didn't realize this with synapse hook

I realize synapse hook and seqera hook previously used a common function that handled this from the base airflow class in `py-orca`

dags/src/nextflow_tower_hook.py
```diff
@@ -0,0 +1,1124 @@
...
+
+        if not ignore_previous_runs:
+            latest_run = self.get_latest_previous_workflow(launch_info)
+            if latest_run:
+                state = latest_run.status.state
+                run_repr = f"{latest_run.run_name} (id='{latest_run.id}', {state=})"
+                if not latest_run.status.is_done:
+                    logger.info(f"Found an ongoing previous run: {run_repr}")
+                    return latest_run.id
+                if state in {WorkflowState.SUCCEEDED, WorkflowState.UNKNOWN}:
```
rxu17, line 832: i'm not too familiar with `UNKNOWN` states in seqera runs, do we want them to be seen as "success" and just return them?

dags/src/nextflow_tower_hook.py
rxu17, line 1: Does this mean we also need to modify everywhere with `NEXTFLOWTOWER_CONNECTION_URI` to like `SEQERA_CONNECTION_URI` like https://github.com/search?q=repo%3ASage-Bionetworks-Workflows%2Forca-recipes+NEXTFLOWTOWER_CONNECTION_URI&type=code for consistency

Even the secrets?

thomasyu888: Good point, how do you think we should operationalize this?

rxu17: I am wondering if we could use the same name as before `NextflowTower`. We currently have 12 tower connection secrets with that name and nextflow-infra still uses tower across the board (plus our slack channels). I think if we want consistency, we'd have to do it across the board so it's not confusing if someone sees Seqera vs Tower. 

Actually Seqera also still uses Nextflow Tower: https://github.com/seqeralabs/tower-cli. Could we just keep it as NextflowTowerHook then?

thomasyu888: Im ok with making this change back to reduce the number of changes since everyone knows NextflowTower is synonymous to Seqera.  That said, I think they've been trying to shift people towards: https://github.com/seqeralabs/seqerakit

dags/src/nextflow_tower_hook.py
thomasyu888: Good point, I'll add this here

dags/src/seqera_hook.py
thomasyu888: Good catch, I'll make modifications.

dags/src/nextflow_tower_hook.py
thomasyu888: @rxu17 Actually... what is the convenience monitoring function we actually need?  I see that all DAGS that trigger nextflow workflows actually has it's own monitoring function within the DAGs themselves.

* https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f6ab134dbfe8787567bff4d145faf7bcab39ce35/dags/agora_nf.py#L120
* https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f6ab134dbfe8787567bff4d145faf7bcab39ce35/dags/genie/genie-nf-validate.py#L65
* https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f6ab134dbfe8787567bff4d145faf7bcab39ce35/dags/challenge-submission-dag.py#L73
* https://github.com/Sage-Bionetworks-Workflows/orca-recipes/blob/f6ab134dbfe8787567bff4d145faf7bcab39ce35/dags/challenge_dag_factory.py#L304

Thoughts on this being a separate Jira ticket, as I don't think any DAGs are using the `nextflowtower.ops.monitor_workflow_status` function?

rxu17: That's true, I think those individual dag functions handles it (I guess just never pointed to using the shared monitoring function before). Yeah I wonder if it's overkill to have a shared function on the hook since each dag want to customize the `task.sensor` as well, it can be a separate `minor` Jira ticket.

This could turn into documentation instead for now that each dag that uses nextflow should use a function like this (if not self-explanatory)

dags/src/nextflow_tower_hook.py
rxu17: Hmm maybe we'll wait until they migrate over completely first...

dags/src/nextflow_tower_hook.py
thomasyu888: So...  I don't have too much experience with the `UNKNOWN` status as well https://docs.seqera.io/platform-cloud/monitoring/overview, but this is a copy of existing `py-orca` code.

It's not clear to me why we would want to return a workflow that is in an unknown state but it's entirely possible when the infrastructure was still in limbo that the workflow technically completed but just could not connect with the status of the underlying compute.

dags/src/seqera_hook.py
```diff
@@ -0,0 +1,1124 @@
...
+            prefix.
+        """
+        previous_workflows = []
+        for workflow in self.list_workflows():
+            if workflow.project_name != launch_info.pipeline:
+                continue
+            prefix = launch_info.run_name
+            if prefix and not (workflow.run_name or "").startswith(prefix):
+                continue
+            previous_workflows.append(workflow)
```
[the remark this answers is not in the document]
thomasyu888: This could be an optimization to look into but going to ignore for now.

dags/src/seqera_hook.py
thomasyu888: Take a look at this commit: https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/180/changes/4449b7ff923e8c2429182a745637845ed74c6eae

dags/src/nextflow_tower_hook.py
rxu17: Yeah I wonder if this is something maybe we can just expose for now (so it's not hidden, and add a NOTE or TODO) and just have individual DAGs handle if needed (while keeping the original behavior of it under `workflow.status.is_done`)? 

Something like:

```python
@property
def is_indeterminate(self) -> bool:
    return self.state == WorkflowState.UNKNOWN
```
Not a big priority

dags/src/nextflow_tower_hook.py
```diff
@@ -0,0 +1,1131 @@
...
+        """Whether the workflow run is done, irrespective of success."""
+        return self.state in TERMINAL_WORKFLOW_STATES
+
+    @property
+    def is_successful(self) -> bool:
+        """Whether the workflow run succeeded."""
+        return self.state == WorkflowState.SUCCEEDED
+
+    @property
+    def is_indeterminate(self) -> bool:
```
thomasyu888, line 221: @rxu17 i added this here

dags/src/nextflow_tower_hook.py
thomasyu888: Done in new commit

dags/src/nextflow_tower_hook.py
rxu17: Thanks! Could you add a quick docstring for this function as well highlighting the potentials/why we exposed this?

tests/dags/src/test_nextflow_tower_hook.py
```diff
@@ -0,0 +1,965 @@
...
+)
+def test_parse_datetime(value, expected):
+    assert parse_datetime(value) == expected
+
+
+# --- Models -----------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "state, is_done, is_successful",
```
rxu17: Can we add a `is_indeterminate` check here in the `pytest.mark.parametrize`?

thomasyu888: Great suggestion: [30c01dc](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/180/commits/30c01dcbb85cebee5c1f86288c360da62ef113a1)

dags/src/nextflow_tower_hook.py
thomasyu888: Done here: [30c01dc](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/180/commits/30c01dcbb85cebee5c1f86288c360da62ef113a1)

dags/src/nextflow_tower_hook.py
thomasyu888: Created this: https://sagebionetworks.jira.com/browse/DPE-1782

tests/dags/src/test_nextflow_tower_hook.py
```diff
@@ -0,0 +1,966 @@
...
+
+def test_client_unwrap_raises_on_missing_key():
+    client = NextflowTowerClient("https://tower.example.org/api", "secret")
+    assert client.unwrap({"workflow": {"id": "1"}}, "workflow") == {"id": "1"}
+    with pytest.raises(HTTPError, match="Expecting 'workflow' key"):
+        client.unwrap({}, "workflow")
+
+
+def test_client_get_items_follows_pagination(monkeypatch):
+    pages = [
```
rxu17, line 414: add test for testing `total` as well since `get_items` also supports `total`?

rxu17 (approved):
LGTM! Just one comment

tests/dags/src/test_nextflow_tower_hook.py
thomasyu888: Done here: [fc4e893](https://github.com/Sage-Bionetworks-Workflows/orca-recipes/pull/180/commits/fc4e893cb115c928a20814b5391e347f00aa879c)

PR #181, opened 2026-07-30
[DPE-1766] Refactor all Tower dags

thomasyu888:
# **Problem:**

Seqera dags need to leverage `src/seqera_hooks.py` and the last remaining synapse_hook feature needs to be implemented

# **Solution:**

Migrated all seqera dags to use `src/seqera_hooks.py` and implemented last synapse_hook feature

# **Testing:**

Tested in #180

### DIFF
--- a/dags/src/synapse_hook.py
+++ b/dags/src/synapse_hook.py
@@ -31,6 +31,20 @@ class SynapseHook:
         )
         return df["id"].tolist()
 
+    def monitor_evaluation_queue(self, evaluation_id: Union[int, str]) -> bool:
+        """Check whether an evaluation queue has submissions awaiting evaluation.
+
+        Args:
+            evaluation_id: Synapse ID of the evaluation queue to monitor.
+
+        Returns:
+            True if the queue has at least one "RECEIVED" submission.
+        """
+        submissions = self.client.getSubmissionBundles(
+            evaluation_id, status="RECEIVED"
+        )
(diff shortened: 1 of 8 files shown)


dags/src/synapse_hook.py
```diff
@@ -30,6 +30,20 @@ def get_submissions_with_status(
         )
         return df["id"].tolist()
 
+    def monitor_evaluation_queue(self, evaluation_id: Union[int, str]) -> bool:
```
rxu17, line 34: Out of curiosity, was this not caught prior because the `challenge-submission-dag` using this wasn't active?

rxu17 (commented):
Based on comments in #180, I propose we keep it as `NextflowTowerHook`

dags/src/synapse_hook.py
thomasyu888: Yes, I think that is the reason.

thomasyu888: Although one thing to ask Jenny is whether or not the `challenge-submission-dag` can actually be moved into `template_dag`, and I think this is the same problem as the `monitoring` function for nextflow workflows

rxu17 (approved):
LGTM!

dags/src/synapse_hook.py
thomasyu888: https://sagebionetworks.jira.com/browse/DPE-1786

PR #184, opened 2026-08-23
[DPE-1746] Add bdf poc dags

rxu17:
# **Problem:**
This is a POC dag factory that's meant to automate the WES somatic pipeline/RNA seq/other pipelines. You can add additional dags via the config file. Because this is meant to be a POC only (for demos) and will be shutdown after the completion of the demos (and final reporting) in a month or so, this will not contain everything that an active, persistent airflow dag should have.

This currently supports the following bioinformatics pipelines:
- sarek
- rnaseq

JIRA Ticket: https://sagebionetworks.jira.com/browse/DPE-1746

Inputs: https://sagebionetworks.jira.com/browse/DPE-1746?focusedCommentId=339906
Outputs: https://sagebionetworks.jira.com/browse/DPE-1746?focusedCommentId=339906

```mermaid
flowchart TD
    A["CurationTask<br/>task_status"]

    A -->|"Poll every 10 seconds"| B{"Task has status COMPLETE?"}
    B -->|"No"| A
    B -->|"Yes"| C["1. SYNSTAGE<br/>nf-synapse"]

    C --> D["2. nf-core pipeline<br/>e.g. Sarek"]
    D --> E["3. SYNINDEX<br/>nf-synapse"]
    E --> F["Add provenance to<br/>indexed Synapse files"]

    %% Highlight the additional steps
    classDef additional fill:#fff3cd,stroke:#d39e00,stroke-width:3px,color:#000;
    classDef standard fill:#e7f1ff,stroke:#4a76a8,stroke-width:2px,color:#000;

    class A,B,F additional;
    class C,D,E standard;
```
Special things to note with this DAG in addition to the typical nextflow pipeline process of nf-synapse, nf-core pipeline:
- Automates the pipeline by polling every 10 seconds for `task_status` annotation on a CurationTask object (that contains our RecordSet samplesheet for input into `nf-synapse`). When `task_status` is `Completed`, it kicks off the pipeline
- Adds provenance to the output files. This provenance requires that `nf-synapse` outputs the indexed files on synapse mapped to their S3 location, so we can actually link the output files. The complicated part of this is each nf-core pipeline has specific files that are the "true" outputs of that nf-core pipeline. Right now for `sarek` and `rnaseq` we tailor the code and config specifically to how to find the outputs of that pipeline. Since the S3 location URIs of the outputs contain the sample name used in generating that file, we used that to map back to the input samplesheet of the whole pipeline to get the input FASTQ file(s) that generated that output.
   - For sarek, outputs are under `../.../annotation` folder
   - For rnaseq, outputs are specifically these files: ` .../<sample_name>/.../quant.sf`

In the future, hopefully this provenance process can be less hacky

# **Testing:**
- Runs on all of the demo samplesheets (when their respective curation task is ready) and successfully creates provenance
- Both DAGs show up on the scheduler in codespaces and runs
image


scripts/aws-sso-to-env.sh
```diff
@@ -13,7 +13,7 @@ fi
 
 ENV_FILE="$(git rev-parse --show-toplevel 2>/dev/null || echo .)/.env"
 
-aws sso login --profile "${PROFILE}"
+aws sso login --profile "${PROFILE}" --use-device-code
```
rxu17, line 18: For some reason I needed to add this for codespaces otherwise when I try to `aws login` it takes me to an error page and the verification just hangs.

docker-compose.yaml
```diff
@@ -75,7 +75,10 @@ x-airflow-common:
       # AWS credentials. Two mutually-exclusive ways to authenticate (see .env.example):
       #   1. AWS_PROFILE + a mounted ~/.aws (SSO profile, auto-refreshes tokens)
       #   2. Exported short-lived AWS_ACCESS_KEY_ID/SECRET/SESSION_TOKEN
-    AWS_PROFILE: ${AWS_PROFILE:-}
```
rxu17, line 78: I'm unsure the implications of removing this for the prod server but for codespaces, it creates an error because it's looking for an aws profile (which defaults to empty here) when no aws profile is defined for the `.env` file in codespaces. With the `AWS_PROFILE: ${AWS_PROFILE:-}`, it overrode the `.env` and set it to "" whenever it wasn't in your shell. Even when I set the AWS_PROFILE in the `.env`, I think codespaces can't see `.aws`? 

I was thinking of just suggesting the short lived credentials route for codespaces. Meanwhile, it should still work locally if someone wanted to go the AWS_PROFILE + a mounted ~/.aws route

dags/bdf_poc_dag_factory.py
```diff
@@ -0,0 +1,896 @@
...
+    Args:
+        dataset: The Dataset instance containing the necessary information.
+
+    Returns:
+        LaunchInfo: The constructed LaunchInfo instance for the synindex step.
+    """
+    return LaunchInfo(
+        run_name=dataset.synindex_run_name,
+        pipeline="Sage-Bionetworks-Workflows/nf-synapse",
+        revision="dpe-1746-add-provenance-tracking", # TODO: update to main once merged
```
rxu17, line 253: Need to update when `nf-synapse` merges it in

dags/bdf_poc_dag_factory.py
```diff
@@ -0,0 +1,896 @@
...
+                {
+                    "synstage": synstage_run_id,
+                    "pipeline": pipeline_run_id,
+                    "synindex": synindex_run_id,
+                },
+                output_rows,
+            )
+
+        ready = wait_for_record_set()
+        fetch = fetch_samplesheet()
```
rxu17, line 826: Ignoring these SonarCloud warnings because this is not functions' runtime return values just an airflow thing where this is used,`XComArg` handles that TaskFlow returns at DAG-parse time

dags/bdf_poc_configs.yaml
```diff
@@ -0,0 +1,67 @@
...
+  bucket_name: ntap-add5-project-tower-bucket
+  staging_key: samplesheets/Sarek_Process/EAGER-somatic/
+  output_prefix: sarek_somatic_GRCh38
+  pipeline: nf-core/sarek
+  revision: "3.1.2"
+  run_number: 996
+  # RecordSet/CurationTask trigger mode
+  use_record_set: true
+  curation_task_id: "6957"
+  record_set_id: syn76458438
```
rxu17, line 54: Note that this recordset doesn't work and `sarek` will error out because it's linked to a samplesheet containing `Link` entities to the actual data but synstage on `nf-synapse` does not handle Link entities and actually stage the S3 files

dags/bdf_poc_dag_factory.py
```diff
@@ -0,0 +1,896 @@
...
+
+import yaml
+from airflow.decorators import dag, task
+from airflow.exceptions import AirflowException
+from airflow.models.dag import DAG
+from airflow.models.param import Param
+from airflow.providers.amazon.aws.hooks.s3 import S3Hook
+from orca.services.nextflowtower import NextflowTowerHook
+from orca.services.nextflowtower.models import LaunchInfo
+from orca.services.synapse import SynapseHook
```
thomasyu888: Is there a reason we are using orca.services here?

dags/bdf_poc_dag_factory.py
```diff
@@ -0,0 +1,896 @@
...
+        syn: The Synapse client instance.
+        params: The dictionary of parameters controlling the download behavior.
+        dest_dir: The destination directory to download the samplesheet to.
+
+    Returns:
+        str: The path to the downloaded samplesheet CSV file.
+    """
+    override_samplesheet_id = params.get("override_samplesheet_id")
+    if override_samplesheet_id:
+        return syn.get(override_samplesheet_id, downloadLocation=dest_dir).path
```
thomasyu888: Under what conditions are people overriding the recordset?

scripts/aws-sso-to-env.sh
thomasyu888: May be good to include this in a quick comment

thomasyu888 (approved):
🔥 I took a look at the DAG logic, and this all makes sense.  I like the DAG factory pattern you picked instead of having one DAG federate across different data types.

The other note is that I wonder if provenance eventually is built upon the output of the `nf-prov` plugin or enhanced within `nf-synapse` but I reading the code, I believe there is very specific provenance requirements pending the workflow.

I'm going to approve, and it _could still be nice_ if Bryan could be a final reviewer or for this DAG factory pattern to be shared in detail with the team.

dags/bdf_poc_dag_factory.py
rxu17: Thanks for the catch!

dags/bdf_poc_dag_factory.py
rxu17: It was originally engineered because maybe someone didn't have access to the input data in the samplesheet, but still wanted to test the correct triggering of the synstage workflow but thinking about it, I don't think that's the case. Will remove.

rxu17:
> The other note is that I wonder if provenance eventually is built upon the output of the nf-prov plugin or enhanced within nf-synapse but I reading the code, I believe there is very specific provenance requirements pending the workflow.

I think seqera may be moving away from `nf-prov` and supporting this instead: https://docs.seqera.io/nextflow/tutorials/data-lineage.

dags/bdf_poc_configs.yaml
rxu17, line 1: Note: explicitly set `schedule` to None for each of these since we won't be actively running them until further ado

